PR #12, opened 2015-01-13
Issue #635 by cableman: Add translation string for "search content"

cableman:
Missing translation string: http://platform.dandigbib.org/issues/635.


kasperg:
Merged and approved.

PR #13, opened 2015-02-11
Issue #808 - Added update function to enable global redirect

cableman:
Ensured that redirect module is enabled by default.

See http://platform.dandigbib.org/issues/808


vibjerg:
Approved and merged.

PR #16, opened 2015-03-18
Issue 484: Fix cookie control module for site names with UTF8 chars.

kasperg:
Cookies with such characters are not supported in IE and other browsers.

Issue http://platform.dandigbib.org/issues/484


kasperg:
I am closing this in favour of #17.

PR #18, opened 2015-04-14
Feature/feature257 opensuggestion

JrgenGNielsen:


ding2.profile
```diff
@@ -57,8 +57,6 @@ function ding2_form_alter(&$form, &$form_state, $form_id) {
     // Set default values in ting search form to help aegir/bulk installations.
     if ($form_id == 'ting_admin_ting_settings') {
       $form['ting']['ting_search_url']['#default_value'] = 'http://opensearch.addi.dk/3.0/';
-      $form['ting']['ting_scan_url']['#default_value'] = 'http://openscan.addi.dk/2.0/';
-      $form['ting']['ting_spell_url']['#default_value'] = 'http://openspell.addi.dk/1.2/';
```
kasperg: This removes the default form values but I cannot se that the form fields are removed.

I would expect that new form fields as well as default values for these would be added as well.

Is this intentional?

kasperg:
This review is a part of ding2/ting_search_autocomplete#4.

@JrgenGNielsen: Review complete. I have one general comment regarding this.

ding2.profile
JrgenGNielsen: the form fields was defined in ting.admin.inc - new form are defined in ting_search_autocomplete

kasperg: OK. I cannot spot anywhere that this setting default values functionality to help aegir installations is handled for the new opensuggestion service. Is it no longer necessary?

JrgenGNielsen: As far as Aegir goes, it would actually be preferable to do without hook_ding_install_tasks() - variables are set with a Drush script *anyway*

kasperg: OK. Then I guess we just leave it be.

kasperg:
Review OK. I will merge this once all pull requests for the change are done.

This is tied to ding2/ting_search_autocomplete#4.

PR #17, opened 2015-03-24
Issue #641 by JrgenGNielsen: Remove cookiecontrol. Add and configure eu_...

JrgenGNielsen:
...cookie_compliance module

PR #19, opened 2015-04-23
Fixed wrong project name in make file

bartvig:
Changed eu-cookie-compliance with eu_cookie_compliance.


kasperg:
@cableman: Can you review this quickly?

The changes from #17 cause the build to break.

cableman:
The PR is missing an reference to a ticket at http://platform.dandigbib.org, but as it somewhat get under http://platform.dandigbib.org/issues/641 I'll accept it.

PR #28, opened 2015-08-20
Issue #1358 by cableman: Updated autocamplete patch to match drupal core version

cableman:


kasperg:
Things that would have been fixed using the new build system.

cableman:
You are so right... was just waiting for you to say it :-D

PR #31, opened 2015-10-08
Issue #1445 by cableman: Added ting infomedia to be enbled during new installations

cableman:


kasperg:
Reviewed and approved.

PR #34, opened 2015-10-23
Support continuous integration using Circle CI

kasperg:
This change adds support for continuous integration using Circle CI.

This includes:

- Running drush make
- Re(building) theme assets to check for uncommitted changes
- Doing a site install
- Creating an artifact containing the code base after each build

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "build"]
-	path = build
-	url = git@github.com:reload/phing-drupal-template.git
--- a/README.md
+++ b/README.md
@@ -3,6 +3,8 @@ Ding2 is a continuation of [ding.TING](http://ting.dk/content/om-dingting)
 [Drupal](http://drupal.org/project/drupal) distribution for libraries as part
 of the [TING concept](http://ting.dk).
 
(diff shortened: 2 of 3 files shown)


kasperg:
I have updated the pull request with checks for compiled frontend assets.

ding2.make
```diff
@@ -356,7 +356,7 @@ libraries[ckeditor][directory_name] = "ckeditor"
 libraries[ckeditor][destination] = "libraries"
 
 libraries[leaflet][download][type] = "get"
-libraries[leaflet][download][url] = "http://leaflet-cdn.s3.amazonaws.com/build/leaflet-0.7.2.zip"
+libraries[leaflet][download][url] = "http://cdn.leafletjs.com/downloads/leaflet-0.7.3.zip"
```
cableman: Should this change to get CI include an library update ?

kasperg: Leaflet 0.7.2 is no longer available and if I did not include this then the build broke as drush make failed.

[This has already been fixed in the consolidated codebase](a841869028bb0f1eb4b7a4f891c996f9159d23f0) so I have removed the commit from this PR.

PR #35, opened 2015-10-23
Use gulp for frontend process

kasperg:
This change adds support for Gulp as a frontend processor. This replaces the current workflow based on CodeKit.

This uses:
- libsass instead of ruby based sass for faster compilation and less dependencies.
- compass-mixins instead of compass to support libsass


cableman:
Could you rebase this, as I known that the sass needs to be recompiled

kasperg:
@cableman: Done.

PR #36, opened 2015-10-29
Gulp help

kasperg:
This change adds [`gulp-help`](https://www.npmjs.com/package/gulp-help) to make using gulp a little bit more developer friendly. This shows a list of available tasks.

I suggest that `help` becomes the default task instead of `watch`.

1 bash 2015-10-29 10-50-17


cableman:
Nice, did not known that extension

PR #37, opened 2015-10-30
Issue #1477 by rimi-itk: Added salt argument to Bpi client constructor

rimi-itk:
See https://github.com/ding2/bpi-client/pull/7 for changes in bpi-client.


kasperg:
Depending on the outcome of  ding2/bpi-client#7 this change may be irrelevant.

kasperg:
This is now irrelevant.

PR #38, opened 2015-10-30
Issue #1478 by rimi-itk: Removed use of "empty" around function calls

rimi-itk:


kasperg:

PR #39, opened 2015-10-30
Fix style warnings in gulp file

kasperg:
Use strict mode and add a missing `;`.


kasperg:
ding2 - scrutinizer 2015-10-30 16-35-07

PR #47, opened 2015-11-10
Fix drush install on Circle CI

kasperg:
The Circle image comes preinstalled with a range of dependencies. `zendframework/zend-stdlib` requires PHP 5.5 and we use 5.4 on Circle. This makes the build fail when we try to install drush.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -12,9 +12,13 @@ dependencies:
     - echo "memory_limit = 512M" > ~/.phpenv/versions/$(phpenv global)/etc/conf.d/memory.ini
     # Disable mail sending. Sendmail is not working on Circle.
     - echo "sendmail_path = /bin/true" > ~/.phpenv/versions/$(phpenv global)/etc/conf.d/sendmail.ini
+    # Clean up existing global composer installation. It contains dependencies
+    # which do not match the PHP version we use. This prevents installation of
+    # other packages.
+    - rm -rf ~/.composer
     # Install drush globally using composer.
-    # Prefer source to get around GitHub API rate limits.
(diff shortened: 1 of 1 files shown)


circle.yml
```diff
@@ -12,9 +12,13 @@ dependencies:
...
     - echo "sendmail_path = /bin/true" > ~/.phpenv/versions/$(phpenv global)/etc/conf.d/sendmail.ini
+    # Clean up existing global composer installation. It contains dependencies
+    # which do not match the PHP version we use. This prevents installation of
+    # other packages.
+    - rm -rf ~/.composer
     # Install drush globally using composer.
-    # Prefer source to get around GitHub API rate limits.
-    - composer global require drush/drush:6.*  --prefer-source --no-interaction
+    # Use prefer-source to get around GitHub API rate limits.
+    - composer global require drush/drush:6.* --prefer-source --no-interaction
```
cableman: We not install drush 7 or 8, they are 100% compatible

kasperg: In our documentation we state that we support Drush 6 so I think that should be the version that we test on.

cableman: I don't think any body is using that tool it's a left over from before ding2tal.

kasperg: We are :).

guddo: I would not recommend changing the supported version of drush while consolidating the core.

kasperg: Agree with @guddo here. I do not think there is a reason for bumping this right now either.

We could consider switching to Travis to test the build using multiple versions of drush (and PHP and...) but that would have other consequences e.g. slower build times, more complex asset handling etc.

PR #52, opened 2015-11-13
Remove node_modules from build artefacts.

kasperg:
They take up space and are not needed.


kasperg:
This reduces the artifact size from ~63MB to ~54MB. 

The build fails because of #51.

PR #53, opened 2015-11-13
Issue #1497 by rimi-itk: Added push and syndication of tags

rimi-itk:


modules/bpi/bpi.push.inc
```diff
@@ -222,6 +222,15 @@ function bpi_convert_to_bpi($node, $category, $audience, $with_images = FALSE, $
     );
   }
 
+  $tags = array();
+  $tags_field = field_view_field('node', $node, variable_get('bpi_field_tags', ''));
```
kasperg: Since we do not retrieve the tags with the intent of displaying them then how about using [`field_get_items()`](https://api.drupal.org/api/drupal/modules%21field%21field.module/function/field_get_items/7) instead of `field_view_field()`?

I can see that `field_view_field()` is being used elsewhere in the code but I see no reason for doing so there either and changing it is of out scope of this review.

Also: How well does this code handle `bpi_field_tags` not being set?

modules/bpi/bpi.admin.inc
```diff
@@ -198,6 +206,20 @@ function bpi_admin_content_mapping_form($form, $form_state) {
     '#default_value' => variable_get('bpi_field_materials', ''),
   );
 
+  $form['bpi_syndication'] = array(
+    '#type' => 'fieldset',
+    '#title' => t('Syndication settings'),
+    '#prefix' => '<div id="bpi-syndication-settings-wrapper">',
+    '#suffix' => '</div>',
```
kasperg: Why is this prefix/suffix necessary? I did not spot any related CSS changes.

kasperg:
@rimi-itk: Review complete. One minor comment and a question. Well done.

modules/bpi/bpi.admin.inc
rimi-itk: The wrapper was added for consistency with #bpi-field-mapper-wrapper. I'll remove the wrapper.

modules/bpi/bpi.push.inc
rimi-itk: I'll clean up the code a bit and add a comment on why I use field_view_field.

kasperg: :+1:

modules/bpi/bpi.admin.inc
kasperg: :+1:

kasperg:
Thanks. Reviewed and approved. The CI fail has nothing to do with your work. This will be merged after the next release.

kasperg:
Merged manually to fix conflicts. Thanks!

PR #54, opened 2015-11-13
Issue #514 by Inlead: Copied related materials styling to Event CT.

guddo:
http://platform.dandigbib.org/issues/514


kasperg:
@guddo: The build fails due to unexpected changes in the compiled assets. 

Could you please rebase your change on master (to include the change in #51), reinstall node modules and recompile assets?

guddo:
@kasperg: Is it really the purpose that everyone should compile as a part of their PR? It may create enormous merge problems, especially now when the codebase is consolidated. Furthermore the output is compressed which makes it hard to compare and thereby debug.

I strongly suggest you to avoid requesting compiled files, and just compile once when the version is to be released. I've removed the compiled file from this PR - let me know if you need it updated.

For the sake of streamlined node configurations, do you have any documentation of how external contributors should install and configure their *ddbasic* - now Ruby config is replaced by Gulp?

guddo:
Don't mind my last question in regards to documentation; I see the `README` was updated accordingly in the root dir - I was looking inside the theme dir  :blush: :santa:

kasperg:
@guddo: Thanks for bringing this up. Everyone must compile frontend assets as a part of their PR. This has been part of [the code guidelines](http://ting.dk/wiki/ding-code-guidelines) since 2012.

This ensure thats:

1. We do not have a dependency on frontend asset libraries on production servers
2. Developers check that their changes do not cause problems when combined with changes submitted by other developers.
3. We avoid problems caused by differences in frontend asset pipeline between individual developer environments
4. We avoid problems caused by differences in frontend asset pipeline between developer and build environments

As I see it this guideline has not been enforced for Ding2 because ddbasic during the ding2tal project decided to use [Codekit](https://incident57.com/codekit/). This app is proprietary and not cross platform. Personally I have had problems getting it to run at all. As a workaround @cableman has compiled frontend assets manually before each release. 

Today in the current situation I agree that most of the issues above are no longer relevant:

* 1) Our build system supports automatic compilation of frontend assets. We do not have to rely on production servers on manual processes to do this.
* 3/4) These problems are actually what is causing the build to fail. However after introducing npm shrinkwrap we enforce the same versions of the frontend asset pipeline across environments. Going forward this should no longer be a problem.

This leaves 2). Historically I do not think this has been much of an issue. `gulp watch` should also ensure that developers have freshly compiled assets on their local machine.

I think your suggestion is fair. I am open to discussing updating the guidelines here.

However: Until it has been decided I will have to ask you to update your pull request.

holt83:
@kasperg 

Have you considered something like this:

http://blog.andrewray.me/dealing-with-compiled-files-in-git/

kasperg:
@holt83: Thanks for chiming in.

I have not. To be honest I did not know about some of the techniques mentioned in this article.

As I see it the general problem with some of these techniques e.g. git hooks is that they are not part of the repository and thus not easily shared between developers. I am sure there are ways to circumvent this but I have not looked into it.

If you have an technical approach which addresses concerns raised by @guddo or myself you are more than welcome to submit a pull request for it.

kasperg:
Approved - although the changes do not apply cleanly.

kasperg:
Merged manually to fix conflicts. Thanks!

PR #55, opened 2015-11-13
Issue #1476 by rimi-itk: Added push and syndication of assets

rimi-itk:


modules/bpi/bpi.images.inc
```diff
@@ -26,7 +26,8 @@ function bpi_syndicate_images($type = 'ajax') {
   $commands[] = ajax_command_ding_popup(
     'bpi-syndicate-images',
     t('Syndicate images'),
-    drupal_render($form)
+    drupal_render($form),
+    array('refresh' => TRUE)
```
kasperg: Why is this needed? It does not seem to be relevant to the change at hand.

modules/bpi/bpi.images.inc
```diff
@@ -54,13 +55,15 @@ function bpi_assets_form($form, $form_state, $assets = array()) {
...
-        '#default_value' => 1,
-        '#prefix' => '<div class="bpi-syndicate-asset"><img src="' . $url . '" width="100" height="100" />',
-        '#suffix' => '</div>',
-      );
+    foreach ($assets as $type => $items) {
+      foreach ($items as $index => $asset) {
+        $form['bpi_assets']['asset::' . $type . '::' . $index] = array(
+          '#type' => 'checkbox',
+          '#default_value' => 1,
+          '#prefix' => '<div class="bpi-syndicate-asset"><img src="' . $asset['path'] . '" width="100" height="100" />' . '<br/> (' . bpi_get_image_display_name($asset['type']) . ')',
```
kasperg: How about using [`theme('image')`](https://api.drupal.org/api/drupal/includes%21theme.inc/function/theme_image/7) here?

modules/bpi/bpi.images.inc
```diff
@@ -54,13 +55,15 @@ function bpi_assets_form($form, $form_state, $assets = array()) {
...
 
-    foreach ($assets as $asset => $url) {
-      $form['bpi_assets'][$asset] = array(
-        '#type' => 'checkbox',
-        '#default_value' => 1,
-        '#prefix' => '<div class="bpi-syndicate-asset"><img src="' . $url . '" width="100" height="100" />',
-        '#suffix' => '</div>',
-      );
+    foreach ($assets as $type => $items) {
+      foreach ($items as $index => $asset) {
```
kasperg: It seems like the structure or the `$assets` parameter has been changed. It used to be a flat array of urls. It is now an array of types where each type is an array of assets of that type. Please update the DocBlock accordingly.

modules/bpi/bpi.images.inc
```diff
@@ -88,32 +91,45 @@ function bpi_assets_form($form, $form_state, $assets = array()) {
  * @ingroup forms
  */
 function bpi_assets_form_submit($form, &$form_state) {
-  $input = array_keys($form_state['input']);
+  $bpi_assets = $_SESSION['bpi']['assets'];
+
```
kasperg: Why is session handling necessary? Could we store the assets as a form value instead?

modules/bpi/bpi.images.inc
```diff
@@ -213,3 +229,71 @@ function bpi_fetch_image($filename, $image_url) {
 
   return file_save_data($result->data, $filename, FILE_EXISTS_REPLACE);
 }
+
+/**
+ * Compute the image type from a field name on a node.
```
kasperg: I have gone through the code a couple of times and I still do not understand what the _type_ of an image is. Could you elaborate?

modules/bpi/bpi.images.inc
```diff
@@ -213,3 +229,71 @@ function bpi_fetch_image($filename, $image_url) {
...
+}
+
+/**
+ * Get syndicated images from session.
+ *
+ * @return mixed
+ *   The value if any.
+ */
+function bpi_get_syndicated_images() {
+  return isset($_SESSION['bpi']['syndicated_images']) ? $_SESSION['bpi']['syndicated_images'] : NULL;
```
kasperg: How about returning an empty array instead of `NULL`?

modules/bpi/bpi.module
```diff
@@ -397,6 +412,116 @@ function bpi_form_node_form_alter(&$form, &$form_state, $form_id) {
...
+ */
+function bpi_place_images(array &$form, $current_language, array $syndicated_images) {
+  $node = $form['#node'];
+  $image_field_names = bpi_fetch_image_fields($node->type);
+  if (!empty($image_field_names)) {
+    foreach ($syndicated_images as $image_type => $images) {
+      foreach ($image_field_names as $image_field_name) {
+        if (bpi_is_bpi_image_field($image_field_name, $node, $image_type)) {
+          $field = &$form[$image_field_name][$current_language];
+          $cardinality = isset($field['#cardinality']) ? $field['#cardinality'] : 1;
```
kasperg: What is the cardinality of a field with an unlimited number of values?

modules/bpi/bpi.module
```diff
@@ -397,6 +412,116 @@ function bpi_form_node_form_alter(&$form, &$form_state, $form_id) {
...
+    // Find <img/> elements in content and replace with file placeholders.
+    $image_pattern = '@<img\s*[^>]*\ssrc=[\'"](?<url>[^\'">]+)[\'"][^>]*>@';
+
+    $content = preg_replace_callback($image_pattern, function($matches) use ($current_language, $url_to_image_map) {
+      $url = $matches['url'];
+      if (isset($url_to_image_map[$url])) {
+        $image = $url_to_image_map[$url];
+        $file = $image['@managed_file'];
+
+        // Build data for file placeholder.
```
kasperg: This format is directly coupled to the `media_wysiwyg` filter. It would be good to have this dependency described somewhere. I could not find any method in `media_wysiwyg` which we could at least add a comment stating this.

modules/bpi/bpi.push.inc
```diff
@@ -191,24 +193,29 @@ function bpi_convert_to_bpi($node, $category, $audience, $with_images = FALSE, $
...
-          // Image pseudo-check.
-          if (@getimagesize($file_url)) {
-            $bpi_content['images'][] = array(
-              'path' => $file_url,
-              'alt' => '',
-              'title' => '',
-            );
+        $image_type = bpi_get_image_type($field_name, $node);
+        if (!empty($image_type)) {
+          $field_value = field_view_field('node', $node, $field_name);
```
kasperg: Should this be [`field_get_items()`](https://api.drupal.org/api/drupal/modules%21field%21field.module/function/field_get_items/7) instead?

modules/bpi/bpi.module
```diff
@@ -397,6 +412,116 @@ function bpi_form_node_form_alter(&$form, &$form_state, $form_id) {
...
+          'view_mode' => 'default',
+          'fields' => array(
+            'format' => 'default',
+            'field_file_image_alt_text[' . $current_language . '][0][value]' => isset($image['alt']) ? $image['alt'] : $image['name'],
+            'field_file_image_title_text[' . $current_language . '][0][value]' => isset($image['title']) ? $image['title'] : $image['name'],
+          ),
+          'type' => 'media',
+          'attributes' => array(
+            'width' => isset($image['width']) ? $image['width'] : 800,
+            'height' => isset($image['height']) ? $image['height'] : 100,
```
kasperg: Why are 800 and 100 default values?

modules/bpi/bpi.module
```diff
@@ -397,6 +412,116 @@ function bpi_form_node_form_alter(&$form, &$form_state, $form_id) {
...
+  if (!empty($syndicated_images)) {
+    $url_to_image_map = array();
+    foreach ($syndicated_images as $type => $images) {
+      foreach ($images as $image) {
+        $url_to_image_map[$image['path']] = $image;
+      }
+    }
+
+    // Find <img/> elements in content and replace with file placeholders.
+    $image_pattern = '@<img\s*[^>]*\ssrc=[\'"](?<url>[^\'">]+)[\'"][^>]*>@';
```
kasperg: Should this be case insensitive?

modules/bpi/bpi.module
```diff
@@ -318,14 +319,17 @@ function bpi_form_node_form_alter(&$form, &$form_state, $form_id) {
...
-    if (isset($form[$body][$current_language][0]['value'])) {
-      $form[$body][$current_language][0]['value']['#default_value'] = $bpi_body;
-    }
-    elseif (isset($form[$body][$current_language][0])) {
-      $form[$body][$current_language][0]['#default_value'] = $bpi_body;
-    }
 
-    if ($bpi_id) {
+    $syndicated_images = bpi_get_syndicated_images();
+    if (is_array($syndicated_images)) {
```
kasperg: The handling of `bpi_get_syndicated_images()` which can either return an array, NULL or FALSE seems fishy to be. Is there anything we can do to improve this?

modules/bpi/bpi.push.inc
```diff
@@ -191,24 +193,29 @@ function bpi_convert_to_bpi($node, $category, $audience, $with_images = FALSE, $
...
+
+          if (!empty($field_value['#items'][0]['fid'])) {
+            $file = file_load($field_value['#items'][0]['fid']);
+            if (is_object($file)) {
+              $bpi_content['assets'][] = array(
```
kasperg: This changes the format of the requests set to BPI? Does all versions of the server support this?

modules/bpi/bpi.push.inc
```diff
@@ -191,24 +193,29 @@ function bpi_convert_to_bpi($node, $category, $audience, $with_images = FALSE, $
...
-        if (!empty($field_value['#items'][0]['uri'])) {
-          $file_url = file_create_url($field_value['#items'][0]['uri']);
-          // Image pseudo-check.
-          if (@getimagesize($file_url)) {
-            $bpi_content['images'][] = array(
-              'path' => $file_url,
-              'alt' => '',
-              'title' => '',
-            );
+        $image_type = bpi_get_image_type($field_name, $node);
```
kasperg: `bpi_get_image_type()` type hints the second argument to array. Line 199 indicates that node is an object. That should result in a fatal error?

modules/bpi/bpi.images.inc
```diff
@@ -213,3 +229,71 @@ function bpi_fetch_image($filename, $image_url) {
...
+ *
+ * @return NULL|string
+ *   The image type.
+ */
+function bpi_get_image_type($image_field_name, array $node) {
+  /*
+   * We assume that the image type is the last part of the field name, i.e.
+   *   field_{{node type}}_{{image type}}
+   */
+  $pattern = '/field_' . preg_quote($node->type, '/') . '_(?<image_type>.+)/';
```
kasperg: The $node parameter is type hinted to array but accessed as an object. That looks odd to me.

kasperg:
@rimi-itk Review complete. I have lots of minor comments and questions.

modules/bpi/bpi.images.inc
rimi-itk: We need the main content form to refresh after closing the assets (images) popup to place syndicated (downloaded) images in the content.

modules/bpi/bpi.images.inc
rimi-itk: We need session values to send values from the main content form and the assets popup. The original code uses this as well (cf. https://github.com/ding2/ding2/blob/master/modules/bpi/bpi.images.inc#L97-L99), but I moved it to the top of the function for clarity.

modules/bpi/bpi.images.inc
rimi-itk: The bpi web service supporting assets (rather than just images) uses a custom string, the asset type, to distinguish between different types of assets connected to content (cf. https://docs.google.com/document/d/1IAUoEo1MgdVkQCh5sGPhSNLuw5Wau1Gg8UsmbVErMUE/edit#heading=h.herx6gykkgqz). A piece of content can have multiple assets of each type. Inline images (in the body field) are stored with the type "body".

modules/bpi/bpi.images.inc
rimi-itk: I use NULL to indicate "no value". This is used by the "fishy" part of the code (https://github.com/ding2/ding2/pull/55/files#r46263721).

modules/bpi/bpi.module
rimi-itk: I thought a lot about this part, but I couldn't come up with a better/simpler solution to having 3 states ("syndicating", "some images syndicated", "no images syndicated"). At some oint, I was using two variables, a boolean and an asset list, but that didn't make the core more clear (as I see it).

modules/bpi/bpi.push.inc
rimi-itk: We need to notify local admins about this change in the bpi web service api. Carsten is working on how to actually do this.

modules/bpi/bpi.module
rimi-itk: It probably should be … I'll change it.

modules/bpi/bpi.module
rimi-itk: Good question! I don't know … I'll find some better defaults …

modules/bpi/bpi.push.inc
rimi-itk: It should fail. And it does. We have tested this extensively, but never run into this issue. I have no idea why, but I'll fix it.

modules/bpi/bpi.images.inc
rimi-itk: Nice catch. I have no excuse for this bug …

modules/bpi/bpi.module
rimi-itk: It's −1. I'll fix the code …

modules/bpi/bpi.push.inc
rimi-itk: Yeah, that seems like the right thing to use. Thanks.

rimi-itk:
Thanks for a very thorough code review. I hope I have fixed everything properly.

modules/bpi/bpi.images.inc
kasperg: Thanks for elaborating.

There are a few things that I find a bit unclear going through the codebase:

1. Even though we are working with assets instead of images there are places in the codebase that only support images e.g. the `` tag in `bpi_assets_form()`.
2. Functions, parameter names and comments - existing as well as new - refer to images instead of assets

The situation might occur because the service supports assets but the only assets we support currently are images. Also: I often read [image types as JPEG, PNG, GIF etc](http://php.net/manual/en/function.imagetypes.php).

I think it would be worth being more explicit about this in the codebase:

- Adding a sentence or two in the DocBlock about the asset types - in the file DocBlock or here
- Reviewing function names, parameters etc. for consistency. An example: This function is called `bpi_get_image_type()` but I think it would be a lot more clear if it was called `bpi_get_asset_type()`.

modules/bpi/bpi.module
kasperg: Alright. Do we need any defaults at all?

modules/bpi/bpi.push.inc
kasperg: :+1:

modules/bpi/bpi.module
kasperg: There might not be a better solution. 

To be honest I was also confused by lines L332 + L341:

- L332: This is the state when we are syndicating images. During this we set the syndication flag to FALSE.
- L341: When the syndication flag is set to FALSE we are in the state that no images were syndicated?

modules/bpi/bpi.images.inc
kasperg: Roger.

modules/bpi/bpi.module
kasperg: Thanks. Adding a comment stating that a field with an unlimited number of values has cardinality -1 would be good. This would explain the otherwise odd looking condition.

modules/bpi/bpi.module
```diff
@@ -397,6 +412,119 @@ function bpi_form_node_form_alter(&$form, &$form_state, $form_id) {
...
+          'view_mode' => 'default',
+          'fields' => array(
+            'format' => 'default',
+            'field_file_image_alt_text[' . $current_language . '][0][value]' => isset($image['alt']) ? $image['alt'] : $image['name'],
+            'field_file_image_title_text[' . $current_language . '][0][value]' => isset($image['title']) ? $image['title'] : $image['name'],
+          ),
+          'type' => 'media',
+          'attributes' => array(
+            'width' => isset($image['width']) ? $image['width'] : 200,
+            'height' => isset($image['height']) ? $image['height'] : 200,
```
kasperg: Are `width` and `height` required attributes in the placeholder? If not I would think it better to leave them out instead of risking a distorted image. If the image on the other hand is very large we risk displaying it as is. Hmmm.

If you have had similar considerations then at least adding a comment why these defaults were chosen would be good.

kasperg:
@rimi-itk: Great. I would appreciate two comments and then we should be good.

modules/bpi/bpi.module
rimi-itk: I have removed the fallback values. On closer inspection of the bpi service manual it turns out the width and height will always be defined.

kasperg:
Thanks a bunch. Reviewed and approved. Awaiting merge.

kasperg:
Merged. Thanks!

PR #57, opened 2015-11-19
Fix build status badge

kasperg:
The badge always seems to be broken. Switching to PNG may help.


kasperg:
This works now with the original code. No need to merge this.

PR #62, opened 2015-11-25
Issue #554 by Inlead: Various permissions.

guddo:
http://platform.dandigbib.org/issues/554


modules/ding_permissions/ding_permissions.module
```diff
@@ -898,6 +901,18 @@ function ding_permissions_secure_permissions($role) {
...
   );
+
+  // Add editor specific permissions by node type,
+  // using override node options module permissions.
+  $node_types = node_permissions_get_configured_types();
+  foreach ($node_types as $node_type) {
+    if (in_array('create ' . $node_type . ' content', $permissions['editor'])) {
+      $permissions['editor'][] = 'override ' . $node_type . ' promote to front page option';
+      $permissions['editor'][] = 'override ' . $node_type . ' authored on option';
+      $permissions['editor'][] = 'override ' . $node_type . ' authored by option';
```
kasperg: Why do we give the override authored on permission? As I read the issue setting the author of the content is not part of the scope.

modules/ding_permissions/ding_permissions.module
```diff
@@ -503,9 +503,11 @@ function ding_permissions_secure_permissions($role) {
       'administer place2book settings',
       'administer shortcuts',
       'administer url aliases',
+      'administer users',
```
kasperg: Adding the administer users permission to editors seems like a bad idea - even though that is what is suggested by the business. It provides to wide permissions compared to what they need. I have asked them to reconsider this.

kasperg:
I have done through this PR. Changes look good but we have to discuss part of this change with the business.

modules/ding_permissions/ding_permissions.info
```diff
@@ -4,6 +4,7 @@ core = 7.x
 package = Ding!
 php = 5.2.4
 dependencies[] = ctools
+dependencies[] = override_node_options
```
kasperg: Please add an update hook which enables the module as well.

JrgenGNielsen:
There's a small change needed in circle.yml needed, to pass tests in Circle CI:
Fra: - rm -rf ~/.composer
Til: - rm -rf ~/.composer/vendor
See also: https://github.com/ding2/ding2/pull/183

guddo:
Are you sure that we should amend in circle.yml in this PR?
I would recommend you to this in a separate PR and rebase.
/cc @kasperg

guddo:
PR #183 is already merged. I have rebased this PR.

JrgenGNielsen:
reviewed and accepted

kasperg:
Rebasing is the right way to go. Thanks @guddo.

PR #69, opened 2015-11-25
Issue #556 by Inlead: Adds pagepreview module.

guddo:
http://platform.dandigbib.org/issues/556


modules/ding_event/ding_event.info
```diff
@@ -19,6 +19,7 @@ dependencies[] = menu_position
 dependencies[] = number
 dependencies[] = og_ui
 dependencies[] = page_manager
+dependencies[] = pagepreview
```
kasperg: I do not think adding this module as a dependency here is needed.

The `ding_event` module does not have any configuration or code related to pagepreview so as I see it there are no actual dependencies between the two.

Since pagepreview is automatically enabled for all content types across the site I am fine with just enabling it in the install profile and leaving it at that.

modules/ding_news/ding_news.info
```diff
@@ -23,6 +23,7 @@ dependencies[] = og
 dependencies[] = og_ui
 dependencies[] = options
 dependencies[] = page_manager
+dependencies[] = pagepreview
```
kasperg: See comment above.

kasperg:
Review complete. A few minor comments. Otherwise good to go. Please remember to rebase your changes on master so the pull request applies cleanly.

kasperg:
Reviewed and approved. There are a lot of update hooks reviewed but not merged at the moment. We will probably have to deal with these merges manually.

kasperg:
Merged. Thanks!

PR #70, opened 2015-11-25
Issue #707 by Inlead: Adds recommendation text to image fields.

guddo:
http://platform.dandigbib.org/issues/472

### DIFF
--- a/modules/ding_base/ding_base.module
+++ b/modules/ding_base/ding_base.module
@@ -535,3 +535,47 @@ function ding_base_rewrite_color_stylesheet() {
   variable_set('color_' . $theme . '_stylesheets', $css);
   variable_set('color_' . $theme . '_files', $paths['files']);
 }
+
+/**
+ * Implements hook_field_widget_form_alter().
+ */
+function ding_base_field_widget_form_alter(&$element, &$form_state, $context) {
+  $names = array(
+    'field_ding_page_title_image',
+    'field_ding_page_list_image',
+    'field_ding_tabroll_image',
+    'field_ding_news_title_image',
+    'field_ding_news_list_image',
+    'field_ding_library_title_image',
+    'field_ding_library_list_image',
+    'field_ding_group_title_image',
+    'field_ding_group_list_image',
+    'field_ding_event_title_image',
+    'field_ding_event_list_image'
+  );
+
+  if (isset($element['#field_name']) && in_array($element['#field_name'], $names)) {
+    $displays = array_filter($context['instance']['display'], function ($i) {
+      return isset($i['settings']['styles']) && !empty($i['settings']['styles']);
+    });
+    if (!empty($displays)) {
+      $display = reset($displays);
+      $style = image_style_load($display['settings']['styles']['max_style']);
+      $effects = array_filter($style['effects'], function ($i) {
+        return isset($i['data']['width']) && isset($i['data']['height']);
+      });
+
+      if (!empty($effects)) {
+        $effect = reset($effects);
+        $width = !empty($effect['data']['width']) ? $effect['data']['width'] : 'any';
+        $height = !empty($effect['data']['height']) ? $effect['data']['height'] : 'any';
+
+        $text = t('Recommended image size is: width %width and height %height', array(
+          '%width' => $width,
+          '%height' => $height,
+        ));
+        $element['#suffix'] = '<div class="description">' . $text . '</div>';
+      }
+    }
+  }
+}
--- a/modules/ding_base/ding_base.module
+++ b/modules/ding_base/ding_base.module
@@ -540,39 +540,54 @@ function ding_base_rewrite_color_stylesheet() {
  * Implements hook_field_widget_form_alter().
  */
 function ding_base_field_widget_form_alter(&$element, &$form_state, $context) {
-  $names = array(
-    'field_ding_page_title_image',
-    'field_ding_page_list_image',
-    'field_ding_tabroll_image',
-    'field_ding_news_title_image',
-    'field_ding_news_list_image',
-    'field_ding_library_title_image',
-    'field_ding_library_list_image',
-    'field_ding_group_title_image',
-    'field_ding_group_list_image',
-    'field_ding_event_title_image',
-    'field_ding_event_list_image'
-  );
-
-  if (isset($element['#field_name']) && in_array($element['#field_name'], $names)) {
-    $displays = array_filter($context['instance']['display'], function ($i) {
-      return isset($i['settings']['styles']) && !empty($i['settings']['styles']);
-    });
-    if (!empty($displays)) {
-      $display = reset($displays);
-      $style = image_style_load($display['settings']['styles']['max_style']);
-      $effects = array_filter($style['effects'], function ($i) {
-        return isset($i['data']['width']) && isset($i['data']['height']);
+  // Our goal only image fields.
+  if (isset($element['#type']) && $element['#type'] == 'media') {
+    $display = NULL;
+    $displays = $context['instance']['display'];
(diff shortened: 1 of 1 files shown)


modules/ding_base/ding_base.module
```diff
@@ -535,3 +535,47 @@ function ding_base_rewrite_color_stylesheet() {
...
+    'field_ding_news_list_image',
+    'field_ding_library_title_image',
+    'field_ding_library_list_image',
+    'field_ding_group_title_image',
+    'field_ding_group_list_image',
+    'field_ding_event_title_image',
+    'field_ding_event_list_image'
+  );
+
+  if (isset($element['#field_name']) && in_array($element['#field_name'], $names)) {
```
kasperg: This change seems very helpful but is the predefined list of field names necessary? In other words: Could we apply this change for all image fields which have necessary configuration?

modules/ding_base/ding_base.module
```diff
@@ -535,3 +535,47 @@ function ding_base_rewrite_color_stylesheet() {
...
+    });
+    if (!empty($displays)) {
+      $display = reset($displays);
+      $style = image_style_load($display['settings']['styles']['max_style']);
+      $effects = array_filter($style['effects'], function ($i) {
+        return isset($i['data']['width']) && isset($i['data']['height']);
+      });
+
+      if (!empty($effects)) {
+        $effect = reset($effects);
```
kasperg: This assumes that the first effect is the one with a width/height set. Is this a safe assumption?

modules/ding_base/ding_base.module
```diff
@@ -535,3 +535,47 @@ function ding_base_rewrite_color_stylesheet() {
...
+        $width = !empty($effect['data']['width']) ? $effect['data']['width'] : 'any';
+        $height = !empty($effect['data']['height']) ? $effect['data']['height'] : 'any';
+
+        $text = t('Recommended image size is: width %width and height %height', array(
+          '%width' => $width,
+          '%height' => $height,
+        ));
```
kasperg: I think this implementation is a bit naive.

1. We cannot assume that the first effect is the one which resizes the image. It might be the current situation but when ding2/ding2#73 land this will no longer hold. In my opinion we should look for known effects which resizes the image and pick the last one which will be applied.
2. Not all presets have width *and* height set. The text does not take this into account.

Finally I do not agree with the text *Recommended image size is* but I will discuss this with the business.

kasperg:
Review complete. I have a couple of issues here which should be addressed.

Also:
- Add some comments about what is going on here and *why*. This is ~40 LOC without a single inline comment
- [Fix the code style comments](https://scrutinizer-ci.com/g/ding2/ding2/inspections/a4ce536c-465a-4131-8e13-f55bf0574eb4)

modules/ding_base/ding_base.module
```diff
@@ -535,3 +535,47 @@ function ding_base_rewrite_color_stylesheet() {
...
+    'field_ding_event_title_image',
+    'field_ding_event_list_image'
+  );
+
+  if (isset($element['#field_name']) && in_array($element['#field_name'], $names)) {
+    $displays = array_filter($context['instance']['display'], function ($i) {
+      return isset($i['settings']['styles']) && !empty($i['settings']['styles']);
+    });
+    if (!empty($displays)) {
+      $display = reset($displays);
```
kasperg: How do we know that the first display is the one we want to work with?

modules/ding_base/ding_base.module
```diff
@@ -535,3 +535,62 @@ function ding_base_rewrite_color_stylesheet() {
...
+        $dimensions['width'] = $preview_image['width'];
+        $dimensions['height'] = round($dimensions['width'] / $preview_image['width'] * $preview_image['height']);
+      }
+      else {
+        $dimensions['height'] = $preview_image['height'];
+        $dimensions['width'] = round($dimensions['height'] / $preview_image['height'] * $preview_image['width']);
+      }
+
+      // If after apllying effects dimensions wasn't changed, when we can't recommend.
+      if ($dimensions['height'] != 600 && $dimensions['width'] != 800) {
```
kasperg: These dimensions should be determined by inspecting the sample image. They only work for the default (or one with the exact same dimensions).

modules/ding_base/ding_base.module
```diff
@@ -535,3 +535,62 @@ function ding_base_rewrite_color_stylesheet() {
...
+      if (!file_exists($preview_file)) {
+        image_style_create_derivative($style, $original_path, $preview_file);
+      }
+
+      // Calculate dimensions.
+      $dimensions = array();
+      $preview_image = image_get_info($preview_file);
+      if ($preview_image['width'] > $preview_image['height']) {
+        $dimensions['width'] = $preview_image['width'];
+        $dimensions['height'] = round($dimensions['width'] / $preview_image['width'] * $preview_image['height']);
```
kasperg: Why is this calculation necessary?

As I understand it:

```php
$dimensions['width'] = $preview_image['width'];
$dimensions['height'] = round($dimensions['width'] / $preview_image['width'] * $preview_image['height']);
```

Is the same as

```php
$dimensions['width'] = $preview_image['width'];
$dimensions['height'] = round($preview_image['width'] / $preview_image['width'] * $preview_image['height']);
```
Which is the same as

```php
$dimensions['width'] = $preview_image['width'];
$dimensions['height'] = round($preview_image['height']);
```

Which leads to

```php
$dimensions['width'] = $preview_image['width'];
$dimensions['height'] = $preview_image['height'];
```

modules/ding_base/ding_base.module
```diff
@@ -535,3 +535,62 @@ function ding_base_rewrite_color_stylesheet() {
...
+      else {
+        $dimensions['height'] = $preview_image['height'];
+        $dimensions['width'] = round($dimensions['height'] / $preview_image['height'] * $preview_image['width']);
+      }
+
+      // If after apllying effects dimensions wasn't changed, when we can't recommend.
+      if ($dimensions['height'] != 600 && $dimensions['width'] != 800) {
+        array_walk($dimensions, function (&$i, $k) {
+          $i =  "{$i} px {$k}";
+        });
```
kasperg: This means that width and height will not be translated.

modules/ding_base/ding_base.module
```diff
@@ -535,3 +535,62 @@ function ding_base_rewrite_color_stylesheet() {
...
+    else {
+      $displays = array_filter($context['instance']['display'], function ($i) {
+        return isset($i['settings']['styles']) && !empty($i['settings']['styles']);
+      });
+      $display = !empty($displays) ? reset($displays) : NULL;
+    }
+
+    if (!empty($display)) {
+      // We will use functionality of image module.
+      $style = image_style_load($display['settings']['styles']['max_style']);
```
kasperg: Can we be sure that *all* displays have a `max_style`?

kasperg:
Reviewed. After the comments in my initial review then testing the dimensions of a transformed image is a clever idea.

However I still think it is flawed. Consider this case:

- The image style scales an image to maximum dimensions of 400x400
- The sample image is 800x600
- The resulting text will be "The image will automatically be resized. The maximum dimensions are 400 px width x 300 px height (Original: 800x600).
- If the user uploaded a 500x500 image then the maximum dimensions would be 400 x 400.

On other words: I think It is quite difficoult to come up with a text that is not misleading. I have raised this with the business as well.

From a technical perspective perhaps [the summary theme](https://api.drupal.org/api/drupal/modules!image!image.api.php/function/hook_image_effect_info/7) info could be useful?

kasperg:
Reviewed and approved. Awaiting merge. Thanks.

kasperg:
Merged. Thanks!

PR #74, opened 2015-11-30
Issue #735 by Inlead: "Promote to frontpage" configuration.

guddo:
http://platform.dandigbib.org/issues/735


modules/ding_frontend/ding_frontend.admin.inc
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Administration form.
+ */
+function ding_frontend_settings($form, &$form_state) {
+  $types = array(
+    'ding_news' => t('News'),
+    'ding_event' => t('Event'),
+  );
+  $form['ding_frontend_fe_promote_types'] = array(
+    '#title' => t('Promote types to Fronpage'),
```
kasperg: Typo here.

modules/ding_frontend/ding_frontend.admin.inc
```diff
@@ -0,0 +1,46 @@
...
+    '#type' => 'checkboxes',
+    '#options' => $types,
+    '#default_value' => variable_get('ding_frontend_fe_promote_types', array('ding_event' => 'ding_event')),
```
kasperg: Please add a `#description` entry to the form element.

modules/ding_frontend/ding_frontend.admin.inc
```diff
@@ -0,0 +1,46 @@
...
+  );
+  foreach ($types as $type) {
+    $conf = variable_get('node_options_' . $type, array('status', 'promote'));
+    if (in_array('promote', $conf)) {
+      unset($conf[array_search('promote', $conf)]);
+    }
+    if ($form_state['values']['ding_frontend_fe_promote_types'][$type]) {
+      $conf[] = 'promote';
+    }
+    variable_set('node_options_' . $type, $conf);
```
kasperg: If we allow setting of these variables after a deployment then their configuration should also be moved away from features. Otherwise subsequent deploys will override the new settings when features are reverted.

I suggest removing the variables from the corresponding features and instead setting their defaults in a `hook_install()` in their respective modules.

modules/ding_frontend/ding_frontend.admin.inc
```diff
@@ -0,0 +1,46 @@
...
+    'ding_event' => t('Event'),
+  );
+  $form['ding_frontend_fe_promote_types'] = array(
+    '#title' => t('Promote types to Fronpage'),
+    '#type' => 'checkboxes',
+    '#options' => $types,
+    '#default_value' => variable_get('ding_frontend_fe_promote_types', array('ding_event' => 'ding_event')),
+  );
+
+  $form = system_settings_form($form);
```
kasperg: I do not think using `system_settings_form()` makes much sense here. The variable could get out of sync with the linked variables and changing the strongarmed variable will not change the actual configuration.

This current submit function looks good so I suggest:

- Using the state of the `node_options` variables to determine default values 
- Not using `system_settings_form()``
- Removing the now obsolete `ding_frontend_fe_promote_types` variable from strongarm.

modules/ding_frontend/ding_frontend.module
```diff
@@ -6,6 +6,33 @@
...
+
+/**
+ * Implements hook_menu().
+ */
+function ding_frontend_menu() {
+  $items = array();
+
+  $items['admin/config/ding/fe-promote'] = array(
+    'title' => 'Promote CT\'s to frontpage',
+    'description' => 'Promote several CT\' to frontpage.',
```
kasperg: This is more related to administering which content types support frontpage promotion.

Also: Please avoid abbreviations here.

modules/ding_frontend/ding_frontend.admin.inc
koykov: This isn't necessary. I used the same variables which used in content types editing forms. Therefore they already exported through Features.

modules/ding_frontend/ding_frontend.admin.inc
koykov: Done.

modules/ding_frontend/ding_frontend.admin.inc
kasperg: I think we misunderstand each other.

As I understand this change results in the following use:

1. By default events are promoted to frontpage. News are not. [This](https://github.com/ding2/ding2/blob/master/modules/ding_event/ding_event.strongarm.inc#L139-L147) [is](https://github.com/ding2/ding2/blob/master/modules/ding_news/ding_news.strongarm.inc) reflected in Features.
2. A local administrator wants to make news promoted to frontpage by default as well.
3. She uses the settings form provided here. This default values in the settings form is that *Events* are ticket off. This is derived from the `ding_frontend_fe_promote_types` variable. The exported value of this allows frontpage promotion of events - not news.
4. She submits the form with both *Events* and *News* ticked off.
5. Drupal adds the `promote` option to the `node_options_ding_news` variable. `node_options_ding_event` already has it. The value of the `ding_frontend_fe_promote_types` variable is not set.
6. When she revisits the form only *Events* will be ticked off. **The default value is derived from the variable `ding_frontend_fe_promote_types` which has not been updated in step 5. She expected both *Events* and *News* to be ticked off**.
7. When a new version of DDB CMS is deployed system administrators will revert all features. **This includes `ding_frontend_fe_promote_types` (which is not changed), `node_options_ding_news` and `node_options_ding_events`. This means that news will no longer be promoted to frontpage by default**
8. The local administrator can go back to step 1.

My suggestion:

1. By default events are promoted to frontpage. News are not. This is reflected by setting variables `node_options_ding_event` and `node_options_ding_news` accordingly in `ding_event_install()` and `ding_news_install()`.
2. A local administrator wants to make news promoted to frontpage by default as well.
3. She uses the settings form provided here. This default values in the settings form is that *Events* are ticket off. This is derived by inspecting the value of the `node_options_ding_event` and `node_options_ding_news` variables.
4. She submits the form with both *Events* and *News* ticked off.
5. Drupal adds the `promote` option to the `node_options_ding_news` variable. `node_options_ding_event` already has it.
6. When she revisits the form both *News* and *Events* will be ticked off as **the default value is derived by inspecting the value of the `node_options_ding_event` and `node_options_ding_news` variables**.
7. When a new version of DDB CMS is deployed system administrators will revert all features. **This does not include `node_options_ding_news` and `node_options_ding_events` as they are not included in Features**.
8. The local administrator does not have to do anything in this regard.

kasperg:
@koykov Thanks for the update. There is still one issue where I think we are misunderstanding each other.

modules/ding_frontend/ding_frontend.admin.inc
```diff
@@ -0,0 +1,60 @@
...
+function ding_frontend_settings($form, &$form_state) {
+  $types = array(
+    'ding_news' => t('News'),
+    'ding_event' => t('Event'),
+  );
+  $form['ding_frontend_fe_promote_types'] = array(
+    '#title' => t('Promote types to Frontpage'),
+    '#type' => 'checkboxes',
+    '#options' => $types,
+    '#default_value' => variable_get('ding_frontend_fe_promote_types', array('ding_event' => 'ding_event')),
```
kasperg: The variable `ding_frontend_fe_promote_types` is never set. I do not think it is needed either.

As mentioned previously the default value in this form element should be retrieved from the value of the existing `node_options_*` variables.

modules/ding_event/ding_event.module
```diff
@@ -108,6 +108,11 @@ function theme_date_time_seperate_lines($vars) {
  */
 function ding_event_post_features_revert() {
   ding_event_install_menu_position('revert');
+  if (module_exists('ding_frontpage')) {
+    module_load_include('inc', 'ding_frontpage', 'ding_frontpage.admin');
+    $custom_promote_frontpage = variable_get('ding_frontend_fe_promote_types', array('ding_event' => 'ding_event'));
+    ding_frontend_set_promote_configuration($custom_promote_frontpage);
+  }
```
kasperg: The variable ding_frontend_fe_promote_types is never set. Instead we rely on the existing `node_options_*` variables. Consequently this code should not be needed.

modules/ding_news/ding_news.module
```diff
@@ -21,6 +21,11 @@ function ding_news_entity_info_alter(&$entity_info) {
  */
 function ding_news_post_features_revert() {
   ding_news_install_menu_position('revert');
+  if (module_exists('ding_frontpage')) {
+    module_load_include('inc', 'ding_frontpage', 'ding_frontpage.admin');
+    $custom_promote_frontpage = variable_get('ding_frontend_fe_promote_types', array());
+    ding_frontend_set_promote_configuration($custom_promote_frontpage);
+  }
```
kasperg: The variable ding_frontend_fe_promote_types is never set. Instead we rely on the existing `node_options_*` variables. Consequently this code should not be needed.

modules/ding_frontend/ding_frontend.admin.inc
makcumm: We can't use `node_options_*` because they imported in feautres, so next feature revert will reset them. Current approach consider that local admin will set `ding_frontend_fe_promote_types` one time, and in next update of site and feature revert, `node_options_*` values will be set from `ding_frontend_fe_promote_types` which saved in  database of updated site.

kasperg: I have not had a chance to finish the review yet but in [one of my earlier comments](https://github.com/ding2/ding2/pull/74#discussion_r46824193) I suggested:

> removing the variables from the corresponding features and instead setting their defaults in a hook_install() in their respective modules

guddo: Let us know when you finish the review, then we'll change this.

kasperg: I re-reviewed the code and that is my only remaining comment.

In the current state the form does not seem to work. Changing the configuration does not change the default value. This is caused by never setting the variable `ding_frontend_fe_promote_types`. I do not think it is needed either.

As mentioned previously the default value in this form element should be retrieved from the value of the existing `node_options` variables. To avoid overriding configuration on deployment we should remove  the `node_options` variables from features and set their default values on install instead.

kasperg:
Scrutinizer reports 1898 issues fixed and 1900 introduced. That suggests that this PR has introduced 2 issues. It is hard to tell where they are so unless Scrutinizer fixes this in subsequent checks this can be ignored.

cableman:
Reviewed and awaiting next release

PR #77, opened 2015-11-30
Issue #333 by Inlead: Remove library linkage requirement for events.

guddo:
http://platform.dandigbib.org/issues/333


kasperg:
Hard to review changed configuration only. I assume that it works as expected. Reviewed and approved.

Thanks!

PR #84, opened 2015-12-03
Issue #1540 by xendk: Don't log username of failed logins.

xendk:
As a bonus, it deletes previous logged failed attempts.


cableman:
Reviewed and approved.

PR #85, opened 2015-12-04
Issue #1517 by bibchi: Handle added or removed facets

bibchi:
Remove ALL facets from the eKurser sidebar. Unless the facet "subject"
is unavailable for building a custom menu. In that case, show all
available facets instead.


cableman:
Code looks fine, just need to update your comment message :-) The first one is correct but the other needs to be prefixed with "Issue #1517 by bibchi: "

You can use "git rebase -i 3b748ae" an change "pick" in the lines to reword (r I think it is) and then save. Then git will ask for new commit messages for those.

cableman:
Reviewed and approved

PR #87, opened 2015-12-07
Revert "Issue #333 by Inlead: Remove library linkage requirement for events."

kasperg:
Reverts ding2/ding2#77


kasperg:
Reverted as #333 is not part of the next release.

PR #89, opened 2015-12-07
Hotfix by Inlead: Handle second variant of data's structure.

guddo:
http://platform.dandigbib.org/issues/1544

### DIFF
--- a/modules/bpi/bpi.push.inc
+++ b/modules/bpi/bpi.push.inc
@@ -154,8 +154,10 @@ function bpi_convert_to_bpi($node, $category, $audience, $with_images = FALSE, $
   // Find the references to the ting date well.
   $materials_drupal = array();
   if ($with_refs) {
+    // Function field_get_items returns data with 2 different structure, depending from stage when it was called.
     $items = field_get_items('node', $node, variable_get('bpi_field_materials', ''));
     foreach ($items as $item) {
+      // First case, we have completly saved node, when it returns ding_object_id.
       if (!empty($item['ting_object_id'])) {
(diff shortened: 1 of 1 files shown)


modules/bpi/bpi.push.inc
```diff
@@ -164,6 +164,17 @@ function bpi_convert_to_bpi($node, $category, $audience, $with_images = FALSE, $
           $materials_drupal[] = $id;
         }
       }
+      elseif(isset($item['value'])) {
```
kasperg: A comment explaining in which situations this occurs would be good.

modules/bpi/bpi.push.inc
```diff
@@ -164,6 +164,17 @@ function bpi_convert_to_bpi($node, $category, $audience, $with_images = FALSE, $
...
+        $endpoints = $item['value']->endpoints;
+        $data = array_filter($endpoints[LANGUAGE_NONE], function ($i) {
+          return $i['entity_type'] == 'ting_object';
+        });
+        $ting_data = current($data);
+        $ting_object = entity_load_single('ting_object', $ting_data['entity_id']);
```
kasperg: I would appreciate some defensive coding here.

- If the `$data` array is empty `$current` will return `false`.
- If the ting object entity with the provided id does not exist then `$ting_object` will be `false`.

kasperg:
Gitte has provided an issue for this: http://platform.dandigbib.org/issues/1544.

It would be good if you could update the commit message accordingly.

kasperg:
Reviewew and approved. Thanks!

PR #100, opened 2015-12-21
Issue #1267 by bartvig: Added script that sets focus on the username in the login form, also when the Login tab is clicked.

guddo:
When user visits the login directly or clicks the login tab, the username field is set to focused.
When the user clicks the up or down key, the form is blurred (unfocused).

http://platform.dandigbib.org/issues/1267

### DIFF
--- a/themes/ddbasic/scripts/ddbasic.login.js
+++ b/themes/ddbasic/scripts/ddbasic.login.js
@@ -1,0 +1,26 @@
+(function($) {
+    "use strict";
+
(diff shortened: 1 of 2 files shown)


cableman:
Reviewed and awaiting merge

kasperg:
Merged. Thanks!

guddo:
I think this should be placed in `ddbasic.common.js` and as `autoFocusInputs()` and cover both login and search forms.

PR #108, opened 2016-01-09
Issue #1342 by holt83: Fix naming of Ting oEmbed module

holt83:
The naming of the oEmbed module in the .info file, makes it hard to find in module lists:

http://platform.dandigbib.org/issues/1342

### DIFF
--- a/modules/ting_oembed/ting_oembed.info
+++ b/modules/ting_oembed/ting_oembed.info
@@ -1,5 +1,5 @@
-name = Add oEmbed support to Ting objects
-description = Use oEmbed module to add oEmbed support for displaying Ting objects
+name = Ting oEmbed
+description = Adds oEmbed support for displaying Ting objects
 package = Ding!
 core = 7.x
 version = 7.x-1.x


kasperg:
Thanks for this @holt83. I agree that this should be changed.

Could you please reformat the subject line for the commit message to follow our standards. It should be

`Issue #1342 by holt83: Fix naming of Ting oEmbed module`

holt83:
Fixed

kasperg:
I am sorry about complicating things but I will have to ask you to create a new ticket for this. Changing the name requires [changing the documentatio](http://platform.dandigbib.org/projects/ddb-cms/wiki/_Konfiguration_fra_A-Z#Konfiguration-af-eKurser-Kursusoversigt) which is "owned" by the business so we cannot piggyback on the existing ticket.

holt83:
Ah I see.. no problem. I will make a new ticket then :)

DDBdrift:
https://platform.dandigbib.org/issues/1342 | Resolved (tag version)

PR #112, opened 2016-01-15
Extend Alma with P2 consent functions

martinb14:


kasperg:
I assume this is no longer relevant as the changes are already included in #207.

PR #113, opened 2016-01-18
Issue #659: WAYF support

cableman:
This PR adds WAYF support using patched wayf_dk module. This is an optional module for the ding_wayf module which don't require SimpleSamlPHP.

This new module ding_wayf_dk communicates directly with wayf.dk and has the option to run against DBC "gate wayf". The latter have not been tested yet.


cableman:
The last 3 issue that scrutinizer fails on is in the exported code from features.

kasperg:
Can we exclude features from the analysis?

modules/alma/lib/AlmaClient/AlmaClient.class.php
```diff
@@ -980,6 +988,37 @@ public function remove_message_service($borr_card, $pin_code, $method, $type) {
...
   }
 
+  /**
+   * Create new user af alma.
+   *
+   * @param $cpr
+   * @param $pin_code
+   * @param $name
+   * @param $mail
+   * @param $branch
```
kasperg: Please add more elaborate documentation. Argument types and descriptions would be good.

modules/ding_wayf_dk/ding_wayf_dk.module
```diff
@@ -0,0 +1,59 @@
+<?php
+/**
+ * @file
+ * This module is a support module to share function between it's sub-modules.
```
kasperg: Is this DocBlock copy/pasted?

modules/ding_wayf_dk/ding_wayf_dk.module
```diff
@@ -0,0 +1,59 @@
...
+ */
+
+/**
+ * Implements hook_dk_wayf_dk_login_service_class().
```
kasperg: The DocBlock (starting with `dk_`) does not seem to match any available hook?

modules/ding_wayf_dk/includes/DingWayfGateway.class.inc
```diff
@@ -0,0 +1,118 @@
...
+ * Simple class for communication with the SimpleSAMLphp DBC gateway.
+ */
+
+namespace WAYF;
+use WAYF\SPortoException;
+
+/**
+ * DingWayfGateway class
+ */
+class DingWayfGateway implements WayfServiceInterface {
```
kasperg: Where is this interface defined?

modules/ding_wayf_dk/modules/ding_wayf_dk_create_user/ding_wayf_dk_create_user.info
```diff
@@ -0,0 +1,6 @@
+name = Ding WAYF (DK) create uers
+description = Implementation of user creation using (wayf.dk).
+core = 7.x
+package = Ding!
+dependencies[] = ding_wayf_dk
+dependencies[] = ding_wayf_dk_login
```
kasperg: `ding_wayf_dk_login` already depends on `ding_wayf_dk`. Is there a reason to return it twice?

modules/ding_wayf_dk/includes/DingWayfGateway.class.inc
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  /**
+   * Generate sp metadata based on configuration.
+   *
+   * @return string
+   *   Message that no metadata is provided by this service class.
+   */
+  public function getMetadata() {
+    return 'No metadata available for the current service class.';
```
kasperg: I do not know when this method is called. Could/should we return information about the configuration here?

modules/ding_wayf_dk/modules/ding_wayf_dk_create_user/ding_wayf_dk_create_user.module
```diff
@@ -0,0 +1,397 @@
...
+ * Handles user creation in the provider with information from WAYF login.
+ */
+
+/**
+ * Implements hook_permission().
+ */
+function ding_wayf_dk_create_user_permission() {
+  return array(
+    'administer wayf user creation' => array(
+      'title' => t('Administer WAYF (user creation)'),
```
kasperg: Why the `()`?

modules/ding_wayf_dk/modules/ding_wayf_dk_create_user/ding_wayf_dk_create_user.module
```diff
@@ -0,0 +1,397 @@
...
+}
+
+/**
+ * Implements hook_wayf_dk_login_authenticate_result_alter().
+ *
+ * Store the schacPersonalUniqueID in session to use later when send user create
+ * to the provider.
+ */
+function ding_wayf_dk_create_user_wayf_dk_login_authenticate_result_alter($result) {
+  if (isset($_SESSION['ding_wayf_dk_login_destination']) && $_SESSION['ding_wayf_dk_login_destination'] == 'wayf/create-user/form') {
```
kasperg: A commet about why this check is needed would be good.

modules/ding_wayf_dk/modules/ding_wayf_dk_create_user/ding_wayf_dk_create_user.module
```diff
@@ -0,0 +1,397 @@
...
+    // Extract unique id form the WAYF attributes (the CPR number).
+    preg_match('/\d{10}$/', reset($result['attributes']['schacPersonalUniqueID']), $unique_id);
+    $unique_id = reset($unique_id);
+
+    // Check if age limit should be set.
+    $limit = variable_get('ding_wayf_user_create_age_limit', '');
+    if (!empty($limit) && is_numeric($limit)) {
```
kasperg: Validation in the settings form should ensure that $limit is numeric.

modules/ding_wayf_dk/modules/ding_wayf_dk_create_user/ding_wayf_dk_create_user.module
```diff
@@ -0,0 +1,397 @@
...
+
+      // Magic function to get the 1900, 1800 or 2000 prefixes for year.
```
kasperg: 1800-prefix. Really?

modules/ding_wayf_dk/modules/ding_wayf_dk_create_user/ding_wayf_dk_create_user.module
```diff
@@ -0,0 +1,397 @@
...
+
+      // Create date time object to use it's diff method.
+      $from = new DateTime();
+      $from->setDate($year, substr($unique_id, 2, 2), substr($unique_id, 0, 2));
+      $to = new DateTime('tomorrow');
+
+      // Check if limit have been reached.
+      if ($from->diff($to)->y < $limit) {
+        // The age limit has to kick in.
+        drupal_set_message('You are not allowed to automatically be created because of the age limit. Please go to your local library to registry.');
```
kasperg: I think it would be good to include the configured the age limit in the message

modules/ding_wayf_dk/modules/ding_wayf_dk_create_user/ding_wayf_dk_create_user.module
```diff
@@ -0,0 +1,397 @@
...
+        case 7:
+        case 8:
+          $year = $year > 57 ? '18' . $year : '20' . $year;
+          break;
+      }
+
+      // Create date time object to use it's diff method.
+      $from = new DateTime();
+      $from->setDate($year, substr($unique_id, 2, 2), substr($unique_id, 0, 2));
+      $to = new DateTime('tomorrow');
```
kasperg: Why tomorrow instead of now?

modules/ding_wayf_dk/modules/ding_wayf_dk_create_user/ding_wayf_dk_create_user.module
```diff
@@ -0,0 +1,397 @@
...
+  $form['user'] = array(
+    '#type' => 'fieldset',
+    '#title' => t('Create user'),
+    '#group' => 'tabs',
+  );
+
+  $form['user']['ding_wayf_user_create_age_limit'] = array(
+    '#type' => 'textfield',
+    '#title' => t('Age limit'),
+    '#description' => t('If set the CPR number will be used to set an age limit. If empty no limit exists.'),
```
kasperg: This description makes no sense to me.

modules/ding_wayf_dk/modules/ding_wayf_dk_create_user/ding_wayf_dk_create_user.module
```diff
@@ -0,0 +1,397 @@
...
+/**
+ * Menu callback for information page before user creation.
+ */
+function ding_wayf_dk_create_user_info_page() {
+  // Get texts from the configuration.
+  $values = variable_get('ding_wayf_user_create_before_info', array());
+
+  return array(
+    '#theme' => 'ding_wayf_dk_info_page',
+    '#title' => empty($values['title']) ? t('Not configured.') : $values['title'],
```
kasperg: Please provide a better default value. Having "Not configured" potentially exposed to end users is never a good idea.

modules/ding_wayf_dk/modules/ding_wayf_dk_create_user/ding_wayf_dk_create_user.module
```diff
@@ -0,0 +1,397 @@
...
+    '#content' => empty($values['content']) ? t('Please configure WAYF configuration to get content on this page under <a href="/admin/config/people/wayf">admin/config/people/wayf</a>.') : $values['content'],
```
kasperg: Please provide a better default value. Having "Not configured" potentially exposed to end users is never a good idea.

modules/ding_wayf_dk/modules/ding_wayf_dk_create_user/ding_wayf_dk_create_user.module
```diff
@@ -0,0 +1,397 @@
...
+    '#default_value' => variable_get('ding_wayf_user_create_age_limit', ''),
+  );
+
+  // Define before create info page.
+  $form['user']['ding_wayf_user_create_before_info'] = array(
+    '#type' => 'fieldset',
+    '#tree' => TRUE,
+    '#collapsible' => TRUE,
+    '#collapsed' => TRUE,
+    '#title' => t('User creating info pager (before)'),
```
kasperg: This title text makes no sense to me. Pager?

modules/ding_wayf_dk/includes/DingWayfGateway.class.inc
```diff
@@ -0,0 +1,118 @@
...
+      watchdog('ding_wayf', 'The gateway returned an error: %error', array('%error' => $error), WATCHDOG_ERROR);
+      throw new SPortoException('The gateway returned an error: ' . $error);
+    }
+
+    // The gateway returns attributes in the $_POST array, if authentication
+    // have gone well.
+    if (empty($_POST['eduPersonTargetedID'])) {
+
+      // Redirect the user by setting the location header.
+      header('Location:' . $this->config['gateway'] . '?returnUrl=' . $this->config['return']['login'] . '&idp=wayf');
```
kasperg: The comment states what is going on - by *why*?

If we try to authenticate and we do not have the expected data then we redirects to the gateway where the user must authenticate?

modules/ding_wayf_dk/includes/DingWayfGateway.class.inc
```diff
@@ -0,0 +1,118 @@
...
+      exit;
+    }
+    else {
+      if (!empty($_POST['eduPersonTargetedID'])) {
```
kasperg: This `if` statement does not make sense. We already check whether `$_POST['eduPersonTargetedID']` is empty on L59.

modules/ding_wayf_dk/modules/ding_wayf_dk_create_user/ding_wayf_dk_create_user.module
```diff
@@ -0,0 +1,397 @@
...
+    '#type' => 'textfield',
+    '#title' => t('Link title'),
+    '#default_value' => isset($defaults['link']) ? $defaults['link'] : '',
+  );
+
+  // Accepting text.
+  $form['user']['ding_wayf_user_create_accpect_text'] = array(
+    '#type' => 'textarea',
+    '#title' => t('Accept text'),
+    '#default_value' => variable_get('ding_wayf_user_create_accpect_text', ''),
```
kasperg: Typos here: `accpect`.

modules/ding_wayf_dk/modules/ding_wayf_dk_create_user/ding_wayf_dk_create_user.module
```diff
@@ -0,0 +1,397 @@
...
+    '#description' => t('The text that is display next to the user create accept check box.'),
+  );
+
+  // Define after create info page.
+  $form['user']['ding_wayf_user_create_after_info'] = array(
+    '#type' => 'fieldset',
+    '#tree' => TRUE,
+    '#collapsible' => TRUE,
+    '#collapsed' => TRUE,
+    '#title' => t('User creating info pager (after)'),
```
kasperg: This title does not make sense.

modules/ding_wayf_dk/modules/ding_wayf_dk_create_user/ding_wayf_dk_create_user.module
```diff
@@ -0,0 +1,397 @@
...
+    ),
+  );
+}
+
+/**
+ * Menu callback for information page after user creation.
+ */
+function ding_wayf_dk_create_user_notice() {
+  // Get texts from the configuration.
+  $values = variable_get('ding_wayf_user_create_after_info', array());
```
kasperg: Naming seems inconsistent here. Some places we use notice - some places create after. Could this be made more consistent?

modules/ding_wayf_dk/modules/ding_wayf_dk_create_user/ding_wayf_dk_create_user.module
```diff
@@ -0,0 +1,397 @@
...
+
+  return array(
+    '#theme' => 'ding_wayf_dk_info_page',
+    '#title' => empty($values['title']) ? t('Not configured.') : $values['title'],
+    '#content' => empty($values['content']) ? t('Please configure WAYF configuration to get content on this page under <a href="/admin/config/people/wayf">admin/config/people/wayf</a>.') : $values['content'],
```
kasperg: No admin instructions in end user texts. Please provide som sane defaults.

modules/ding_wayf_dk/modules/ding_wayf_dk_create_user/ding_wayf_dk_create_user.module
```diff
@@ -0,0 +1,397 @@
...
+    // Send user to front page.
+    drupal_goto('<front>');
+  }
+
+  // This is a little hack to get around the panels design requirements of the
+  // ddbasic theme.
+  $form = array(
+    '#prefix' => '<div class="default-layout"><div class="layout-wrapper">',
+    '#suffix' => '<br /></div></div>',
+  );
```
kasperg: I image this could be a problem that might occour multiple times. Could we extract it into a theme function and use [`#theme_wrappers`](https://api.drupal.org/api/drupal/developer%21topics%21forms_api_reference.html/7#theme_wrappers)?

modules/ding_wayf_dk/modules/ding_wayf_dk_create_user/ding_wayf_dk_create_user.module
```diff
@@ -0,0 +1,397 @@
...
+    '#type' => 'checkbox',
+    '#title' => t('I accept the user agreements'),
+    '#description' => variable_get('ding_wayf_user_create_accpect_text', ''),
+    '#required' => TRUE,
+  );
+
+  $form['name'] = array(
+    '#type' => 'textfield',
+    '#title' => t('Name'),
+    '#description' => t('You full name.'),
```
kasperg: Typo: Your.

modules/ding_wayf_dk/modules/ding_wayf_dk_create_user/ding_wayf_dk_create_user.module
```diff
@@ -0,0 +1,397 @@
...
+    '#maxlength' => 100,
```
kasperg: Why this `maxlength`? If it is caused by other Drupal limitations then please add a comment stating why.

modules/ding_wayf_dk/modules/ding_wayf_dk_create_user/ding_wayf_dk_create_user.module
```diff
@@ -0,0 +1,397 @@
...
+    '#required' => TRUE,
+    '#attributes' => array(
+      'autocomplete' => 'off',
+    ),
```
kasperg: Why is `'autocomplete' => 'off'` needed?

modules/ding_wayf_dk/modules/ding_wayf_dk_create_user/ding_wayf_dk_create_user.module
```diff
@@ -0,0 +1,397 @@
...
+  );
+
+  // Get branch options form the active provider.
+  global $user;
+  $provider = _ding_provider_get_provider('reservation');
+  $options = ding_provider_invoke('reservation', 'options', $user);
+  $form['branch'] = $options[$provider['module'] . '_preferred_branch'];
```
kasperg: Detection of preferred branch leaves much to be desired. Also a problem [here](https://github.com/ding2/ding2/blob/master/modules/ding_reservation/ding_reservation.module#L187).

modules/ding_wayf_dk/modules/ding_wayf_dk_create_user/ding_wayf_dk_create_user.module
```diff
@@ -0,0 +1,397 @@
...
+    '#type' => 'submit',
+    '#value' => t('Create me'),
+  );
+
+  $link = '/wayf/logout';
+  if (user_is_logged_in()) {
+    $link = '/user/logout';
+  }
+  $form['cancel'] = array(
+    '#markup' => '<a class="ding-wayf--cancel" href="' . $link . '">' . t('Cancel') . '</a>',
```
kasperg: Why not use a `l()` here?

modules/ding_wayf_dk/modules/ding_wayf_dk_create_user/ding_wayf_dk_create_user.module
```diff
@@ -0,0 +1,397 @@
...
+
+  $form['submit'] = array(
+    '#type' => 'submit',
+    '#value' => t('Create me'),
+  );
+
+  $link = '/wayf/logout';
+  if (user_is_logged_in()) {
+    $link = '/user/logout';
+  }
```
kasperg: Why is this necessary? How can the user be logged in to WAYF and Drupal but not have a user account?

modules/ding_wayf_dk/modules/ding_wayf_dk_create_user/ding_wayf_dk_create_user.module
```diff
@@ -0,0 +1,397 @@
...
+      if ($from->diff($to)->y < $limit) {
+        // The age limit has to kick in.
+        drupal_set_message('You are not allowed to automatically be created because of the age limit. Please go to your local library to registry.');
+
+        // Redirect to front page.
+        drupal_goto('<front>');
+      }
+    }
+
+    $_SESSION['ding_wayf_dk_create_user'] = $unique_id;
```
kasperg: `ding_wayf_dk_create_user` is an odd name for a value which contains the id for a user. How about `ding_wayf_dk_create_user_id`,  `ding_wayf_dk_create_user_cpr` or the like?

modules/ding_wayf_dk/modules/ding_wayf_dk_create_user/ding_wayf_dk_create_user.module
```diff
@@ -0,0 +1,397 @@
...
+
+  if (isset($_SESSION['ding_wayf_dk_create_user'])) {
+    unset($_SESSION['ding_wayf_dk_create_user']);
+  }
+}
+
+/**
+ * User creation form submit handler.
+ */
+function ding_wayf_dk_create_user_form_sbumit($form, &$form_state) {
```
kasperg: Typo here. That should also ensure that the submit handler is automatically assigned to the form.

modules/ding_wayf_dk/modules/ding_wayf_dk_create_user/ding_wayf_dk_create_user.module
```diff
@@ -0,0 +1,397 @@
...
+
+  return $form;
+}
+
+/**
+ * Implements hook_wayf_dk_login_pre_logout().
+ *
+ * Ensures that session is cleaned up.
+ */
+function ding_wayf_dk_create_user_wayf_dk_login_pre_logout() {
```
kasperg: This hook is placed between the form and its submit handler. That seems odd.

modules/ding_wayf_dk/modules/ding_wayf_dk_create_user/ding_wayf_dk_create_user.module
```diff
@@ -0,0 +1,397 @@
...
+    unset($_SESSION['ding_wayf_dk_create_user']);
+  }
+}
+
+/**
+ * User creation form submit handler.
+ */
+function ding_wayf_dk_create_user_form_sbumit($form, &$form_state) {
+  $values = $form_state['values'];
+  $cpr = $_SESSION['ding_wayf_dk_create_user'];
```
kasperg: As I see it there is no check to ensure that `$_SESSION['ding_wayf_dk_create_user']` is set when rendering the form and there is no check to see if it is set here either.

How about adding the check during form creation and if set moving it to a form value? This reduces the lifetime of session values which can be hard to track.

modules/ding_wayf_dk/modules/ding_wayf_dk_create_user/ding_wayf_dk_create_user.module
```diff
@@ -0,0 +1,397 @@
...
+    ding_provider_invoke('user', 'create', $cpr, $values['pin'], $values['name'], $values['mail'], $values['branch']);
+
+    // Login the user and ensure he/her is redirect to notice page.
+    $_REQUEST['destination'] = 'wayf/create-user/notice';
+    ding_wayf_dk_login_wayf_dk_login_create_user(array(
+      'schacPersonalUniqueID' => array('urn:mace:terena.org:schac:personalUniqueID:dk:CPR:' . $cpr),
+    ));
+
+    // Remove request fields for the login form.
+    drupal_get_messages(NULL, TRUE);
```
kasperg: Request fields?

modules/ding_wayf_dk/modules/ding_wayf_dk_create_user/ding_wayf_dk_create_user.module
```diff
@@ -0,0 +1,397 @@
...
+  }
+  catch (Exception $exception) {
+    drupal_set_message(t('You already have an user account at the library. So we have just logged you in.'));
```
kasperg: Does *any* exception thrown mean that the user already exist? That seems very suspicous.

modules/ding_wayf_dk/modules/ding_wayf_dk_create_user/plugins/content_types/ding_wayf_dk_create_user.inc
```diff
@@ -0,0 +1,55 @@
...
+function ding_wayf_dk_login_create_user_content_type_edit_form($form, &$form_state) {
+  return $form;
+}
+
+
+/**
+ * Submit function.
+ */
+function ding_wayf_dk_login_create_user_content_type_edit_form_submit(&$form, &$form_state) {
+  $form_state['conf']['icon'] = $form_state['values']['icon'];
```
kasperg: Why is this needed? I do not see any edit form with an `icon` entry?

modules/ding_wayf_dk/modules/ding_wayf_dk_login/ding_wayf_dk_login.install.php
```diff
@@ -0,0 +1,54 @@
...
+  $requirements = array();
+
+  // Ensure translations don't break at install time.
+  $t = get_t();
+
+  if ($phase == 'runtime') {
+    if (!variable_get('wayf_hash', FALSE)) {
+      $requirements['wayf_hash'] = array(
+        'value' => $t('WAYF not configured correctly'),
+        'description' => $t('WAYF is not properly configured, you need to set $conf[\'wayf_hash\'] in setttings.php.'),
```
kasperg: Typo: `setttings`.

modules/ding_wayf_dk/modules/ding_wayf_dk_login/ding_wayf_dk_login.install.php
```diff
@@ -0,0 +1,54 @@
...
+function ding_wayf_dk_login_requirements($phase) {
+  $requirements = array();
+
+  // Ensure translations don't break at install time.
+  $t = get_t();
+
+  if ($phase == 'runtime') {
+    if (!variable_get('wayf_hash', FALSE)) {
+      $requirements['wayf_hash'] = array(
+        'value' => $t('WAYF not configured correctly'),
```
kasperg: I would expect the value of the requirement to be `WAYF`, `Ding! WAYF` or something like that.

modules/ding_wayf_dk/modules/ding_wayf_dk_login/ding_wayf_dk_login.install.php
```diff
@@ -0,0 +1,54 @@
...
+    'wayf_dk_login' => 0,
+  ));
+
+  // Enable NemID.
+  variable_set('wayf_dk_login_organizations_active', array(
+    'sikker-adgang.dk' => 'sikker-adgang.dk',
+  ));
+
+  // Set default language.
+  variable_set('wayf_dk_login_organizations_name_language', 'da');
```
kasperg: Should this use [`language_default()`](https://api.drupal.org/api/drupal/includes!bootstrap.inc/function/language_default/7) instead of defaulting to `da`?

modules/ding_wayf_dk/modules/ding_wayf_dk_create_user/templates/ding-wayf-dk-info-page.tpl.php
```diff
@@ -0,0 +1,13 @@
+<div class="default-layout">
+  <div class="layout-wrapper">
+    <h2 class="pane-title"><?php print render($title); ?></h2>
+    <div class="pane-content">
```
kasperg: Perhaps a comment explaining the relationship with panels would be an idea here.

modules/ding_wayf_dk/modules/ding_wayf_dk_login/ding_wayf_dk_login.module
```diff
@@ -0,0 +1,255 @@
...
+/**
+ * Delimiter for extracting a part of a field.
+ */
+define('DING_WAYF_DK_LOGIN_FIELD_DELIMITER', ':');
+
+/**
+ * Implements hook_menu_alter().
+ *
+ * Allow users to logout of WAYF as they may be logged into WAYF but not the
+ * drupal site/library system.
```
kasperg: Nitpicking: This should probably be an inline comment in the function instead of part of the DocBlock. What if more changes are added to the hook?

modules/ding_wayf_dk/modules/ding_wayf_dk_login/ding_wayf_dk_login.module
```diff
@@ -0,0 +1,255 @@
...
+ */
+function ding_wayf_dk_login_menu_alter(&$items) {
+  $items['wayf/logout']['access callback'] = TRUE;
+}
+
+/**
+ * Implements hook_wayf_dk_login_create_user().
+ */
+function ding_wayf_dk_login_wayf_dk_login_create_user($attributes) {
```
kasperg: Why is this a part of the login module and not the create user module? As I see it the function deals with user creation and thus seems to fit within the scope of the create user module. It does not seem to be called elsewhere.

modules/ding_wayf_dk/modules/ding_wayf_dk_login/ding_wayf_dk_login.module
```diff
@@ -0,0 +1,255 @@
...
+  if (ding_provider_implements('wayf', 'login_credentials')) {
+    // Get credentials from the provider.
+    $credentials = ding_provider_invoke('wayf', 'login_credentials', ding_wayf_dk_login_get_credentials($attributes));
+
+    // Check if CPR mapping is activated.
+    if (variable_get('ding_wayf_dk_login_debug', 0)) {
+      $credentials['name'] = variable_get('ding_wayf_dk_login_debug_cpr', '');
+    }
```
kasperg: I do not understand what is going on here. The comment refers to "CPR mapping" and uses variables relation to debugging.

modules/ding_wayf_dk/modules/ding_wayf_dk_login/ding_wayf_dk_login.module
```diff
@@ -0,0 +1,255 @@
...
+
+    if (!empty($credentials)) {
+      // Set values to send with the login form.
+      $form_state = array(
+        'values' => $credentials,
+      );
+
+      // Submit login form.
+      drupal_form_submit('user_login_block', $form_state);
```
kasperg: Do we really have to do it this way? I would hope we could use a less crude API.

modules/ding_wayf_dk/modules/ding_wayf_dk_login/ding_wayf_dk_login.module
```diff
@@ -0,0 +1,255 @@
...
+      $errors = form_get_errors();
+      if (!empty($errors)) {
+        if (module_exists('ding_wayf_dk_create_user')) {
+          // Clean out message from the login form.
+          drupal_get_messages(NULL, TRUE);
+
+          drupal_set_message(t('You do not have an user account at the public library. If you wish to create an user account !user_url . You are still logged in at WAYF, so you can !logout to logout of WAYF.', array(
+            '!logout' => l(t('click her'), 'wayf/logout'),
+            '!user_url' => l(t('click her'), 'wayf/create-user/before-page'),
```
kasperg: Typos: "click here".

modules/ding_wayf_dk/modules/ding_wayf_dk_login/ding_wayf_dk_login.module
```diff
@@ -0,0 +1,255 @@
...
+ */
+function ding_wayf_dk_login_preprocess_wayf_dk_login(&$variables) {
+  $image = array(
+    '#theme' => 'image',
+    '#path' => drupal_get_path('module', 'ding_wayf_dk_login') . '/images/nemid_logo.png',
+    '#alt' => t('Login with NemID'),
+    '#title' => t('Login with NemID'),
+  );
+
+  $variables['login_url']['#text'] = t('Login with') . ' ' . drupal_render($image);
```
kasperg: I appreciate the alt and title here but it does not make sense in relation to the link text which also contains "login with".

modules/ding_wayf_dk/modules/ding_wayf_dk_login/ding_wayf_dk_login.module
```diff
@@ -0,0 +1,255 @@
...
+
+  drupal_goto($fallback_url);
+}
+
+/**
+ * Implements hook_form_wayf_dk_login__settings_form_alter().
+ *
+ * Adds debug option to map test WAYF CPR number to a provider valid one.
+ */
+function ding_wayf_dk_login_form_wayf_dk_login__settings_form_alter(&$form, &$form_state, $form_id) {
```
kasperg: Why does this integrate with the WAYF admin form instead of providing it's own? I am not necessarily against it - just asking.

modules/ding_wayf_dk/modules/ding_wayf_dk_login/ding_wayf_dk_login.module
```diff
@@ -0,0 +1,255 @@
...
+    '#type' => 'checkbox',
+    '#title' => t('Enable debug'),
+    '#default_value' => variable_get('ding_wayf_dk_login_debug', FALSE),
+    '#description' => t('Enable debug to do CPR mapping with WAYF test users.'),
+  );
+
+  $form['settings']['debug']['ding_wayf_dk_login_debug_cpr'] = array(
+    '#type' => 'password',
+    '#title' => t('CPR map'),
+    '#description' => t('Map student test CPR (0708741234).'),
```
kasperg: I do not understand this description. Could you elaborate it?

modules/ding_wayf_dk/modules/ding_wayf_dk_login/ding_wayf_dk_login.module
```diff
@@ -0,0 +1,255 @@
...
+function ding_wayf_dk_login_get_credentials(array $attributes) {
+  // Set default required attributes.
+  $required_attributes = array(
+    'schacPersonalUniqueID' => array(
+      'field' => 'CPR',
+      'authname' => TRUE,
+    ),
+  );
+
+  // Ask the provider about required attributes and give it a change to override
```
kasperg: change = chance?

modules/ding_wayf_dk/modules/ding_wayf_dk_login/ding_wayf_dk_login.module
```diff
@@ -0,0 +1,255 @@
...
+function ding_wayf_dk_login_get_required_attributes(array $attributes, array $required_attributes) {
+  $attribute_list = array();
+
+  if (isset($attributes) && is_array($required_attributes)) {
+    foreach ($required_attributes as $attribute => $value) {
+      if (isset($attributes[$attribute])) {
+        $attribute_list[$attribute] = ding_wayf_dk_login_extract_data($attributes[$attribute], $value['field']);
+      }
+    }
+
```
kasperg: Nitpicking: Superfluous whitespace here.

modules/ding_wayf_dk/modules/ding_wayf_dk_login/ding_wayf_dk_login.module
```diff
@@ -0,0 +1,255 @@
...
+/**
+ * Extract required attributes from available WAYF attributes.
+ *
+ * @param array $attributes
+ *   WAYF attributes.
+ * @param array $required_attributes
+ *   Required attributes.
+ *
+ * @return array
+ *   Will be empty if not all required attributes is present.
```
kasperg: This does not seem to hold true? If there are two required attributes and the attributes only contain one of them I cannot spot anywhere in the function which will lead to it returning an empty array.

modules/ding_wayf_dk/modules/ding_wayf_dk_login/ding_wayf_dk_login.module
```diff
@@ -0,0 +1,255 @@
...
+      }
+    }
+
+  }
+
+  return $attribute_list;
+}
+
+/**
+ * Extracts a field from a SAML attribute string.
```
kasperg: This seems rather odd - possibly caused by the data format. Could the DocBlock include an example?

themes/ddbasic/sass/module/_user.scss
```diff
@@ -471,21 +471,54 @@
...
+.pane-wayf-dk-login {
+  display: inline-block;
+
+  .ding-wayf--login-block {
+    background-color: #fff;
+
+    @include border-radius($button-border-radius);
+
+    a {
+      color: #000;
```
kasperg: Do we really want to use a static colour here?

themes/ddbasic/scripts/ddbasic.common.js
```diff
@@ -56,11 +56,25 @@
     });
   }
 
+  /**
+   * HACK: this is an hack that have to be fixed later.
```
kasperg: Please add a comment explaining why this hack is needed. This will help us understand when/how it can be fixed later on.

modules/ding_wayf_dk/modules/ding_wayf_dk_login/ding_wayf_dk_login.module
```diff
@@ -0,0 +1,255 @@
...
+}
+
+/**
+ * Implements hook_wayf_dk_login_authenticate_result_alter().
+ *
+ * Restore destination after all the WAYF redirects.
+ */
+function ding_wayf_dk_login_wayf_dk_login_authenticate_result_alter($result) {
+  if (isset($_SESSION['ding_wayf_dk_login_destination'])) {
+    $_REQUEST['destination'] = $_SESSION['ding_wayf_dk_login_destination'];
```
kasperg: Should we remove the destination from the session afterwards instead of waiting for the user to logout?

modules/ding_wayf_dk/includes/DingWayfGateway.class.inc
```diff
@@ -0,0 +1,118 @@
+<?php
+/**
+ * @file
+ * Simple class for communication with the SimpleSAMLphp DBC gateway.
```
kasperg: If this class deals with the SimpleSAMLphp DBC gateway then why is it included in this module which uses wayf.dk?

kasperg:
@cableman: Review complete.

There are multiple things here:

1. The Circle build fails because patches to `wayf_dk_login` does not apply. Please fix this.
2. The Scrutinizer build fails because generated code violates coding standards. We can merge despite this failure. Going forward we should exclude generated code from analysis. See #121.
3. There seems to be two issues at stake here. The relationship between these dimensions is unclear to me in both description and code. I would like to see this clarified.
  1. Whether we should integrate directly between wayf.dk and DDB CMS using the `wayf_dk_login` module or through SimpleSAMLphp
  2. Whether we should allow WAYF login only or login and user creation

modules/alma/lib/AlmaClient/AlmaClient.class.php
cableman: It have not been updated with comments

modules/ding_wayf_dk/ding_wayf_dk.module
cableman: No, this module file is used to make changes and share code across it's two sub-module in the modules folder (ding_wayf_dk_login and ding_wayf_dk_create_user).

It was an empty module file until support for gatewayf was requested by DDB.

modules/ding_wayf_dk/ding_wayf_dk.module
cableman: That was an typo :-)

modules/ding_wayf_dk/includes/DingWayfGateway.class.inc
cableman: It's defined inside the wayf_dk module as a way to ensure that the functions required in services class are implemented in the class.

It's part of the patch https://www.drupal.org/files/issues/allow_other_libraries-2648680-1.diff to the module on DO. Which is patched in the make file.

modules/ding_wayf_dk/modules/ding_wayf_dk_create_user/ding_wayf_dk_create_user.info
cableman: Removed the dependency ´

modules/ding_wayf_dk/includes/DingWayfGateway.class.inc
cableman: This function is not used when using gateway WAYF, but it's required by the WayfServiceInterface. It's normally used in a menu callback to update metadata at wayf.dk, so you don't have to copy&past when using wayf.dk directly.

There is also a page in the admin UI where you can see the metadata, so this will give correct content on that page. Namely that the current selected service class (if gateway wayf) does not have any metadata (as this is handed in the gataway by DBC).

cableman: Added comment about this to the documentation block

modules/ding_wayf_dk/modules/ding_wayf_dk_create_user/ding_wayf_dk_create_user.module
cableman: Removed the () in the title.

modules/ding_wayf_dk/modules/ding_wayf_dk_create_user/ding_wayf_dk_create_user.module
cableman: The function is now better documented about why it's need and how it's used.

modules/ding_wayf_dk/modules/ding_wayf_dk_create_user/ding_wayf_dk_create_user.module
cableman: Yes, it's an standard function to detect CPR numbers, so for completeness.

modules/ding_wayf_dk/modules/ding_wayf_dk_create_user/ding_wayf_dk_create_user.module
cableman: It's due to the time of day, when we created this months ago we did a lot of testing and this was the correct why to do it.

We don't have the test code any more and can't remember the why, but it's something about the day they turn (the limit) they should be able to be created.

modules/ding_wayf_dk/modules/ding_wayf_dk_create_user/ding_wayf_dk_create_user.module
cableman: Age limit added to the string and it's now wrapped in t().

modules/ding_wayf_dk/modules/ding_wayf_dk_create_user/ding_wayf_dk_create_user.module
cableman: I see, but it's hard to give any good defaults here... for aakb.dk it's a legal text and links to custom pages with the library rules etc.

modules/ding_wayf_dk/modules/ding_wayf_dk_create_user/ding_wayf_dk_create_user.module
cableman: Description have been corrected

modules/ding_wayf_dk/modules/ding_wayf_dk_create_user/ding_wayf_dk_create_user.module
cableman: It should be "page", tried to reformulate the title and description

modules/ding_wayf_dk/modules/ding_wayf_dk_create_user/ding_wayf_dk_create_user.module
cableman: It should be "page", tried to reformulate the title and description

modules/ding_wayf_dk/includes/DingWayfGateway.class.inc
cableman: Comment updated

modules/ding_wayf_dk/includes/DingWayfGateway.class.inc
cableman: If removed, it's copied from the old WAYF module

modules/ding_wayf_dk/modules/ding_wayf_dk_create_user/ding_wayf_dk_create_user.module
cableman: Renamed and update function added

modules/ding_wayf_dk/modules/ding_wayf_dk_create_user/ding_wayf_dk_create_user.module
cableman: I see, but it's hard to give any good defaults here... for aakb.dk it's a legal text and links to custom pages with the library rules etc.

modules/ding_wayf_dk/modules/ding_wayf_dk_login/ding_wayf_dk_login.module
cableman: To have all related configuration in the same form. The admin page for WAYF is "tab" based.

modules/ding_wayf_dk/includes/DingWayfGateway.class.inc
cableman: This module is able to communicate with wayf.dk directly or with the changes requested by DDB, the gateway at DBC that them communicates with DBC.

I'm not a fan of this extra layer of complexity, but if this should not be part WAYF module it would require a whole new module that has noting to do with wayf.dk.

cableman:
New PR on the way, which closes this.

PR #114, opened 2016-01-18
Issue #534 : Re-rolled patch to fix interest period in alma

cableman:


DDBdrift:
https://platform.dandigbib.org/issues/534 | Resolved (tag version)

PR #117, opened 2016-01-21
Feature/1343 alter pincode

JrgenGNielsen:


modules/ding_user/ding_user.module
```diff
@@ -267,6 +267,15 @@ function ding_user_form_profile2_form_alter(&$form, &$form_state) {
         '#type' => 'fieldset',
         '#title' => t('Change pincode'),
       );
+      if (!module_exists('ding_wayf')) {
```
cableman: This will only check if the module is activated, but that do not indicate if the user have used WAYF to login. The old pas code should be required also when WAYF module exists, but user logged in without using WAYF.

PR #118, opened 2016-01-21
Feature/774 creator materialtype in record list

JrgenGNielsen:


cableman:
Could you rebase the code, so the first two commits (about makefiles) is not part of the PR!

PR #119, opened 2016-01-22
Issue #1517 by bibchi: Get subject-terms from tingclient instead of t…

bibchi:
…he facetbrowser

Query the subject-terms for the ekurser-menu directly from the
tingclient, instead of getting them from the facetbrowser-form. Thus
removing the dependency on 'facet.subject' in the websites
facet-settings.


JrgenGNielsen:
reviewet og godkendt

PR #120, opened 2016-01-25
Feature/431 frontpage event limit

JrgenGNielsen:


modules/ding_event/ding_event.admin.inc
```diff
@@ -50,12 +70,16 @@ function ding_event_admin_currency_submit($form, &$form_state) {
...
+ *   A keyed array containing the current state of the form.
  */
-function ding_event_permission() {
-  return array(
-    'administer event settings' => array(
-      'title' => t('Administer event settings')
-    ),
-  );
+function ding_event_admin_validate(array $form, array &$form_state) {
+  if (!ctype_digit($form_state['values']['ding_event_frontend_items_per_page'])) {
```
kasperg: It seems like this could be replaced with setting [`#element_validate = array('element_validate_integer_positive')`](https://api.drupal.org/api/drupal/includes%21form.inc/function/element_validate_integer_positive/7) on the element.

modules/ding_event/ding_event.module
```diff
@@ -23,11 +23,11 @@ function ding_event_menu() {
     'file path' => drupal_get_path('module', 'system'),
   );
 
-  $items['admin/config/ding_event/currency'] = array(
-    'title' => 'Event currency',
-    'description' => 'Select which currency to show on events.',
+  $items['admin/config/ding_event/settings'] = array(
+    'title' => 'Ding event settings',
+    'description' => 'Administer event settings.',
```
kasperg: :+1:

modules/ding_event/ding_event.module
```diff
@@ -227,3 +227,24 @@ function ding_event_preprocess_views_view(&$vars) {
...
+  );
+}
+
+/**
+ * Implements hook_views_pre_build().
+ */
+function ding_event_views_pre_build(&$view) {
+  if ($view->name == "ding_event" && $view->current_display == 'ding_event_list_frontpage') {
+    $view_items_per_page = $view->display['ding_event_list_frontpage']->display_options['pager']['options']['items_per_page'];
+    $view->items_per_page = variable_get('ding_event_frontend_items_per_page', $view_items_per_page);
```
kasperg: Will this force the exported view / feature to be listed as overridden when `ding_event_frontend_items_per_page` is set? If yes: Is there any way to avoid that from happening while keeping the desired result?

modules/ding_event/ding_event.admin.inc
```diff
@@ -6,27 +6,44 @@
...
     '#type' => 'radios',
     '#title' => t('Currency'),
     '#options' => drupal_map_assoc(array('kr', '€')),
     '#default_value' => $currency,
+    '#description' => t("Select which currency to use on events"),
+  );
+
+  $view = views_get_view('ding_event');
+  $view_items_per_page = $view->display['ding_event_list_frontpage']->display_options['pager']['options']['items_per_page'];
+  $default_items_per_page = variable_get('ding_event_frontend_items_per_page', $view_items_per_page);
```
kasperg: The naming of this variable is a bit misleading - `frontend_items_per_page` could refer to any page in the frontend. How about `ding_event_frontpage_items_per_page` or `ding_event_frontpage_items`?

modules/ding_event/ding_event.admin.inc
```diff
@@ -6,27 +6,44 @@
...
  * @return mixed
  */
-function ding_event_admin_currency_setting_form($form, &$form_state) {
+function ding_event_admin_setting_form(array $form, array &$form_state) {
   $currency = variable_get('ding_event_currency_type', 'Kr');
 
-  $form['ding_event_currency'] = array(
+  $form['ding_event'] = array(
     '#type' => 'fieldset',
     '#title' => t('Event currency'),
```
kasperg: This fieldset currently spans the entire form (currency + frontpage items) but the title no longer fits all the child elements.

I suggest one of two options:

1. Drop the fieldset
2. Add another fieldset for frontpage items

kasperg:
@JrgenGNielsen: Review complete. I have a couple of minor comments and a question.

Also: Could you please ¶rebase your pull request, remove e418378 and 12fb166 and force push the branch](https://robots.thoughtbot.com/git-interactive-rebase-squash-amend-rewriting-history). These two commits are not relevant in the shared history.

modules/ding_event/ding_event.module
JrgenGNielsen: It don't affect the exported view :-)

modules/ding_event/ding_event.admin.inc
JrgenGNielsen: But the variable name is already 'ding_event_frontend_items_per_page'??

modules/ding_event/ding_event.module
kasperg: :+1:

modules/ding_event/ding_event.admin.inc
kasperg: `frontend` is not the same as `frontpage`. The current name indicates that it spans the entire frontend but it only affects the front page.

JrgenGNielsen: ah, right - I read that as you objecting to naming the variable "frontend_items_per_page", and suggesting it to be prefixed "ding_event_"

PR #106, opened 2016-01-06
Add code and contribution guidelines

kasperg:
This moves [our code guidelines](http://ting.dk/wiki/ding-code-guidelines) from ting.dk to the codebase where they can be discussed and updated by developers.

This also updates the code guidelines with the changes to the repository structure, introduction of automated analysis.

Finally contribution guidelines are added with requirements and best practices for developers when submitting pull requests.

PR #121, opened 2016-01-26
Exclude generated Features code from Scrutinizer

kasperg:
Generated code does not comply with our coding standards and causes the
build to fail. It is not feasible for us to fix this so we exclude it
instead.

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -28,11 +28,15 @@ filter:
         - '*.panelizer.inc'
         - '*.strongarm.inc'
         - '*.views_default.inc'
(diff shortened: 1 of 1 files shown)


kasperg:
This also removes >1500 issues in existing Features files.

cableman:
Should there exist an issue for this? Mostly to have correct commit messages!

kasperg:
@cableman: In my opinion [there is no need to create an issue for changes that are not relevant to the business](https://github.com/ding2/ding2/pull/106/files#diff-6a3371457528722a734f3c51d9238c13R9).

PR #127, opened 2016-01-28
Enhancement #1516 by holt83: Introduce Metatag module and fix inconsistent page titles

holt83:
This PR introduces the Metatag module and uses it to fix the inconsistent page titles in DDB CMS.

This is important for SEO, bookmarks and accessibility.

At the same time, introducing the Metatag module, lays the groundwork for future metatag related additions to DDB CMS. This could be Facebook open-graph, Dublin Core or even Integration with Webtrends metatag.

For more info: http://platform.dandigbib.org/issues/1516


guddo:
`[current-page:pager][site:name]` - is it on purpose that there is no separator for this? And do we really want to show the current page number? I mean, who cares..

holt83:
It's a special token defined by the metatag module and it will insert the separator itself.

Since it doesn't do any harm if there's no paging, I thought we might as well include it. It's also included in all the titles by default when you install the module.

On most pages it will have no use, but on the search pages I actually think it's needed.

JrgenGNielsen:
reviewed and accepted

cableman:
Manual merged to resolve conflict

PR #133, opened 2016-02-02
Issue #465 by Inlead: Waiting time when no items available.

inlead:
http://platform.dandigbib.org/issues/465


kasperg:

The Scrutinizer fail is caused by a bug in that platform. We disregard this failure.

kasperg:
Merged. Thanks!

PR #137, opened 2016-02-08
Issue 469 by Pjohans (original Morten Bartvig): Add module to set loc…

JrgenGNielsen:
…al configuration

### DIFF
--- a/modules/ding_local_config/README.md
+++ b/modules/ding_local_config/README.md
@@ -1,0 +1,2 @@
+# ding_local_config
+Local configuration for Ding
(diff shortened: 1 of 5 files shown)


JrgenGNielsen:
More Scrutinizer error messages than I can fix - mostly error messages about incorrect line indentation.

cableman:
After some conversion in the core team, we have concluded that we agree on the approach (hook_entity_view) but would like it to be part of the module that have the original configuration in this case ding_ting_frontend.

Maybe as a tab under "admin/config/ting/settings" with own permissions.

holt83:
I'm curious as to why you would not use hook_ding_entity_view()?

hook_entity_view() seems to be called when viewing all entitites (nodes included). hook_ding_entity_view() on the other hand seems to only be called when viewing Ting objects, isn't that what we want?

JrgenGNielsen:
Moved to: https://github.com/ding2/ding2/pull/165

PR #101, opened 2015-12-21
Issue #1018 by JrgenGNielsen: Update expire feature.

guddo:
http://platform.dandigbib.org/issues/1018

PR #138, opened 2016-02-10
Issue #1018 by JrgenGNielsen: Update expire feature.

kasperg:
This superseeds #101 as a fix for http://platform.dandigbib.org/issues/1018. It tries to update to a non-existing version of the varnish module. It is the expire module which should be updated.


JrgenGNielsen:
approved

PR #139, opened 2016-02-11
Issue #1609 by xendk: Properly return false with zero cache lifetime

xendk:
If setting ting_cache_liftetime to zero for development, everything
breaks. Ting object pages are rendered, but the data is missing.

Somewhere someone is checking that ting_cache_get() returns '=== FALSE',
or something like that, and it happens to not return anything if the
lifetime is 0.

Simple fix.


JrgenGNielsen:
reviewed and accepted

PR #140, opened 2016-02-11
Issue #1610 by xendk: Make CTools argument return the proper object type

xendk:
The CTools argument handler returns a TingClientObject rather than the
expected TingEntity.

This makes it rather useless as contexts expect the latter.


JrgenGNielsen:
reviewed and accepted

PR #141, opened 2016-02-11
Issue #1611 by xendk: Implement custom labels for material types

xendk:
Allows for configuring the labels used for ting object types in the
frontend, so "lydbog (net)" can be displayed as "Lydbog", for instance.

### DIFF
--- a/modules/ting/ting.admin.inc
+++ b/modules/ting/ting.admin.inc
@@ -140,6 +140,7 @@ function ting_admin_ting_settings($form_state) {
  * Form builder; Configure custom names for ting types.
  *
  * @ingroup forms
+ *
  * @see system_settings_form()
  */
 function ting_admin_types_settings($form_state) {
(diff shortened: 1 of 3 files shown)


JrgenGNielsen:
reviewed and accepted

cableman:
This PR conflict with other changes in the same code lines and I'am not sure what is correct. Could you please take another look at the PR.

xendk:
Looking into it.

xendk:
OK, the conflict is with #175, but reading up on the issue, I'm not convinced that it fixes the issue. Following up in http://platform.dandigbib.org/issues/466

xendk:
Done, but I don't know what that means for this PR. If the other PR gets reverted, this should apply again, but if the other gets re-worked, I'll just have to rebase this yet again, if I rebase to the current version.

DDBdrift:
https://platform.dandigbib.org/issues/1611 | Rejected

PR #142, opened 2016-02-11
Issue 1617 by xendk: Add hooks to allow provider to add fields to login form

xendk:
Also makes the previously private ding_provider_function() public (it was just too useful to be private).


JrgenGNielsen:
reviewed and accepted

PR #143, opened 2016-02-12
Issue 1620 by xendk: Implement generic search field formatter

xendk:
Formats a field as a link to search for the value in the configured
search index.

See http://platform.dandigbib.org/issues/1620


JrgenGNielsen:
reviewed and accepted

PR #146, opened 2016-02-18
Issue #1626 by xendk: Don't load materials deleted in the datawell

xendk:
Ting entity controller shouldn't return objects when the well doesn't know it anymore.


guddo:
Can anyone drop a few words to why this has not been merged yet? /cc @kasperg @cableman

kasperg:
The business is having [a lengthy discussion about the issue](http://platform.dandigbib.org/issues/1626).

xendk:
Superseeded by #499 and #1621

PR #153, opened 2016-02-24
Issue 773 by Pjohans: Disable custom alias settings

pjohans:
This is a bit hacky. I tried to disable the custom alias checkbox, but that caused path auto functionality to fail (path became node/id instead of bibliotek/name). This solution simulates a disabled checkbox with a javascript.

### DIFF
--- a/modules/ding_library/ding_library.module
+++ b/modules/ding_library/ding_library.module
@@ -207,8 +207,7 @@ function ding_library_form_ding_library_node_form_alter(&$form, &$form_state) {
 function ding_library_disable_custom_alias($form, &$form_state) {
   if (isset($form['path']['pathauto'])) {
     $form['path']['pathauto']['#description'] = t('Custom alias has been disabled');
-    $form['path']['pathauto']['#attached']['js'][] =
-      drupal_get_path('module', 'ding_library') . '/ding_library.js';
+    $form['path']['pathauto']['#attached']['js'][] = drupal_get_path('module', 'ding_library') . '/ding_library.js';
 
     unset($form['path']['alias']);
   }
--- a/modules/ding_library/ding_library.module
+++ b/modules/ding_library/ding_library.module
@@ -191,7 +191,7 @@ function ding_library_ctools_plugin_directory($owner, $plugin_type) {
  * Implements hook_form_FORM_ID_alter().
  */
 function ding_library_form_ding_library_node_form_alter(&$form, &$form_state) {
-  $form['#after_build'][] = 'ding_library_remove_path_auto';
+  $form['#after_build'][] = 'ding_library_disable_custom_alias';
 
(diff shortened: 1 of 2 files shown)


xendk:
Rather than add more JS, wouldn't it be easier just to set the form elements '#disabled' ?

https://api.drupal.org/api/drupal/developer%21topics%21forms_api_reference.html/7#disabled

pjohans:
That was my first shot, but as i said that caused path auto functionality to fail

xendk:
Tried setting #value to the same value as #default_value for the elements?

pjohans:
What do you mean? Do you suggest to '#disable' AND set '#value' = '#default_value' for the checkbox?

xendk:
Yep. I'd wager that #disable disables the usual validate-input-and-put-in-#value code, which causes the elements to have no value. Setting it explicitly might work. Check out https://api.drupal.org/api/drupal/includes%21form.inc/function/_form_builder_handle_input_element/7

pjohans:
i'll take on that bet (odds 5 to 1), and give it a try tomorrow

holt83:
Did you try setting #access = FALSE?

And is there a special reason for using an after_build form callback instead of just making your alterations directly in the from alter?

xendk:
`$form['path']['#access'] = FALSE;` works, but completely hides the settings, seemed that wasn't the intention.

However, I'm all for, node edit pages are crowded as it is, removing irrelevant stuff is a UX improvement.

holt83:
Ah I see now that is the current solution.

So they want to disable the setting, but it should still be shown? I don't get it.

xendk:
To show the user that there *is* an alias, but not allow them to change it?

modules/ding_library/ding_library.module
```diff
@@ -202,10 +202,15 @@ function ding_library_form_ding_library_node_form_alter(&$form, &$form_state) {
 }
 
 /**
- * After build method. Take out path auto option for ding libraries
+ * After build method. Add javascript to simulate disabled checkbox
```
kasperg: This is a really hacky solution - and not that secure either. Based on the comments I agree about the method.

I would appreciate a comment in the code explaining why this approach is necessary.

kasperg:
@pjohans: Review complete. Two things:

- I have a request for additional inline comments
- The behavior raises [quite a few style warnings from the JavaScript](https://scrutinizer-ci.com/g/ding2/ding2/inspections/0900bdde-d562-43aa-affb-6afaf18e0eac/issues/files/modules/ding_library/ding_library.js?status=new&orderField=path&order=asc&honorSelectedPaths=0). If you think these should be ignored then please explain why.

pjohans:
this one has an error message from CircleCI : drush: command not found .. ??

xendk:
Try merging master and pushing. There's been some changes in the Circle setup recently.

PR #154, opened 2016-02-24
Issue 1627 by Pjohans: Remove second title from event view on frontpage

pjohans:


kasperg:

PR #155, opened 2016-02-24
Issue 431 by Pjohans: Correct default value on currency

pjohans:


modules/ding_event/ding_event.admin.inc
```diff
@@ -15,14 +15,13 @@
...
   $currency = variable_get('ding_event_currency_type', 'Kr');
-
   $form['#submit'][] = 'ding_event_admin_submit';
 
   $form['ding_event']['ding_event_currency_type'] = array(
     '#type' => 'radios',
     '#title' => t('Currency'),
     '#options' => drupal_map_assoc(array('kr', '€')),
-    '#default_value' => $currency,
+    '#default_value' => strtolower($currency),
```
kasperg: This seems odd.

1. If we want the lowercase `kr` format then we should at least change the format of the default value as well (L17).
2. We would still need the `strtolower` for backwards compatability. We should be able to fix this using a `hook_update` to convert the variable instead of fixing it runtime. Having an unexpected format in a database could lead to problems later on.

pjohans:
the variable was okay. The options array had kr in lowercase - fixed

kasperg:
@pjohans: Great!

Reviewed and approved.

PR #156, opened 2016-02-25
Issue 1487: Fix BPI preview

rimi-itk:
http://platform.dandigbib.org/issues/1487


kasperg:

The Scrutinizer failure is a false negative. 7 fixed and 7 introduced errors.

PR #163, opened 2016-02-26
Issue #1629 by Inlead: Decode item title when reserving.

guddo:
http://platform.dandigbib.org/issues/1629

### DIFF
--- a/modules/ding_reservation/ding_reservation.module
+++ b/modules/ding_reservation/ding_reservation.module
@@ -388,7 +388,8 @@ function ding_reservation_reserve_form_submit($form, &$form_state) {
       // @todo use profile
       if (!empty($reservation_result['branch']) && ding_provider_implements('reservation', 'branch_name')) {
         $branch_name = ding_provider_invoke('reservation', 'branch_name', $reservation_result['branch']);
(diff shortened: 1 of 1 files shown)


kasperg:
Reviewed. I agree that this fixes the encoding issue in this specific but I am considering whether this is the right approach to a fix.

The problem indicates that we have reservations with encoded titles. I cannot spot anywhere else in the codebase where we decode reservation titles. In that case I would assume that we have the same encoding problem in the reservation list etc. If that is the case then I think we should aim for a fix closer to the root of the problem.

@guddo @AnatolyMuntean: How do you see this?

guddo:
@kasperg: The difference between confirmation popup and reservation list could be usage of ding_entity vs ting_entity. Would you suggest to decode it in [getTitle() of ding_periodical](https://github.com/ding2/ding2/blob/master/modules/ding_periodical/ding_periodical_reservable.class.inc#L64) already?

kasperg:
@guddo: Thanks for the quick followup.

I considered something like decoding it in `getTitle()` but I have not looked into the consequences and wanted to get your opinion of it.

Regarding usage of the different classes I cannot spot any decoding in either class. That leaves me wondering.

AnatolyMuntean:
Decoding in `getTitle()` doesn't help. There was no deep dig into where it comes from encoded.

kasperg:
OK. I guess we have to leave it at that. Approved and merged.

PR #165, opened 2016-03-03
469 : Add option for admin to configure holdings display

JrgenGNielsen:
original commit by Morten Bartvig


JrgenGNielsen:
Code moved to ding_ting_frontend, with a tab under "admin/config/ting/settings". 
Also, uses hook_ding_entity_view() as suggested by @holt83

JrgenGNielsen:
Moved to https://github.com/ding2/ding2/pull/166

PR #166, opened 2016-03-03
469 : Add option for admin to configure holdings display

JrgenGNielsen:
original commit by Morten Bartvig
469: Fix errors found by Scrutinizer
469: Fix errors found by Scrutinizer


kasperg:

Scrutinizer fails because of old issues and is thus allowed to fail.
Cricle CI fails because of problems with Drush setup unrelated to this change and is thus allowed to fail.

PR #167, opened 2016-03-03
Issue #1614: place2book

cableman:
This PR fixes the 500 error from place2book when ticket type is not set.

It also cleans up some of the code and remove an undefined error in the edit node form.

### DIFF
--- a/modules/ding_place2book/ding_place2book.install
+++ b/modules/ding_place2book/ding_place2book.install
@@ -1,12 +1,11 @@
 <?php
-
 /**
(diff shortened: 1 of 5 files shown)


cableman:
Scrutinizer throws a lot of documentation error, which comes from my experiment to clean up the code a bit (100+ issues fixed).

I don't think we should use more time on this as there should be a whole new module in the pipe-line as place2book has a whole new API.

cableman:
Think this have been out-date.

PR #169, opened 2016-03-03
Issue #1638 by Inlead: Add auto focus on login and search form.

guddo:
http://platform.dandigbib.org/issues/1636


guddo:
This needs more work.

PR #170, opened 2016-03-04
Issue #556 by Inlead: Add pagepreview module as ding2 dependency.

guddo:
http://platform.dandigbib.org/issues/556


kasperg:

Circle CI fails because of problems with Drush setup unrelated to this change and is thus allowed to fail.

PR #174, opened 2016-03-07
Issue #659: Added gatewayf to varnish VCL

cableman:
Pipe gatewayf request through varnish to ensure they are not cached.

Also set verified to TRUE in the alma self reg. call.

See http://platform.dandigbib.org/issues/659#note-31 about these corrections.


kasperg:

Circle CI fails because of problems with Drush setup unrelated to this change and is thus allowed to fail.

PR #175, opened 2016-03-07
Issue #466 by Inlead: Redirect directly to object if only 1 item of that type.

inlead:
http://platform.dandigbib.org/issues/466


JrgenGNielsen:
reviewed and accepted

PR #176, opened 2016-03-08
Issue #1414 and #1193 (mobile menu and provider user fixes)

cableman:
Fixed issues with the provider users profile (limit view access and disclosure of provider users names) also fixes the mobile menu by using (/user/me/{view/edit/status}).


holt83:
Awesome PR :)

Did you consider:
$items['user']['delivery callback'] = 'ding_user_deliver_html_page';

Instead of hook_page_delivery_callback_alter() implementation?

cableman:
Hi @holt83 

I missed the 'delivery callback' option in the menu hook. I might look into that one :-)

modules/ding_user_frontend/ding_user_frontend.module
```diff
@@ -330,3 +329,54 @@ function ding_user_frontend_build_menu($path = NULL) {
...
+  if (preg_match('/^user\/(\d+)\/(.*)$/', $path, $matches)) {
+    // As default change the path to use "me".
+    $path = 'user/me/' . $matches[2];
+
+    // To allow site administrators to edit other user they have to be filter
+    // out of this alter. When accessing another drupal user account.
+    if ($user->uid != $matches[1] && (array_search("administrators", $user->roles) || array_search("local administrator", $user->roles))) {
+      // This is an "administrators" or "local administrator" trying to edit
+      // another user. To "allow" this we reset the path.
+      $path = $original_path;
```
kasperg: The structure here seems odd:

1. L345 changes the path and L352 resets it. I would prefer if we check if all conditions are met and set it once. I appreciate the level of commenting explaining the situation though.
2. I think `array_search()` should be replaced with `in_array()`. We do not care about positioning and if one of the roles we are looking for have position 0 then that will evaluate to `false`.
3. I am considering whether we should do this check based on permissions instead of roles. Would that be possible?

modules/ding_user/ding_user.module
```diff
@@ -33,6 +33,28 @@ function ding_user_preprocess_html(&$variables) {
...
+ *        provider users on an site and the update will take a long time to
+ *        preform. But if any one tries to access exiting provider users page
+ *        via an alias they will get a "page not found".
+ */
+function ding_user_pathauto_alias_alter(&$alias, $context) {
+  if ($context['module'] == 'user') {
+    $user = reset($context['data']);
+    // The function ding_user_is_provider_user can not be used here. Because if
+    // the user is logging in for the first time the "authmap" table have not
+    // been populated when this alias function is called.
```
kasperg: :+1:

modules/ding_user/ding_user.module
```diff
@@ -91,6 +113,12 @@ function ding_user_menu_alter(&$items) {
   $items['user/%user/view']['title'] = t('User profile');
   $items['user/%user/edit']['title'] = t('Edit user profile');
   $items['user/%pm_arg/status']['title'] = t('User status');
+
+  // Limit access to provider users page.
+  $items['user/%user/view']['access callback'] = 'ding_user_provider_user_access';
+  $items['user/%user/view']['access arguments'] = array(1);
+  $items['user/%user']['access callback'] = 'ding_user_provider_user_access';
+  $items['user/%user']['access arguments'] = array(1);
```
kasperg: :+1:

modules/ding_user/ding_user.module
```diff
@@ -33,6 +33,28 @@ function ding_user_preprocess_html(&$variables) {
...
 
 /**
+ * Implements hook_pathauto_alias_alter().
+ *
+ * Ensures that provider user don't have aliases create for their accounts.
+ *
+ * @Note: There is no update function for this as there are potential 100.000 +
+ *        provider users on an site and the update will take a long time to
+ *        preform. But if any one tries to access exiting provider users page
+ *        via an alias they will get a "page not found".
```
kasperg: I do not think this approach is the right one. It results in obsolete code (`ding_user_page_delivery_callback_alter` and `ding_user_deliver_html_page`) being run on many requests - even after any unwanted aliases have been removed.

I agree that there might be many provider users but [people have provided SQL queries which perform the cleanup](http://platform.dandigbib.org/issues/1414#note-14). How long do they take to execute in practice?

If they in fact take to long then I would prefer an approach where we create queue jobs to perform the deletion at cron time. After finished the related code will no longer be executed.

kasperg:
@cableman: Review complete.

I appreciate the level of commenting here but I have a few comments.

modules/ding_user_frontend/ding_user_frontend.module
holt83: Is all this necessary at all?

I just don't see why you would generate links on the form /user/me/* for other users than the current user. IMO the only check needed is to verify that the link is for the current user ($user->uid == $matches[0]).

cableman:
I have updated the code base on the review and fixed the issue mentioned.

JrgenGNielsen:
reviewed and accepted

PR #177, opened 2016-03-09
Issue #839 by Inlead: Role validation on user create.

NicolaeGrigoriu:
http://platform.dandigbib.org/issues/839

### DIFF
--- a/modules/ding_user/ding_user.module
+++ b/modules/ding_user/ding_user.module
@@ -192,6 +192,13 @@ function ding_user_form_alter(&$form, $form_state, $form_id) {
         );
       }
       break;
+
+    case 'user_register_form':
+      global $user;
+      if (user_access('administer users') && !in_array('administrators', $user->roles)) {
+        $form['#validate'][] = '_ding_user_roles_form_validate';
+      }
+      break;
   }
 }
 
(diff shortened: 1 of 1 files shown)


modules/ding_user/ding_user.module
```diff
@@ -937,3 +944,17 @@ function ding_user_form_contact_site_form_alter(&$form, $form_state, $form_id) {
...
+/**
+ * Check if newly created user has assigned new roles.
+ *
+ * @param array $form
+ *   Register form elements.
+ * @param array $form_state
+ *   State of the submitted form.
+ */
+function _ding_user_roles_form_validate(array $form, array &$form_state) {
+  if (!array_filter($form_state['values']['roles_change'])) {
```
JrgenGNielsen: check should be on roles, not roles_change - see if at least one more value is selected, apart from " authenticated user"

NicolaeGrigoriu: On user registration the roles array is not populated, it is on user edit. For user registration roles_change array is populated. So, only on this array can be verified if a role has been assigned.

JrgenGNielsen: quiet right - my bad

JrgenGNielsen:
reviewed and accepted

cableman:
Merged by hand to resolve conflict.

PR #179, opened 2016-03-09
Issue #1100 by Inlead: Implement administration_menu.

inlead:
http://platform.dandigbib.org/issues/1100

### DIFF
--- a/ding2.profile
+++ b/ding2.profile
@@ -835,6 +835,7 @@ function ding2_admin_menu_shortcuts() {
     $content = variable_get('admin_menu_components', array());
 
     if (empty($content)) {
+      module_load_include('inc', 'admin_menu', 'admin_menu');
       admin_menu_theme_settings();
     }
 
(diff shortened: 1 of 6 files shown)


JrgenGNielsen:
44 issues in Scrutinizer, but none related to this PR.
reviewed and accepted

ding2.profile
```diff
@@ -829,3 +829,24 @@ function ding2_set_cookie_page() {
...
+/**
+ * Enabling Shortcuts plugin for Administration Menu module.
+ */
+function ding2_admin_menu_shortcuts() {
+  if (module_exists('admin_menu')) {
+    $content = variable_get('admin_menu_components', array());
+
+    if (empty($content)) {
+      module_load_include('inc', 'admin_menu', 'admin_menu');
+      admin_menu_theme_settings();
```
kasperg: Why is calling `admin_menu_theme_settings()` necessary here?

The method is used to generate a form, the return value is not used and [the function does not seem to have any sideeffects](http://api.drupalhelp.net/api/admin_menu/admin_menu.inc/function/admin_menu_theme_settings/7.3).

If there is a reason please add a comment explaining why.

guddo: It's truly hard to recall what you've been doing 6 months ago.
We don't assign this function to any variable and its not used. I'll remove it.

PR #183, opened 2016-03-10
Fix Drush installation on Circle CI

kasperg:
We only need to delete the vendor folder to remove preexisting
dependencies. Deleting the entire folder causes Composer to install in
~/.config/composer and ~/.config/composer/vendor/bin is outside the
project path.

This causes subsequent calls to drush to fail.


cableman:
Reviewed and approved

guddo:
@JrgenGNielsen Why do you want this to be committed within other PRs?

PR #185, opened 2016-03-10
Issue #594 by Inlead: Display language on non-Danish fiction items.

inlead:
http://platform.dandigbib.org/issues/594


guddo:
**Requirement change**: Do not check whether this is fiction or not, instead check whether the item is of type "bog" (keep in mind that it may be collections with several items).

cableman:
Looks good, just need to be rebased.

guddo:
Done

PR #191, opened 2016-03-16
1660: Update translations on update.

kasperg:
Without the hook_update the translations are not imported.


kasperg:
I have to merge this myself.

The scrutinizer failure is not related to the change at hand so that can be ignored.

PR #193, opened 2016-03-17
#659: Fixes for Gatewayf

kasperg:
This pull request fixes a range of issues with the Gatewayf Implementation:

- The verified flag should be set to "true" instead of TRUE which is converted to 1.
- Retrieval of texts should conform to the value format of `text_format` field types
- Prevent watchdog from crashing when hash is not set
- Fix juggling of WAYF attributes during registration
- Prevent CPR, PIN and email address from being logged during self registration

### DIFF
--- a/modules/ding_gatewayf/ding_gatewayf.module
+++ b/modules/ding_gatewayf/ding_gatewayf.module
@@ -169,14 +169,15 @@ function ding_gatewayf_login() {
 function ding_gatewayf_logout() {
   $config = _ding_gatewayf_load_configuration();
 
-  // Ensure gatewayf attributes store in the session is removed.
-  unset($_SESSION[DING_GATEWAYF_ATTRIBUTES]);
+  // Remove indication of WAYF login.
+  unset($_SESSION[DING_GATEWAYF_IS_LOGGED_IN]);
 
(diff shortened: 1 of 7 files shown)


modules/ding_gatewayf/ding_gatewayf.module
```diff
@@ -169,14 +169,15 @@ function ding_gatewayf_login() {
...
 
-  // Get the sites base url, which is used to redirect the user after logout.
-  global $base_url;
+  // Get the logout url for the site, which is used to redirect the user after
+  // logout. This ensures that the logout process is completed.
+  $logout_url = url('user/logout', array('absolute' => TRUE));
 
   // Redirect to the gateway.
-  header('Location:' . $config['service']['endpoint'] . '?returnUrl=' . $base_url . '&op=logout');
+  header('Location:' . $config['service']['endpoint'] . '?returnUrl=' . $logout_url);
```
cableman: You have missed the "&op=logout" that tells gatewayf that you what to logout?

cableman:
Reviewed and it looks good. Just had one comment about the logout URL missing operation (op) parameter.

cableman:
Reviewed and awaiting merge

kasperg:
I will go ahead and merge this then.

PR #195, opened 2016-03-22
1671: Pluggable cover providers

xendk:
Reworks ting_covers to rely on other modules to find the covers. Adds
ting_covers_addi and an update hook to migrate settings.

Adds menu callback to ensure source file of covers are available for
image.module to scale.

PR #196, opened 2016-03-23
1672: Carousel makeover

xendk:
Major rewrite of carousel, including: multiple carousels, pre-loaded
tabs, progressive loading, pre-loading of inactive tabs, tab switching
transitions, progressive fetching of covers from backend.

Requires https://github.com/ding2/ding2/pull/195


modules/ting_search_carousel/ting_search_carousel.pages.inc
```diff
@@ -7,33 +7,25 @@
...
- * @param array $index
- *   Array index storing specific data.
+ * @param int $index
+ *   Configuration index.
+ * @param int $offset
+ *   Offset into result to fetch from.
  */
-function ting_search_carousel_result($index) {
+function ting_search_carousel_result($index, $offset) {
   $content = FALSE;
```
JrgenGNielsen: leftover code? $content is defined as an empty string two lines down

JrgenGNielsen:
apart from a small detail: top nice! reviewed and accepted

modules/ting_search_carousel/ting_search_carousel.pages.inc
xendk: Quite. I'll remove it.

xendk:
@JrgenGNielsen I've added 3 small cleanup commits. No functionality change, just changing the field description to not refer to ADDI, logging as ting_cover rather than ting_cover_addi, and obsoleted file references in the info file.

xendk:
@JrgenGNielsen I've added two more cleanup commits. One prints out the description for the first tab, and the second removes TRUE/FALSE as a class.

cableman:
Merged by hand to resolve conflict in compiled CSS.

PR #200, opened 2016-03-29
408 by Inlead: Attempt to save users and fix strlen problem.

inlead:
http://platform.dandigbib.org/issues/448

### DIFF
--- a/modules/ding_user/ding_user.module
+++ b/modules/ding_user/ding_user.module
@@ -478,30 +478,39 @@ function _ding_user_update_authname($authname) {
   // which is still different from what it should be.
   $query = db_select('authmap', 'a')
     ->fields('a', array('uid'))
-    ->condition('authname', $authname_chunk . '%', 'LIKE')
-    ->condition('authname', $authname, '<>');
-  $uid = $query
+    ->condition('authname', $authname_chunk . '%', 'LIKE');
+  $uids = $query
     ->execute()
-    ->fetchField();
+    ->fetchCol();
 
   // In the case there is still a user with requested authname,
   // update his authname tokens in database.
-  if ($uid) {
+  if ($uids) {
+    $latest_uid = max($uids);
+    $original_uid = min($uids);
+
+    // In case of same provider user created erroneously, delete the one
+    // with latest uid assigned. But being aware that there are actually
+    // more users with same authname chunk, yet with different uid's.
+    if ($latest_uid != $original_uid) {
+      user_delete($latest_uid);
+    }
+
     db_update('authmap')
       ->fields(array(
         'authname' => $authname,
       ))
-      ->condition('uid', $uid, '=')
+      ->condition('uid', $original_uid, '=')
+      ->condition('authname', $authname, '<>')
       ->execute();
 
     db_update('users')
       ->fields(array(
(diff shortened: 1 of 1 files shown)


modules/ding_user/ding_user.module
```diff
@@ -451,58 +451,66 @@ function ding_user_default_authname($name) {
...
 
-  // Try to find the users uid based on the old hash value (length of the old
-  // hash should be 96).
-  $uid = db_select('authmap', 'a')
+  // Pick the user id that has almost the authname value that we seek,
+  // which is still different from what it should be.
+  $query = db_select('authmap', 'a')
     ->fields('a', array('uid'))
-    ->condition('authname', $auth_name_old)
+    ->condition('authname', $authname_chunk . '%', 'LIKE');
```
xendk: This makes me edgy. I would prefer to add another condition that only selects authnames where strlen(authname) < strlen(proper_authname), so it'll only consider the "short" authnames. Should be just another ->condition().

Having a collision in the first 32 chars (which is 24 bytes of the hash) is unlikely, but sorting out good authnames feels better.

JrgenGNielsen:
@inlead : any reply to Thomas' note?

modules/ding_user/ding_user.module
```diff
@@ -451,58 +451,66 @@ function ding_user_default_authname($name) {
...
+  // update his authname tokens in database.
+  if ($uids) {
+    $latest_uid = max($uids);
+    $original_uid = min($uids);
+
+    // In case of same provider user created erroneously, delete the one
+    // with latest uid assigned. But being aware that there are actually
+    // more users with same authname chunk, yet with different uid's.
+    if ($latest_uid != $original_uid) {
+      user_delete($latest_uid);
```
kasperg: This seems to assume that there will be 2 UIDs matching the chunck at most since we only delete the latest. How can be we sure this is the case?

Also: How can we be sure that the latest uid is the right one to delete?

modules/ding_user/ding_user.module
kasperg: I agree. This is a complex issue and avoiding further problems is good.

kasperg: Based on [the discussion in Redmine](http://platform.dandigbib.org/issues/448#note-51) a correct authname should always be longer than an incorrect one.

Consequently we should look for existing authnames that form the beginning of the proper authname and update these instead of comparing the first 32 chars.

modules/ding_user/ding_user.module
```diff
@@ -451,58 +451,66 @@ function ding_user_default_authname($name) {
   // characters, leaving enough room for our suffix.
```
kasperg: Please update the comment on the line above. The correct number of chars is 88.

modules/ding_user/ding_user.module
xendk: If the select above filters out proper authnames, this should delete all found partial authnames.

xendk: In the case a user has two accounts, the one with the higher UID should be the last created, and most likely last used.

Order by last login and take the last used?

inlead:
Closed in favour of #368.

PR #202, opened 2016-04-04
Issue #1661 by holt83: Fix permission on admin content view.

holt83:
http://platform.dandigbib.org/issues/1661


JrgenGNielsen:
reviewet og godkendt

PR #115, opened 2016-01-20
Issue1313

pjohans:

PR #203, opened 2016-04-06
1662: Reapply pseudo delta support for relations

JrgenGNielsen:
Redo code from gielfeldt on new codebase.

### DIFF
--- a/modules/ting_reference/ting_reference.module
+++ b/modules/ting_reference/ting_reference.module
@@ -121,10 +121,36 @@ function ting_reference_field_insert($entity_type, $entity, $field, $instance, $
   field_cache_clear('field:' . $entity_type . ':' . $entity_id);
 }
 
+
 /**
  * Implements hook_field_update().
  */
 function ting_reference_field_update($entity_type, $entity, $field, $instance, $langcode, &$items) {
+
+  // When nodes are saved problematically and edit form is not used,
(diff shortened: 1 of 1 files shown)


kasperg:
I have tried to look into why this problem reoccured and have found this following:

1. The original change *was* applied in fec0f01. Not from gielfeldt but from inlead
2. PR #115 reverted this change - probably because [the related issue](http://platform.dandigbib.org/issues/1313) contained a change that no longer applied.

To get this right once and for all (hopefully) can you confirm that with the suggested change [issue #1313](http://platform.dandigbib.org/issues/1313) also stays fixed?

modules/ting_reference/ting_reference.module
```diff
@@ -121,41 +121,81 @@ function ting_reference_field_insert($entity_type, $entity, $field, $instance, $
   field_cache_clear('field:' . $entity_type . ':' . $entity_id);
 }
 
+
 /**
  * Implements hook_field_update().
  */
 function ting_reference_field_update($entity_type, $entity, $field, $instance, $langcode, &$items) {
+
+  // When nodes are saved problematically and edit form is not used,
```
kasperg: problematically => programatically?

kasperg:

@JrgenGNielsen: There is a small typo which you are welcome to fix.

PR #208, opened 2016-04-07
Issue1655

pjohans:


pjohans:
I don't know why CircleCI fails on install - just did a clean install from the featrue branch with no problems. What to do?

xendk:
Tried rebasing the feature branch to the current master?

modules/ting_oembed/ting_oembed.info
```diff
@@ -7,3 +7,4 @@ version = 7.x-1.x
 ;files[] =
 dependencies[] = oembed
 dependencies[] = ting_oembed_features
+dependencies[] = virtual_field
```
kasperg: You are absolutely right to add `virtual_field` as a dependency here, but then we can remove it from `ting_oembed_features.info` as well.

kasperg:
There seems to be a circular dependency between [`ting_oembed`](https://github.com/pjohans/ding2/blob/Issue1655/modules/ting_oembed/ting_oembed.info#L9) and [`ting_oembed_features`](https://github.com/pjohans/ding2/blob/Issue1655/modules/ting_oembed/ting_oembed_features/ting_oembed_features.info#L10).

That is not a good thing and perhaps that could cause the problems. I think the right way to go about this would be to have `ting_oembed_features` (the module using the new field) depend on `ting_oembed` (the module declaring the new field) and have other modules depend on `ting_oembed_features`.

pjohans:
okay i'll try that

kasperg:
Reviewed and approved. Ready for merge.

PR #211, opened 2016-04-08
Fbs wayf registration

xendk:
Create user support for FBS.

Also minor cleanups (an exception class that was not used, for instance).


modules/ding_provider/ding_provider.exceptions.inc
```diff
@@ -80,13 +80,3 @@ class DingProviderUserException extends Exception {
     }
   }
 }
-
```
cableman: It this exception no longer used by the providers ?

modules/fbs/prancer/SwaggerApiError.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace Reload\Prancer;
```
cableman: This is for sure outside this PR's scope, but should the namespace have been DDB and not Reload?

cableman:
Reviewed with minor comment.

modules/ding_provider/ding_provider.exceptions.inc
xendk: Never have been. It provides no functionality, it has a private property nobody uses, and the only one using it was ding_user catching it.

But ding_user should catch any exception, and as nobody is throwing it anyway, I removed it.

modules/fbs/prancer/SwaggerApiError.php
xendk: Do DDB want to maintain a generic Swagger spec to PHP code generator?

xendk:
Rebased.

modules/ding_provider/ding_provider.exceptions.inc
cableman: Okay, super :-)

modules/fbs/prancer/SwaggerApiError.php
cableman: I have not used Swagger, so I don't know... so I think that would be a: No

cableman:
Reviewed and awaiting next release

PR #214, opened 2016-04-11
1197 by Inlead: Ported toggle_format changes from easyOPAC.

inlead:
http://platform.dandigbib.org/issues/1197


JrgenGNielsen:
There's a couple of Scrutinizer errors - perhaps you'll take a look at them?

guddo:
The rest of the Scrutinizer failures can be ignored.

modules/ding_toggle_format/js/jquery.cookie.min.js
```diff
@@ -1,10 +1,137 @@
...
+      }));
+    };
+
+    api.withConverter = init;
+
+    return api;
+  }
+
+  return init();
+}));
```
holt83: We seem to be using a much older version of this in ding_facetbrowser module. IMO this should be added via Libraries API since it's external and so it can easily be shared among modules.

JrgenGNielsen:
reviewed and accepted

inlead:
Closed as requested by the business.

PR #216, opened 2016-04-12
P2 pr plus tests

xendk:
This is the Behat tests with the P2 PR merged in.

Currently fails on the P2 branch not having committed updated CSS.

This will collide with the upcoming phpunit/Selenium test PR.

Expects a parallelism of two on circle. One will run tests in PhantomJS, the other Selenium.

PR #217, opened 2016-04-12
1751: Phpunit tests from ding2_selenium

xendk:
This integrates and runs the tests from ding2_selenium on CircleCI.

Collides with #216, these needs to be merged.

Issue: http://platform.dandigbib.org/issues/1751

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -1,0 +1,24 @@
+
+test-setup: setup-modules setup-config
+
+test-setup-phpunit: setup-modules setup-config-phpunit
+
+# Install a test site.
+install-site:
+	drush site-install ding2
+
+# Setup test modules on site.
+setup-modules:
+	# Nothing to be done yet.
+
+setup-config:
+	# Configure ting.module
+	drush vset --exact ting_agency 100200
+	drush vset --exact ting_search_url 'http://oss-services.dbc.dk/opensearch/4.0.1/'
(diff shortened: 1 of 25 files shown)


tests/Makefile
```diff
@@ -0,0 +1,35 @@
...
+	# Configure ting.module
+	drush vset --exact ting_agency 100200
+	drush vset --exact ting_search_url 'http://oss-services.dbc.dk/opensearch/4.0.1/'
+	drush vset --exact ting_search_profile test
+
+setup-config-phpunit: setup-config
+	# ting.module.
+	drush vset --exact ting_agency '775100'
+	drush vset --exact ting_search_profile 'ding2taltest'
+	drush vset --exact ting_search_url 'http://opensearch.addi.dk/4.0.1/'
```
guddo: I think you should try http://oss-services.dbc.dk/opensearch/4.0.1/ - it's open for the public.

xendk: I followed the instructions on https://github.com/ding2/ding2_selenium I assumed it was important.

guddo: It is, however that might be the reason that there is no items available - since this URL is restricted by IPs.
Similar to the issue you have with moreinfo.

DDBdrift:
https://platform.dandigbib.org/issues/1751 | Closed

PR #220, opened 2016-04-14
Issue 1631: Always return an entity from ding_user_provider_profile

pjohans:
This issue is caused by ask_vopros.module::ask_vopros_preprocess_html() calling field_get_items() with an array of profile2 instead of an entity (because the dummy_provider does not implement a profile).  
I have chosen to fix it in ding_user, since the profile2_load_by_user() returns an array if given  profile_type does not match and another profile is found - in this case ding_staff, and it is the responsibility of ding_user to return the profile.

### DIFF
--- a/modules/ding_user/ding_user.module
+++ b/modules/ding_user/ding_user.module
@@ -368,7 +368,8 @@ function ding_user_profile_update_submit(&$form, &$form_state) {
  */
 function ding_user_provider_profile($account) {
   $profile_type = ding_user_get_provider_profile_type();
-  return profile2_load_by_user($account, $profile_type->type);
+  $profile2 = profile2_load_by_user($account, $profile_type->type);
+  return is_array($profile2) ? reset($profile2) : $profile2;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


xendk:
Shouldn't it return NULL if no provider profile is found, rather than a random profile? Anyone calling ding_user_provider_profile() would expect to get a provider profile, not a staff/whatever profile.

kasperg:
I agree with @xendk here.

`ding_user_provider_profile()` should return `NULL` if there is no provider profile for the user.

pjohans:
If it returns NULL the issue will not be fixed. Ask vopros will pass NULL to field_get_items which in turn will pass NULL to entity_extract_ids and that will cause the exception to be thrown. You could argue that ask_vopros should do some datacheck.

xendk:
If ask_vopros is the problem, why didn't you make a PR against that (fixing ding_user_provider_profile() to return null if no provider profile is found is still a good idea)?

Not every project moves at the glacial tempo of Ding core.

pjohans:
okay i'll do a patch on ask_vopros and return a NULL as suggested - glaciers move faster and faster with the environment heating up

pjohans:
No patch needed for ask_vopros - it does the datacheck ok

kasperg:
@pjohans: Thats what I thought. Thanks.

kasperg:
Merged.

PR #227, opened 2016-04-27
16 by Inlead: Add link to bookmarks in bookmark popup.

inlead:
http://platform.dandigbib.org/issues/16


kasperg:
The new list service in P2 makes this PR obsolete.

PR #231, opened 2016-04-29
Issue #1346: Selectable TLS/SSL version in alma

cableman:
Allow the site administrator to select the security protocol to use when communicating with Alma.

See http://platform.dandigbib.org/issues/1346


JrgenGNielsen:
reviewed and accepted

PR #232, opened 2016-04-30
Issue #1612 by holt83: Handle all covers on the client.

holt83:
http://platform.dandigbib.org/issues/1641


holt83:
The changes in this PR isn't relevant anymore after several changes to ting_covers module and is therefore closed.

PR #233, opened 2016-05-04
1723 by Inlead: Do not use deprecated version of addReservations.

inlead:
http://platform.dandigbib.org/issues/1723


xendk:
Shouldn't there be a bit more code to deal with the fact that the return types are different?

guddo:
Please see https://github.com/ding2/ding2/pull/239.

PR #235, opened 2016-05-10
Ding wayf dependency issues 

cableman:
Fixed "500" error when ding_gatewayf_registration is not enabled.

See http://platform.dandigbib.org/issues/1725


kasperg:
Reviewed and approved. Thanks

PR #241, opened 2016-05-27
1195: Disable overlay module

JrgenGNielsen:
http://platform.dandigbib.org/issues/1195


cableman:
Reviewed and awaiting merge

PR #243, opened 2016-05-30
1695: No sort used message, if sort is selected

JrgenGNielsen:
http://platform.dandigbib.org/issues/1695


cableman:
Reviewed and awaiting merge

PR #244, opened 2016-05-31
Issue773

pjohans:
Here is a new pull-request on this one. please dismiss the old one ( https://github.com/ding2/ding2/pull/153 )


cableman:
Somethings from the orginal PR (https://github.com/ding2/ding2/pull/153) that need to be looked at:

@pjohans: Review complete. Two things:

 * I have a request for additional inline comments
 * The behavior raises quite a few style warnings from the JavaScript. If you think these should be ignored then please explain why.

cableman:
Please correct the Scrutinizer errors.

PR #249, opened 2016-06-08
1749: Make provider file paths relative to provider dir

kasperg:
The current implementation is inconsistent:

1. The “file” option under “provides” are relative to the Drupal root dir
2. The root “file” option is relative to the ding_provider root dir

Part 2 results in developers with settings functions outside the .module
file have to jump through hoop to load the correct file.

Making all files paths relative to the Drupal root would require the 
smallest number of changes but that is not possible due to the way
the hook_menu() “file path” option works.

Instead we make all file paths relative to the provider root directory.
This is also the simplest going forward as all providers follow this
pattern anyway.

This change:
- Updates “file path” op ion  in the the ding_provider hook_menu
implementation for loading settings forms to be relative to the 
provider module root dir.
- Updates provider function invocation to load files relatative to the
provider module root dir
- Update all existing providers (alma, connie, fbs, openruth, wyrna) to
set file paths relative to the individual provider module root 
directories.

### DIFF
--- a/modules/alma/alma.module
+++ b/modules/alma/alma.module
@@ -54,35 +54,33 @@ function alma_get_interest_periods() {
  * Implements hook_ding_provider().
  */
 function alma_ding_provider() {
-  $path = drupal_get_path('module', 'alma');
-
   return array(
     'title' => 'Alma provider',
     'settings' => 'alma_settings_form',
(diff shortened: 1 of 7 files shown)


cableman:
Reviewed and awaiting merge

PR #250, opened 2016-06-08
1750 by Inlead: Added missing clear cache that is called after DIBS payment is accepted.

inlead:
http://platform.dandigbib.org/issues/1750

### DIFF
--- a/modules/fbs/includes/fbs.user.inc
+++ b/modules/fbs/includes/fbs.user.inc
@@ -150,3 +150,15 @@ function fbs_user_create($cpr, $pin_code, $name, $mail, $branch_id) {
     throw $exception;
   }
 }
+
+/** 
+ * Implements clear_cache().
+ *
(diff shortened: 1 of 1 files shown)


modules/fbs/includes/fbs.user.inc
```diff
@@ -110,3 +110,15 @@ function fbs_user_update_pincode($account, $new_pincode) {
   $result['creds']['library_card_number'] = $patron_info['library_card_number'];
   return $result;
 }
+
+/**
+ * Implements clear_cache().
+ *
+ * This hook is only called from ding_dibs after a payment have been accepted.
+ *
+ * @param stdClass $account
```
cableman: This should be "object" as identified by scrutinizer. It must be "object" as that is a instance of a class.

cableman:
Reviewed only very minor issue with the comment.

cableman:
Reviewed and awaiting merge

PR #252, opened 2016-06-13
Issue #1732 by holt83: Use Webtrends metatags to track onsite search.

holt83:
http://platform.dandigbib.org/issues/1732


JrgenGNielsen:
reviewed and accepted

PR #253, opened 2016-06-14
Issue #1731 by holt83: Track user selected facets in Webtrends OSS

holt83:
Enables tracking of user selected facets in Webtrends onsite search statistics. Without this it is not possible to differentiate searches with or without facets in these stats.

These need special handling since they are implemented as query parameters and therefore not part of the search key in the URL.

Note that this PR is based on additions from https://github.com/ding2/ding2/pull/252 and therefor include those commits.

http://platform.dandigbib.org/issues/1731

### DIFF
--- a/modules/ding_webtrends/ding_webtrends.module
+++ b/modules/ding_webtrends/ding_webtrends.module
@@ -92,7 +92,7 @@ EOT;
     $meta_tag['#tag'] = 'meta';
     $meta_tag['#attributes'] = array(
       'name' => 'WT.oss',
-      'content' => $search_info['search_key'],
+      'content' => $search_info['type'] . ':' . $search_info['search_key'],
     );
     drupal_add_html_head($meta_tag, 'search_query_meta_tag');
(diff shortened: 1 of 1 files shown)


modules/ding_webtrends/ding_webtrends.module
```diff
@@ -96,4 +87,79 @@ EOT;
...
+  }
+}
+
+/**
+ * Implements hook_panels_pane_content_alter().
+ */
+function ding_webtrends_panels_pane_content_alter($content, $pane, $args, $contexts) {
+  if ($pane->type == 'search_result' && $pane->configuration['type'] == 'ting_search') {
+    $search_result = drupal_static('ting_search_results', FALSE);
+    $search_key = $search_result->search_key;
```
cableman: You should test that $search_results is not FALSE here, before using it.

cableman:
Reviewed with just one comment.

kasperg:
This issue is obsolete. A combined approach to tracking in Webtrends will be addressed elsewhere.

PR #254, opened 2016-06-15
1755: Update ask_vopros

xendk:
Ask_vopros has moved and gotten a new version.


JrgenGNielsen:
I assume version 1.3 of Ask_vopros has been tested?
PR itself is reviewed and accepted

xendk:
Another update, there was an colorpicker missing when not using the overlay module bug.

I've squashed the branch too.

PR #265, opened 2016-07-04
Issue #278 by Inlead: Remove overlay when ESC is pressed.

inlead:
http://platform.dandigbib.org/issues/278

### DIFF
--- a/modules/ting_search/js/ting_search_overlay.js
+++ b/modules/ting_search/js/ting_search_overlay.js
@@ -60,7 +60,7 @@
     // Make sure that the overlay is removed when ESC is pressed.
     $(window).keydown(function(event) {
       // Keycode for the ESC-button.
-      if (event.which == 27) {
+      if (event.which === 27) {
         var $overlay = $('.search-overlay--wrapper');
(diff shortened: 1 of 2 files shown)


modules/ting_search/js/ting_search_overlay.js
```diff
@@ -56,6 +56,17 @@
         Drupal.TingSearchOverlay(true);
       }
     });
+
+    // Make sure that the overlay is removed when ESC is pressed.
+    $(window).keydown(function(event) {
+      // Keycode for the ESC-button.
+      if (event.which == 27) {
```
cableman: Should that have been '===' at the fix above with key-code "13"

modules/ting_search/js/ting_search_overlay.js
```diff
@@ -56,6 +56,17 @@
...
+        var $overlay = $('.search-overlay--wrapper');
+        if ($overlay !== undefined) {
+          $overlay.remove();
```
cableman: Could "Drupal.TingSearchOverlay(true);" be used here?

kasperg:

PR #266, opened 2016-07-04
1766: Removed the translation file from P2 as it contains the whole site

cableman:
See http://platform.dandigbib.org/issues/1766 for more information.


kasperg:
Reviewed, approved and merged.

PR #267, opened 2016-07-04
49 by Inlead: Possibility of making Red-text in WYSIWYG.

inlead:
http://platform.dandigbib.org/issues/49


DDBdrift:
https://platform.dandigbib.org/issues/49 | Closed

PR #269, opened 2016-07-05
Core: Patch oEmbed to prevent fatal errors

kasperg:
The following error occurs: 
“Fatal error: Unsupported operand types in /oembed.module on line 129”

Patch oEmbed module to prevent this. Patch has already been committed
but no release has been made yet.

See https://www.drupal.org/node/2021015.


guddo:
Shouldn't this be merged by now?

xendk:
I'll second the motion.

PR #270, opened 2016-07-05
1401: Fix path to openlist provider file

kasperg:
After 1749 the fixes to the path is no longer required.

This fixes a fatal error: “require_once(profiles/ding2/modules/p2/ting_openlist/profiles/ding2/modules/p2/ting_openlist/ting_openlist.handler.inc): failed to open stream: No such file or directory”

This is required because of the change in #249.


cableman:
Reviewed

PR #272, opened 2016-07-05
1671: Fix migration of addi variables

kasperg:
Addi variables are not migrated properly. This results in the following error:

PHP Fatal error:  SOAP-ERROR: Parsing WSDL: Couldn't load from '/moreinfo.wsdl' : failed to load external entity "/moreinfo.wsdl"

The error is caused by a faulty migration of the addi variables.


cableman:
Reviewed and approved

PR #273, opened 2016-07-05
1764: Added ding_p2_installer as dependency to ensure it is installed on new sites

cableman:
Se http://platform.dandigbib.org/issues/1764


kasperg:
Reviewed and approved. Merged.

PR #278, opened 2016-07-11
Issue #1797: Enable ting_covers_addi

cableman:
See http://platform.dandigbib.org/issues/1797

### DIFF
--- a/modules/ding_ting_frontend/ding_ting_frontend.info
+++ b/modules/ding_ting_frontend/ding_ting_frontend.info
@@ -20,6 +20,7 @@ dependencies[] = pm_existing_pages
 dependencies[] = strongarm
 dependencies[] = ting
 dependencies[] = ting_covers
+dependencies[] = ting_covers_addi
 dependencies[] = ting_fulltext
 dependencies[] = ting_oembed_features
 dependencies[] = ting_proxy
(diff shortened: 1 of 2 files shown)


modules/ding_ting_frontend/ding_ting_frontend.info
```diff
@@ -20,6 +20,7 @@ dependencies[] = pm_existing_pages
 dependencies[] = strongarm
 dependencies[] = ting
 dependencies[] = ting_covers
+dependencies[] = ting_covers_addi
```
kasperg: Now that we have split ting_covers into a core module with plugins I do not think it makes sense to force a dependency to one of the plugins.

Instead I suggest that we enable it in `ding_ting_frontend_install()`.

modules/ding_ting_frontend/ding_ting_frontend.install
```diff
@@ -1,22 +1,33 @@
...
-  variable_del('ding_ting_frontend_group_holdings_available');
+function ding_ting_frontend_update_7001() {
+  module_enable(array('ding_ekurser', 'ting_oembed_features'));
 }
+
+/**
+ * Enable Ting cover addi module.
+ */
+function ding_ting_frontend_update_7002(&$sandbox) {
+  module_enable(array('ting_covers_addi'));
```
kasperg: This assumes that all sites want to have it enabled. I only think this makes sense for sites which do not have a ting_covers implementation enabled already - e.g. eReolen.

I suggest that we check for hook implementations and enable based on that.

kasperg:
Closed in favour of  #401.

PR #282, opened 2016-07-13
1599 by Inlead: Add IPE Filter module.

inlead:
http://platform.dandigbib.org/issues/1599


guddo:
This PR is made for the 7.x-2.5.1 release.

translations/da.po
```diff
@@ -10641,10 +10641,6 @@ msgid "perform bookmark"
 msgstr ""
 
 #: /admin/people?render=overlay
-msgid "administer frontpage settings"
-msgstr ""
-
-#: /admin/people?render=overlay
```
kasperg: :+1: for cleanup.

modules/ding_ipe_filter/tests/tests/IPEFilterTest.php
```diff
@@ -0,0 +1,95 @@
...
+    $this->abstractedPage->waitForPage();
+    $this->open("/");
+    sleep(5);
+    $this->click("id=panels-ipe-customize-page");
+    sleep(5);
+    $this->click("xpath=(//a[contains(text(),'Add')])[2]");
+    sleep(5);
+  }
+}
+?>
```
kasperg: Missing newline and unnecessary newline.

modules/ding_ipe_filter/ding_ipe_filter.module
```diff
@@ -0,0 +1,146 @@
...
+  $assets_loaded = &drupal_static('ipe_assets_loaded', FALSE);
+  $settings = variable_get('ding_ipe_filter_table', array());
+  $region_id  = $variables['region_id'];
+  $renderer = $variables['renderer'];
+
+  $buttons = array();
+  $menu_width = 'calc(100% - 18px)';
+  $attributes = array(
+    'class' => 'ctools-use-modal ding-ipe-menu-item',
+    'style' => 'width: ' . $menu_width . '; height: auto;',
```
kasperg: Why is it necessary to set width like this?

modules/ding_ipe_filter/ding_ipe_filter.module
```diff
@@ -0,0 +1,146 @@
...
+function ding_ipe_filter_theme_registry_alter(&$registry) {
+  global $user;
+  $ipe_roles = variable_get('ding_ipe_filter_roles', array());
+
+  if (!empty($user->roles)) {
+    $role_names = array_values($user->roles);
+    $user_roles = array_combine($role_names, $role_names);
+    $roles = array_intersect(array_keys($user_roles), $ipe_roles);
+  }
+  if (!empty($roles)) {
```
kasperg: Using the permission system to handle access to the filter would simplify this code as well.

modules/ding_ipe_filter/ding_ipe_filter.install
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * @file
+ * Define functions on module install/uninstall.
+ */
+
+/**
+ * Implements hook_install().
+ */
+function ding_ipe_filter_install() {
+  $task = page_manager_get_task('page');
```
kasperg: Please add a comment explain why this install task is needed.

modules/ding_ipe_filter/ding_ipe_filter.admin.inc
```diff
@@ -0,0 +1,90 @@
...
+}
+
+/**
+ * Defines admin form.
+ */
+function ding_ipe_filter_default_settings_form() {
+  $form = array();
+  $form['filter_fieldset'] = array(
+    '#type' => 'fieldset',
+    '#title' => t('Default IPE filter settings.'),
```
kasperg: This title does not make sense compared to the form elements which deals with roles. Please consider if there are more descriptive regarding that the field actually does.

If it has something to do regarding which roles can use the filter then why is the Drupal permission system not used?

modules/ding_ipe_filter/ding_ipe_filter.module
NicolaeGrigoriu: Menu width is calculated based on the width of the links.

guddo:
PR has been rebased to current master.

modules/ding_ipe_filter/ding_ipe_filter.module
kasperg: My point is: The module includes `ipe.css` and I would expect it to be set there. If there is a specific reason to set it here instead of in the CSS file then I would like a code comment explaining why.

guddo: Fixed in the latest commit.

kasperg: :+1: Appreciate it.

modules/ding_ipe_filter/css/ipe.css
```diff
@@ -0,0 +1,40 @@
+div.panels-ipe-newblock {
```
NicolaeGrigoriu: This is needed for the IPE menu to be properly displayed

modules/ding_ipe_filter/css/ipe.css
```diff
@@ -0,0 +1,53 @@
...
+}
+.hide {
+  display: none;
+}
+.left {
+  left: 245px;
+}
+.top {
+  top: -157px !important;
+}
```
kasperg: This targets some very board selectors with which does not seem to be directly related to IPE: `.hide`, `.left`, `.top`, `.layout-wrapper .group-blocks--wrapper`.

I would suggest replacing `.hide` with `.element-hidden` which is a standard part of Drupal which does the same thing.

`.left` and `.top` has styling that seems very specific to our use case. How about renaming them to something which communicates the scope more clearly e.g. `ding-ipe-popup-left`?

kasperg:
Excellent. Appreciate the followup @guddo

PR #286, opened 2016-08-02
1823: Extract proper value for author

xendk:
Extract author properly, so it's available in theming.


JrgenGNielsen:
reviewed and approved

PR #288, opened 2016-08-04
1845: Update admin_views to 1.6

xendk:
Ref http://platform.dandigbib.org/issues/1845


cableman:
Reviewed and approved

PR #291, opened 2016-08-15
1887 by Inlead: Support the facetSpell method.

inlead:
http://platform.dandigbib.org/issues/1887


cableman:
Reviewed and approved

PR #293, opened 2016-08-24
Issue #1900 by Inlead: Fixing already declared function name.

inlead:
http://platform.dandigbib.org/issues/1900

Follow up on http://platform.dandigbib.org/issues/287 , where this P2 module declares function name that already exist.


JrgenGNielsen:
reviewed and accepted

PR #296, opened 2016-09-02
Enhancement #1516 by holt83: Remove all occurences of current user name

holt83:
http://platform.dandigbib.org/issues/1516

Now also adds denpendencies for Metatag panels to ensure it's activated on a fresh install


kasperg:
Update approved. Awaiting merge.

PR #298, opened 2016-09-07
Issue 1655: Update hook to set as active search module

pjohans:
There might be more to this one, but at least we need an update hook

See http://platform.dandigbib.org/issues/1655

### DIFF
--- a/modules/ding_ekurser/ding_ekurser.install
+++ b/modules/ding_ekurser/ding_ekurser.install
@@ -43,11 +43,3 @@ function ding_ekurser_uninstall() {
     variable_set('search_active_modules', $search_modules);
   }
 }
-
-/**
- * Implement hook_update_N.
- * Set as active searchmodule and import translations.
- */
-function ding_ekurser_update_7001() {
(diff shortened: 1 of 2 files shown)


modules/ding_ekurser/ding_ekurser.install
```diff
@@ -43,3 +43,11 @@ function ding_ekurser_uninstall() {
     variable_set('search_active_modules', $search_modules);
   }
 }
+
+/**
+ * Implement hook_update_N.
```
cableman: The "implement" statement should not be part of the comment.

Update hooks should only have a single line as comment with a short description.

modules/ding_ekurser/ding_ekurser.install
```diff
@@ -43,3 +43,11 @@ function ding_ekurser_uninstall() {
...
+ * Set as active searchmodule and import translations.
+ */
+function ding_ekurser_update_7001() {
+  ding_ekurser_install();
```
cableman: Why it this needed? It would be executed in the modules install process!

PR #305, opened 2016-09-15
P2

martinb14:


cableman:
Please make a pull request for each issue in the issue tracker, so they can be addressed and reviewed individually.

Remember to use the right commit message form when writing commit messages.

PR #310, opened 2016-09-22
Fixes #1867, #1869 and #1871 The old userprofile panel is removed 

martinb14:
http://platform.dandigbib.org/issues/1867
http://platform.dandigbib.org/issues/1871
http://platform.dandigbib.org/issues/1869


cableman:
Commit message should be on the form:

[issue number]: [short summary of change]. The subject should be kept around 50 characters long. The subject must not be more than 69 characters long. Strive for about 50 characters.

Only on issue and then the issue should be linked in the issue tracker.

cableman:
The code is correct but it does not fix the issue: http://platform.dandigbib.org/issues/1871 as this is not the solution request and this alternative has not be approved in the issue tracker.

cableman:
Also see comment in http://platform.dandigbib.org/issues/1869

JrgenGNielsen:
reviewed and accepted

PR #322, opened 2016-09-22
1923: Frontpage panels config

martinb14:
http://platform.dandigbib.org/issues/1923
Pages configuration for p2 is in seperate module. Should all these be feature exported into the coremodules?


cableman:
This PR only contains empty line fixed and a merge conflict markers ?

PR #325, opened 2016-09-22
#1885.a: Only show interaction for provider users

martinb14:
http://platform.dandigbib.org/issues/1885


modules/p2/ding_entity_rating/ding_entity_rating.module
```diff
@@ -417,26 +400,38 @@ function ding_entity_rating_remove_from_list($entity, $account) {
...
   
-  return array(
-    'ding_entity_rating' => array(
-      'title' => t('It pays to judge'),
-      'points' => 3,
-      'completed' => $complete,
-      'markup' => theme('ding_entity_rating_ding_user_complete_description', array('history_list' => $history_list, 'account' => $user)),
-    ),
-  );
+  return "<div class='description'>" .
```
JrgenGNielsen: Why drop the theme function in favor of outputting HTML directly?
If you need to return HTML, you should create a render array with a #theme property, and use drupal_render to uotput the rendered HTML.

modules/p2/ding_list/ding_list.module
```diff
@@ -788,51 +788,30 @@ The link is active for 7 days from today, after which it will be disabled for se
...
-      ,
-    );
+}
+/**
+ * Implements hook_ding_interaction_view().
+ */
+function ding_list_ding_interaction_view() {
+  $user_lists = ding_list_get_list(DING_LIST_TYPE_USER_LIST);
+  if (empty($user_lists)) {
+    return "<div class='description'>" .
```
JrgenGNielsen: this would also preferably be a rendered render array.

modules/p2/ding_list/ding_list.module
```diff
@@ -806,14 +809,21 @@ function ding_list_ding_interaction_info() {
  */
 function ding_list_ding_interaction_view() {
   $user_lists = ding_list_get_list(DING_LIST_TYPE_USER_LIST);
-  if (empty($user_lists)) {
-    return "<div class='description'>" .
-    t("Did you know you can create your own lists and add objects to them?") .
-    "</div><ul class='actions'><li>" . l(t('> Create my own list'), 'list/create') . "</li></ul>";
+  if (true || empty($user_lists)) {
```
JrgenGNielsen: the "true || " is a leftover from testing, right?

JrgenGNielsen:
reviewed and accepted

PR #327, opened 2016-09-22
#1864: Fixed followed lists notification count

martinb14:
http://platform.dandigbib.org/issues/1864


cableman:
Commit message should not start with "#" as this is the default comment char in most editors used with git. So this will create issue on some systems.

Please follow:

> [issue number]: [short summary of change]. The subject should be kept around 50 characters long. The subject must not be more than 69 characters long. Strive for about 50 characters.

cableman:
Code approved when commit message have been fixed.

JrgenGNielsen:
reviewed and accepted

PR #330, opened 2016-09-22
#1902: Special user was an experimental feature, it has been removed

martinb14:
http://platform.dandigbib.org/issues/1902


cableman:
Commit message should not start with "#" as this is the default comment char in most editors used with git. So this will create issue on some systems.

Please follow:

> [issue number]: [short summary of change]. The subject should be kept around 50 characters long. The subject must not be more than 69 characters long. Strive for about 50 characters.

cableman:
Please rebase the PR, so it can be merged

kasperg:
Closed in favour of #378.

PR #335, opened 2016-09-26
Issue #1606 by Inlead: Change the interestperiod change logic.

bozeak:
http://platform.dandigbib.org/issues/1606

### DIFF
--- a/modules/alma/includes/alma.reservation.inc
+++ b/modules/alma/includes/alma.reservation.inc
@@ -247,7 +247,7 @@ function alma_reservation_create($account, $id, $options = array(), $expiry = NU
     $int_per = variable_get('ding_reservation_disable_interest_date', '');
 
     // Interest period.
-    if ($int_per) {
+    if (empty($int_per)) {
       if (!isset($options['interest_period'])) {
         if (!isset($expiry)) {
           $interest_period = alma_reservation_default_interest_period($profile);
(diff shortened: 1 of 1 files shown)


kasperg:
The Scrutinizer error is not related to the change at hand. I have ignored it.

modules/alma/includes/alma.reservation.inc
```diff
@@ -247,7 +247,7 @@ function alma_reservation_create($account, $id, $options = array(), $expiry = NU
     $int_per = variable_get('ding_reservation_disable_interest_date', '');
 
     // Interest period.
-    if ($int_per) {
```
kasperg: How is this change related to the issue at hand?

I see that we retrieve a value called ding_reservation_*disable*_interest_date which would infer that if the variable has a nonempty value then interest dates should be disabled. The current implementation infers that is is the other way around.

This change reverses this but this would in practice also mean that anyone who has intentionally set `ding_reservation_disable_interest_date` to enable interest dates would have interest perioids enabled.

If this is to be changed then I would like a separate issue for this.

modules/alma/includes/alma.reservation.inc
```diff
@@ -332,7 +332,7 @@ function alma_reservation_update($account, $ids, $options) {
   // Processing IDs. If at least one is updated, return true.
   foreach ($ids as $id) {
     if (isset($reservations[$id])) {
-      $expiry_date = !empty($options['interest_period']) ? alma_reservation_format_date(strtotime($reservations[$id]['valid_from']) + $options['interest_period'] * 86400) : $reservations[$id]['valid_to'];
```
kasperg: :+1:

modules/alma/includes/alma.reservation.inc
bozeak: Yep, this change is not related to current PR. Rolled back.

kasperg: Thanks for the followup @bozeak.

kasperg:
Reviewed and approved. Thanks! Awaiting merge.

PR #338, opened 2016-09-27
1858: Button removed from anons

martinb14:
http://platform.dandigbib.org/issues/1858


JrgenGNielsen:
reviewed and accepted

PR #340, opened 2016-09-28
1925: Sharing of lists improved

martinb14:
http://platform.dandigbib.org/issues/1925


JrgenGNielsen:
reviewed and accepted

PR #346, opened 2016-09-30
1877: Prefix cache with uid and wipe on listupdates

martinb14:
http://platform.dandigbib.org/issues/1877

### DIFF
--- a/modules/p2/ding_serendipity/ding_serendipity.module
+++ b/modules/p2/ding_serendipity/ding_serendipity.module
@@ -212,7 +212,6 @@ function ding_serendipity_get_info() {
 function ding_serendipity_fetch(array $context = NULL) {
   // Allow modules to modify the provided context.
   drupal_alter('serendipity_context', $context);
-
   if (variable_get('ding_serendipity_enable_cache', FALSE)) {
     // Caching of serendipity.
     $cache_id = array();
@@ -221,8 +220,11 @@ function ding_serendipity_fetch(array $context = NULL) {
         $cache_id[] = $ct;
       }
     }
-    $cache_id = "ding_serendipity" . implode(':', $cache_id);
 
+    // Prefix cid with uid for per-user serendipity cache clear.
+    $user_key = isset($context['current-user']) ? ("_" . $context['current-user']) . "_" : "";
+    $cache_id = "ding_serendipity" . $user_key . implode(':', $cache_id);
+    
     $cache = cache_get($cache_id);
     if (!empty($cache) && REQUEST_TIME < $cache->expire) {
       $cache->data['cache'] = REQUEST_TIME . ':' . $cache->expire . '<br/>';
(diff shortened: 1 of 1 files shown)


modules/p2/ding_serendipity/ding_serendipity.module
```diff
@@ -350,6 +352,29 @@ function ding_serendipity_fetch(array $context = NULL) {
...
 /**
+ * Flush serendipity cache of one user.
+ *
+ * @param int $uid
+ *   The user uid.
+ */
+function ding_serendipity_user_cache_clear($uid=NULL) {
+  if ($uid === NULL) {
+    $uid = $GLOBALS['user']->uid;
+  }
```
kasperg: I would much prefer if the parameter to this function was not optional.

1. Making it optional forces us to rely on globals. We not not want to do that. 
2. The only case where the function is used could just as well pass the uid as a parameter.
3. Not making it optional also means we do not have to document behavior if the parameter is omitted in the DocBlock (which is not the case either now)

modules/p2/ding_serendipity/ding_serendipity.module
```diff
@@ -221,8 +220,11 @@ function ding_serendipity_fetch(array $context = NULL) {
         $cache_id[] = $ct;
       }
     }
-    $cache_id = "ding_serendipity" . implode(':', $cache_id);
 
+    // Prefix cid with uid for per-user serendipity cache clear.
+    $user_key = isset($context['current-user']) ? ("_" . $context['current-user']) . "_" : "";
```
kasperg: Nitpicking: What is going on with the placement of `()` in the first branch of the ternary statement? It does not seem like they are needed at all.

modules/p2/ding_serendipity/ding_serendipity.module
```diff
@@ -350,6 +352,25 @@ function ding_serendipity_fetch(array $context = NULL) {
...
+function ding_serendipity_user_cache_clear($uid) {
+  cache_clear_all("ding_serendipity_" . $uid . "_", "cache", TRUE);
+}
+
+/**
+ * Flush serendipity cache of current user if list elements change.
+ */
+function ding_serendipity_entity_presave($entity, $type) {
+  if (isset($entity->type) && $entity->type == "ding_list_element" && user_is_logged_in() && isset($entity->uid)) {
+    ding_serendipity_user_cache_clear($entity->uid);
```
kasperg: This clear the cache for the owner of the element - not the current user as suggested by the DocBlock.

That may be intentional but if that is the case why is there a need to check if the current user is logged in?

I would expect one of the following:

Clear serendipity cache for owner of list elements on when changed by whoever.

```php
if (isset($entity->type) && $entity->type == "ding_list_element" && isset($entity->uid)) {
```

or

Clear serendipity chance for current user when changing list elements:

```php
global $user;
if (isset($entity->type) && $entity->type == "ding_list_element" && !empty($user->uid)) {
```

kasperg:
Reviewed and approved. Thanks! Awaiting merge.

PR #350, opened 2016-10-05
1235: Fix undefined Drupal.settings.ting_search_carousel error

JrgenGNielsen:
Drupal.settings.ting_search_carousel is defined in
template_preprocess_ting_search_carousel(), but it is not called
by p2/ting_carousel

### DIFF
--- a/modules/ting_search_carousel/js/ting_search_carousel.js
+++ b/modules/ting_search_carousel/js/ting_search_carousel.js
@@ -363,6 +363,10 @@
    */
   Drupal.behaviors.ting_search_carousel = {
     attach: function (context, settings) {
+      // template_preprocess_ting_search_carousel() is not called by p2/ting_carousel
+      if (settings.ting_search_carousel === undefined) {
+        return;
+      }
       $.each(settings.ting_search_carousel.carousels, function (id, carousel_settings) {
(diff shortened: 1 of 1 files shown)


modules/ting_search_carousel/js/ting_search_carousel.js
```diff
@@ -363,6 +363,10 @@
    */
   Drupal.behaviors.ting_search_carousel = {
     attach: function (context, settings) {
+      // template_preprocess_ting_search_carousel() is not called by p2/ting_carousel
+      if (settings.ting_search_carousel === undefined) {
+        return;
+      }
```
kasperg: I appreciate the defensive coding but I think this is the wrong way to go about this.

If `P2\ting_carousel` cannot load `ting_search_carousel` properly then it should not load it at all.

I do not know why `P2\ting_carousel` [requires this JS](https://github.com/ding2/ding2/blob/fb1f429ab42d08b068a51aecd57cf776d77db528/modules/p2/ting_carousel/ting_carousel.module#L48-L49) but if it is not used then it should not be loaded at all. Instead `P2\ting_carousel` should only require what is actually needed.

Perhaps @martinb14 or @BirkAndMe can help out here?

DDBdrift:
https://platform.dandigbib.org/issues/1235 | Resolved (tag version)

PR #352, opened 2016-10-06
1270: Fix ill titles

xendk:
Dig out the title from the right place.

http://platform.dandigbib.org/issues/1270


JrgenGNielsen:
reviewed and approved

PR #357, opened 2016-10-07
1565: Update ting_oembed_features

JrgenGNielsen:
Move oembed fields to under abstract.


cableman:
Reviewed and approved

PR #358, opened 2016-10-07
1861: Defined more specific js includes

BirkAndMe:
http://platform.dandigbib.org/issues/1861


JrgenGNielsen:
I suggest we use https://github.com/ding2/ding2/pull/350 - it adds an input check on settings, and has less code duplication.

DDBdrift:
https://platform.dandigbib.org/issues/1861 | Closed

PR #362, opened 2016-10-11
Issue1745

pjohans:


cableman:
The PR has expire in relation to the issues it solved due to the new DDBasic theme

PR #363, opened 2016-10-11
Issue #1404 by Inlead: Handle OG references and port old settings.

inlead:
http://platform.dandigbib.org/issues/1404

### DIFF
--- a/modules/ding_place2book/ding_place2book.install
+++ b/modules/ding_place2book/ding_place2book.install
@@ -69,7 +69,7 @@ function ding_place2book_update_7001(&$sandbox) {
   // Remove field.
   field_delete_field('field_place2book_tickets');
 
-  //Get settings.
+  // Get settings.
   $ding_place2book = variable_get('ding_place2book', array());
   $ding_place2book_event_nodes = variable_get('ding_place2book_event_nodes', array());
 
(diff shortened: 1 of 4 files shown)


guddo:
Hang on with merging this one. Just had a discussion with the business and the service provider (place2book.com), and they would like to port the settings from the existing events too.

We are waiting for test environments and explanations before we are able to continue.

kasperg:
The implementation looks sound to me. I have one question:

You have implemented the migration as a separate module. Why is this?

I would have expected the logic to be implemented as a hook_update. I fear that due to permissions some libraries will not be able to enable this module.

kasperg:
The implementation looks sound to me. I have one question:

You have implemented the migration as a separate module. Why is this?

I would have expected the logic to be implemented as a hook_update. I fear that due to permissions some libraries will not be able to enable this module.

kasperg:
Second thoughts:

- It may be that some manual setup is required before users can carry out a migration. That may explain why it is not a hook_update.
- Libraries which have permissions to enable ding_place2book will also have permissions to enable the migration module.

guddo:
The reason to not do it in hook_update is due to the need of the manual mapping before migration.

Not everyone has permission to enable modules, they would need to call DBC. Should we hook update enable the submodule as a part of this PR?

kasperg:
@guddo: Nice idea regarding the hook update. I added this in f000f11 before merging.

PR #369, opened 2016-10-12
1905: Update field_groups module

xendk:
Upgrade field_groups module (security update), and fix ddbasic to deal with it.


JrgenGNielsen:
reviewed and approved

PR #371, opened 2016-10-13
1890: Added share operation. Limited sharing.

martinb14:
http://platform.dandigbib.org/issues/1890


JrgenGNielsen:
reviewed and accepted

PR #374, opened 2016-10-14
38: Add services_views module

xendk:
Needed by ding_services.


JrgenGNielsen:
reviewed and accepted

PR #384, opened 2016-10-18
#731 by Inlead: Add new sort criteria for opensearch v4.3 or higher.

makcumm:
http://platform.dandigbib.org/issues/731


JrgenGNielsen:
reviewed and approved

PR #388, opened 2016-10-20
1989: add provider mockup to alma

JrgenGNielsen:


DDBdrift:
https://platform.dandigbib.org/issues/1989 | Closed

PR #390, opened 2016-10-21
1749: Reduce path to dibs provider

kasperg:
We no longer need to provide the full module path for includes.


kasperg:
This is a continuation of #249.

JrgenGNielsen:
reviewed and accepted

PR #355, opened 2016-10-07
Issue #1654 by Inlead: Fix issue with duplicated opening_hours block.

bozeak:
http://platform.dandigbib.org/issues/1654

PR #391, opened 2016-10-21
1654 by Inlead: Fix issue with duplicated opening_hours block.

inlead:
http://platform.dandigbib.org/issues/1654


inlead:
https://github.com/ding2/ding2/pull/355

PR #395, opened 2016-10-24
1916: Fix bookmark import when there is no flag module

kasperg:
Without this change the installation process will fail. There is no flag_content table as the flag module is not enabled.


JrgenGNielsen:
reviewed and accepted

PR #396, opened 2016-10-25
1990: Fixed query used to fetch facets

rimi-itk:
This fixes http://platform.dandigbib.org/issues/1990.

Before this fix, the _second page_ of nodes from BPI was used to build the facets filters, but if this list of nodes was empty (when BPI ws contains very few nodes) the filters were not built.


JrgenGNielsen:
reviewed and approved

PR #398, opened 2016-10-26
1992: Add user_consent to openruth and fbs

BirkAndMe:
http://platform.dandigbib.org/issues/1992


cableman:
Reviewed and approved

PR #401, opened 2016-10-28
1671: Ensure ting_covers_addi is activated on fresh install

holt83:
http://platform.dandigbib.org/issues/1671#change-19017

### DIFF
--- a/modules/ting_covers/ting_covers.install
+++ b/modules/ting_covers/ting_covers.install
@@ -5,6 +5,14 @@
  */
 
 /**
+ * Implements hook_install().
+ */
+function ting_covers_install() {
+  // Ensure that a cover provider is active on a fresh install.
+  module_enable(array('ting_covers_addi'));
+}
+
+/**
  * Changed fetched covers from managed to unmanaged files.
  */
 function ting_covers_update_7100(&$sandbox) {


modules/ding_ting_frontend/ding_ting_frontend.info
```diff
@@ -19,7 +19,7 @@ dependencies[] = page_manager
 dependencies[] = pm_existing_pages
 dependencies[] = strongarm
 dependencies[] = ting
-dependencies[] = ting_covers
```
kasperg: Changing the dependency to `ting_covers_addi` is problematic. It compares to adding a dependency to a provider implementation like `alma`.

- The dependency is misleading. `ding_ting_frontend` depends on `ting_covers` - not the addi covers provider.
- It will in effect prevent sites from disabling the addi integration even though they want to use some other cover provider.

If the module should be enabled by default on install then it should be added to `ding_ting_frontend_install()`.

holt83: I see what you mean.

Just noticed that ting_covers_addi is activated on update in ting_covers, so wouldn't it be better to also ativate it on install in ting_covers?

Also the update hook which activates it checks if the variable addi_wsdl_url is set. This will never be set on a fresh install, so should we just activate it anyway?

GitCop:

 * Commit: 83138d274232440e14c692450d7aa2a03c431fd0

 * Commit: 3c9d943d47d87a228325b8a4268cc075857d936a

---

GitCop:

 * Commit: 83138d274232440e14c692450d7aa2a03c431fd0

 * Commit: 3c9d943d47d87a228325b8a4268cc075857d936a

---

kasperg:
@holt83 Please ensure that the commit message has the correct format. Leave out the name. We have your GitHub profile anyway.

In your case it should be:

`1671: Ensure ting_covers_addi is activated on fresh install`

kasperg:
Reviewed, approved and merged. Thanks!

PR #405, opened 2016-11-03
2007 by Inlead: Missing argument from commit f6e05762.

inlead:
http://platform.dandigbib.org/issues/2007

Original commit:
https://github.com/ding2/ding2/commit/f6e05762


cableman:
Reviewed and approved

PR #406, opened 2016-11-03
287 by Inlead: Various fixes to campaign module.

inlead:
http://platform.dandigbib.org/issues/287


JrgenGNielsen:
reviewed and approved

PR #407, opened 2016-11-04
605 by Inlead: Check for set interest date.

inlead:
http://platform.dandigbib.org/issues/605


cableman:
Reviewed and approved

PR #408, opened 2016-11-07
1380: Ensure Free text is shown for events

holt83:
http://platform.dandigbib.org/issues/1380


themes/ddbasic/templates/views/ding_event/views-view-field--ding-event--field-ding-event-price.tpl.php
```diff
@@ -19,5 +19,5 @@
   * regardless of any changes in the aliasing that might happen if
   * the view is modified.
   */
-$price = empty($output) ? t('Free') : $output;
+$price = empty($output) || $output ? t('Free') : $output;
```
JrgenGNielsen: This will always return t('Free') - is it a leftover from debugging?

JrgenGNielsen:
reviewed and accepted

kasperg:
Merged. Thanks!

PR #409, opened 2016-11-08
2011: Translate allowed values

BirkAndMe:
http://platform.dandigbib.org/issues/2011


cableman:
Reviewed and approved

PR #415, opened 2016-11-15
2003: Added default consent messages

BirkAndMe:
http://platform.dandigbib.org/issues/2003


JrgenGNielsen:
reviewed and approved

PR #418, opened 2016-11-15
2017: Always show username on public lists

BirkAndMe:
http://platform.dandigbib.org/issues/2017


JrgenGNielsen:
reviewed and approved

PR #420, opened 2016-11-17
2019: Show shared lists when token given

BirkAndMe:
http://platform.dandigbib.org/issues/2019


JrgenGNielsen:
reviewed and approved

PR #422, opened 2016-11-17
1859: Share of private list flow

BirkAndMe:
http://platform.dandigbib.org/issues/1859


JrgenGNielsen:
reviewed and approved

PR #425, opened 2016-11-17
2035: Fixed titles in serendipity info

martinb14:
http://platform.dandigbib.org/issues/2035


JrgenGNielsen:
reviewed and accepted

PR #427, opened 2016-11-21
1892: Removed check on list save

BirkAndMe:
http://platform.dandigbib.org/issues/1892


JrgenGNielsen:
reviewed and accepted

PR #428, opened 2016-11-21
2041: Access to loan history list

BirkAndMe:
http://platform.dandigbib.org/issues/2041


JrgenGNielsen:
reviewed and accepted

PR #430, opened 2016-11-24
430 by Inlead: Remove rendering opening hours on search result.

inlead:


DDBdrift:
https://platform.dandigbib.org/issues/430 | Resolved (tag version)

PR #431, opened 2016-11-24
2021: Note on share list form

BirkAndMe:
http://platform.dandigbib.org/issues/2021


JrgenGNielsen:
reviewed and approved

PR #435, opened 2016-11-25
B2020

BirkAndMe:
http://platform.dandigbib.org/issues/2020


JrgenGNielsen:
reviewed and accepted

modules/ding_permissions/ding_permissions.module
```diff
@@ -508,6 +508,11 @@ function ding_permissions_secure_permissions($role) {
       'view the administration theme',
     ),
     'editor' => array(
+      'eck add ding_type ding_interaction entities',
```
guddo: What does "eck" mean?

martinb14: Eck is the Entity Construction Kit module.

PR #436, opened 2016-11-25
2033: Remove local missing materials from serendipity

martinb14:
http://platform.dandigbib.org/issues/2033


JrgenGNielsen:
reviewed and accepted

PR #439, opened 2016-12-01
2064: Reverse bookmarks import order

BirkAndMe:
http://platform.dandigbib.org/issues/2064


JrgenGNielsen:
reviewed and approved

PR #440, opened 2016-12-01
2055: Removed install argument

BirkAndMe:
http://platform.dandigbib.org/issues/2055


JrgenGNielsen:
reviewed and approved

PR #442, opened 2016-12-01
1997 by Inlead: Re-enable toolbar module.

inlead:
http://platform.dandigbib.org/issues/1997


JrgenGNielsen:
reviewed and approved

PR #443, opened 2016-12-01
1890: Private sharer checks list permissions

BirkAndMe:
http://platform.dandigbib.org/issues/1890


JrgenGNielsen:
reviewed and approved

PR #446, opened 2016-12-02
1882: Carousel item width

BirkAndMe:
http://platform.dandigbib.org/issues/1882


DDBdrift:
https://platform.dandigbib.org/issues/1882 | Resolved (tag version)

PR #448, opened 2016-12-02
2065: serendipity adhr output

BirkAndMe:
http://platform.dandigbib.org/issues/2065


JrgenGNielsen:
reviewed and approved

PR #452, opened 2016-12-05
2072: Fix metatag version in make

holt83:
http://platform.dandigbib.org/issues/2072


cableman:
Reviewed and apporved

PR #456, opened 2016-12-06
2074: Added extra values to bpi data

rimi-itk:
In the near future a couple of new fields, url and data, will be added to the BPI web service. In order to not break running sites when these are added (the bpi client makes all fields required), we add the fields now and they will simply be ignored by the web service until it's updated.

http://platform.dandigbib.org/issues/2074


cableman:
Reviewed and approved

PR #458, opened 2016-12-07
2083: Don't cache pages with messages in varnish

xendk:
As described in the title..


cableman:
Reviewed and approved

PR #459, opened 2016-12-08
2005 by Inlead: Remove providers from Staff list.

bozeak:
Here are exported changes for the next issues:

- http://platform.dandigbib.org/issues/2005
- http://platform.dandigbib.org/issues/2014


cableman:
Reviewed and approved

PR #462, opened 2016-12-09
Feature/2013 responsive notifications menu

JrgenGNielsen:
trying to compile a css that CircleCI will accept


JrgenGNielsen:
see also: https://github.com/ding2/ding2/pull/447 - only, here the css compiles to correspond to CircleCI's setup

PR #465, opened 2016-12-12
174 by Inlead: Add classification to overall display.

bozeak:
http://platform.dandigbib.org/issues/174


modules/ting/ting.field.inc
```diff
@@ -481,6 +498,18 @@ function ting_field_formatter_view($entity_type, $entity, $field, $instance, $la
         }
         break;
 
+      case 'ting_literary_form_default':
+        if (!empty($entity->reply->record['dc:subject']['dkdcplus:DK5'][0])) {
```
NicolaeGrigoriu: It is good, but the wrapper function **getClassification** also could be used. https://github.com/easySuite/ding2/blob/master/modules/ting/ting.entities.inc#L67.

So, overall construction can have some form
`if ($classification = getClassification()) {
   $literary_form = (!$classification || $classification == '86') ? t('Fiction') : t('Non-Fiction');
  // Your rest code goes here.
}`

guddo:
Moved to https://github.com/ding2/ding2/pull/482.

PR #467, opened 2016-12-13
2019: Ding list support new mail system

BirkAndMe:
http://platform.dandigbib.org/issues/2019

### DIFF
--- a/modules/p2/ding_list/ding_list.module
+++ b/modules/p2/ding_list/ding_list.module
@@ -828,10 +828,9 @@ function ding_list_mail($key, &$message, $params) {
         '@user' => $username,
         '@library' => variable_get('site_name', ''),
       ));
-      $message['body'][] = nl2br(t('Please use the following link to access this list.
-The link is active for 7 days from today, after which it will be disabled for security reasons.
-
-<a href="!link">!link</a>', array('!link' => $params['link'])));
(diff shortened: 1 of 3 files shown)


modules/p2/ding_list/ding_list.module
```diff
@@ -828,10 +828,9 @@ function ding_list_mail($key, &$message, $params) {
         '@user' => $username,
         '@library' => variable_get('site_name', ''),
       ));
-      $message['body'][] = t('Please use the following link to access this list.
-The link is active for 7 days from today, after which it will be disabled for security reasons.
-
-!link', array('!link' => $params['link']));
+      $message['body'][] = nl2br(t("Please use the following link to access this list.\nThe link is active for 7 days from today, after which it will be disabled for security reasons.\n\n!link.", array(
+        '!link' => l($params['link'], $params['link'], array('absolute' => TRUE)),
```
kasperg: The same value `$params['link']` is used both for the text and the url for the link. That seems odd to me.

The use of `array('absolute' => TRUE))` indicates that it is not a full url already and showing a path to the user does not make much sense.

I see from [where this is called](https://github.com/ding2/ding2/blob/master/modules/p2/ding_list/include/form.inc#L650) that the `$params['link']` is already an absolute url so making it absolute again is unnecessary. 

In general calling a variable `$link` containing an url is confusing.

kasperg:
Thanks. Merged.

PR #468, opened 2016-12-13
2063 by Inlead: Export to feature.

guddo:
http://platform.dandigbib.org/issues/2063

PR #469, opened 2016-12-13
2063 by Inlead: Changes wrong function.

inlead:
http://platform.dandigbib.org/issues/2063


inlead:
Merged in https://github.com/ding2/ding2/pull/468.

PR #471, opened 2016-12-14
Feature/2038 ding interaction padding

JrgenGNielsen:
trying to compile a css that CircleCI will accept
see also: #445 - only, here the css compiles to correspond to CircleCI's setup


JrgenGNielsen:
reviewed and approved

PR #480, opened 2016-12-16
1618: Blacklisting of branches.

inlead:
http://platform.dandigbib.org/issues/1912

### DIFF
--- a/modules/fbs/fbs.module
+++ b/modules/fbs/fbs.module
@@ -212,7 +212,7 @@ function fbs_settings_form() {
     '#type' => 'actions',
     'submit' => array(
       '#type' => 'submit',
-      '#value' => t('Update list')
+      '#value' => t('Update list'),
     ),
   );
 
--- a/modules/fbs/fbs.module
+++ b/modules/fbs/fbs.module
@@ -581,3 +581,22 @@ function _fbs_wayf_login($name, $value = NULL) {
 
   return NULL;
 }
+
+/**
+ * Implements hook_ting_pre_execute().
+ *
+ * Filter out blacklisted branches from search.
+ */
+function fbs_ting_pre_execute($request) {
+  // We only want to modify searches.
+  if (gettype($request) != 'TingClientSearchRequest') {
+    return;
+  }
+  $fbs_branches_blacklist = variable_get('fbs_branches_blacklist', array());
+  $query = $request->getQuery();
+  if (!empty($fbs_branches_blacklist)) {
+    $query .= ' AND (holdingsitem.branch = (* not ' . implode(' not ',
+        $fbs_branches_blacklist) . '))';
+    $request->setQuery($query);
+  }
+}
--- a/modules/fbs/fbs.module
+++ b/modules/fbs/fbs.module
@@ -198,6 +198,42 @@ function fbs_settings_form() {
     '#description' => t('Allow users to delete their reservations as well as ready for pickup ones.'),
   );
 
+  // Update list of branches.
+  $form['update'] = array(
+    '#type' => 'fieldset',
+    '#title' => t('Update from library system'),
+  );
+
+  $form['update']['fbs_branches_last_updated'] = array(
+    '#markup' => '<label>Last updated:</label>' . (variable_get('fbs_branches_last_updated', 0) ? date('Y-m-d H:i:s', variable_get('fbs_branches_last_updated')) : '<em>' . t('Never') . '</em>'),
+  );
+
+  $form['update']['fbs_update'] = array(
+    '#type' => 'actions',
+    'submit' => array(
+      '#type' => 'submit',
+      '#value' => t('Update list')
+    ),
+  );
+
+  // Branches and blacklisted branches.
+  $form['branches'] = array(
+    '#type' => 'fieldset',
+    '#title' => t('Branches Blacklist'),
+    '#description' => t('Choose which branches to blacklist from searches'),
+  );
+  $fbs_branches = variable_get('fbs_branches', array());
+  $fbs_branches_blacklist = variable_get('fbs_branches_blacklist', array());
+  foreach ($fbs_branches as $id => $branch) {
+    $form['branches']['fbs_branch_' . $id] = array(
+      '#type' => 'checkbox',
+      '#title' => check_plain($branch),
+      '#default_value' => isset($fbs_branches_blacklist[$id]) ? TRUE : FALSE,
+    );
+  }
+
+  $form['#submit'][] = 'fbs_settings_form_submit';
+
   return system_settings_form($form);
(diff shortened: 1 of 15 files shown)


modules/fbs/fbs.module
```diff
@@ -520,3 +623,22 @@ function _fbs_wayf_login($name, $value = NULL) {
...
+    return;
+  }
+  $fbs_branches_blacklist = variable_get('fbs_branches_blacklist', array());
+  $query = $request->getQuery();
+  if (!empty($fbs_branches_blacklist)) {
+    $query .= ' AND (holdingsitem.branch = (* not ' . implode(' not ',
+        $fbs_branches_blacklist) . '))';
+    $request->setQuery($query);
+  }
+}
```
holt83: I think this need to be:
$query.= '(' . $query . ') AND (holdingsitem.branch = (* not ' . implode(' not ', $fbs_branches_blacklist) . '))';
To ensure proper evaluation of the query.

modules/fbs/fbs.module
```diff
@@ -520,3 +623,22 @@ function _fbs_wayf_login($name, $value = NULL) {
...
+/**
+ * Implements hook_ting_pre_execute().
+ *
+ * Filter out blacklisted branches from search.
+ */
+function fbs_ting_pre_execute($request) {
+  // We only want to modify searches.
+  if (gettype($request) != 'TingClientSearchRequest') {
+    return;
+  }
```
holt83: Will gettype work here? Look at possible values in documentation: http://php.net/manual/en/function.gettype.php

Think it's better to use is_a or instanceof.

Also are we targeting TingCollectionRequest also? If not get_class should be used instead.

modules/fbs/fbs.module
```diff
@@ -198,6 +198,61 @@ function fbs_settings_form() {
...
+  $form['branches'] = array(
+    '#type' => 'fieldset',
+    '#title' => t('Holdings Branches Blacklist'),
+    '#description' => t('Choose which branches to blacklist from searches'),
+  );
+  $fbs_branches = variable_get('fbs_branches', array());
+  $fbs_branches_blacklist = variable_get('fbs_branches_blacklist', array());
+  foreach ($fbs_branches as $id => $branch) {
+    $form['branches']['fbs_branch_' . $id] = array(
+      '#type' => 'checkbox',
```
xendk: This should be a "checkboxes" element.

modules/fbs/fbs.module
```diff
@@ -208,6 +263,46 @@ function fbs_settings_form_submit($form, &$form_state) {
...
+  // Blacklisted branches for holdings/availability.
+  $fbs_branches = variable_get('fbs_branches', array());
+  $fbs_branches_blacklist = array();
+  foreach ($form_state['values'] as $key => $value) {
+    $matches = array();
+    if (preg_match('/^fbs_branch_(\w+-\d+)$/', $key, $matches)) {
+      $branch_id = $matches[1];
+      if (isset($fbs_branches[$branch_id]) && $value) {
+        if ($value) {
+          $fbs_branches_blacklist[$branch_id] = $branch_id;
```
xendk: This should be about two lines with "checkboxes" and (possibly) array_filter().

modules/fbs/fbs.module
```diff
@@ -341,6 +436,14 @@ function fbs_branches() {
     }
 
     asort($branches);
+
+    // Filtering blacklisted branch if such exists.
+    $blacklisted = variable_get('fbs_user_branches_blacklist', array());
+    if (!empty($blacklisted)) {
```
xendk: If statement not needed, foreach will loop zero times on an zero length array.

modules/fbs/fbs.module
xendk: gettype wont work.

Why shouldn't we target collections too?

modules/fbs/fbs.module
xendk: And elements in `$fbs_branches_blacklist` should be quoted, as `holdingsitem.branch` apparently is the string (ref: http://platform.dandigbib.org/issues/1618#note-31)

Which also means that `$fbs_branches_blacklist` cannot be used directly, but needs to be mapped from IDs to names. Alternatively, `$fbs_branches_blacklist` is an ID => name array, and this uses array_values() while the other uses array_keys().

modules/fbs/fbs.module
```diff
@@ -198,6 +198,61 @@ function fbs_settings_form() {
...
+    ),
+  );
+
+  // Branches and blacklisted branches.
+  $form['branches'] = array(
+    '#type' => 'fieldset',
+    '#title' => t('Holdings Branches Blacklist'),
+    '#description' => t('Choose which branches to blacklist from searches'),
+  );
+  $fbs_branches = variable_get('fbs_branches', array());
```
xendk: No...

Use `fbs_branches()`, and either kill the update button (why do we need it?) or implement caching in `fbs_branches()`.

modules/fbs/fbs.module
```diff
@@ -198,6 +198,61 @@ function fbs_settings_form() {
...
+    '#title' => t('User Branches Blacklist'),
+    '#description' => t('Choose which branches to blacklist from user profile.'),
+  );
+
+
+  $fbs_user_branches_blacklist = variable_get('fbs_user_branches_blacklist', array());
+
+  foreach ($fbs_branches as $bid => $user_branch) {
+    $form['user_branches']['fbs_user_branch_' . $bid] = array(
+      '#type' => 'checkbox',
```
xendk: Same.

modules/fbs/fbs.module
```diff
@@ -198,6 +198,61 @@ function fbs_settings_form() {
...
+      '#title' => check_plain($user_branch),
+      '#default_value' => isset($fbs_user_branches_blacklist[$bid]) ? TRUE : FALSE,
+    );
+  }
+
+  $a = 1;
+
+  $form['#submit'][] = 'fbs_settings_form_submit';
```
xendk: Unnecessary.

modules/fbs/fbs.module
xendk: Ah, just spotted that `fbs_branches()` filters the branches. It shouldn't.

Rather, there should be a `fbs_reservation_branches()` that returns the filtered list, which should be used where it matters.

modules/fbs/fbs.module
holt83: You're right: We should target collections too. Don't know what I was thinking about when I wrote that.

I guess the best option is instanceof then: http://stackoverflow.com/questions/3017684/what-is-the-difference-between-is-a-and-instanceof

xendk:
@kasperg superseeded by #576

PR #481, opened 2016-12-16
2089 by Inlead: Do not flush caches on update_hooks.

inlead:
http://platform.dandigbib.org/issues/2089


inlead:
Patch not needed.

PR #483, opened 2016-12-20
1527: Redirect to registration page for unknown WAYF logins

arnested:
http://platform.dandigbib.org/issues/1527

Redirect unknown users logging in via WAYF to the registration page.

We avoid redirecting to the front page (as it used to be) because messages will are lost on sites that strip sessions cookies from the front page.


JrgenGNielsen:
reviewed and approved

PR #486, opened 2016-12-20
430: Fix search_api warnings and notices

holt83:


cableman:
Reviewed and approved

PR #487, opened 2016-12-20
1404 by Inlead: Fixes displaying of prices and of reserve button.

avoykov:
http://platform.dandigbib.org/issues/1404


JrgenGNielsen:
reviewed and approved

PR #488, opened 2016-12-20
430: Fix patch on wrong module

holt83:


JrgenGNielsen:
Reviewed and approved

PR #494, opened 2016-12-23
2002 by Inlead: Cleanup inactive users.

AnatolyMuntean:
http://platform.dandigbib.org/issues/2002

### DIFF
--- a/modules/ding_provider/ding_provider.admin.inc
+++ b/modules/ding_provider/ding_provider.admin.inc
@@ -27,5 +27,19 @@ function ding_provider_admin_overview() {
     }
   }
 
-  return $build;
+  $build['wipe_provider_users_threshold'] = array(
+    '#type' => 'select',
+    '#title' => t('Delete provider users that are inactive for:'),
+    '#options' => array(
+      '0' => t('None (keep all users)'),
+      '31536000' => t('1 year'),
+      '63072000' => t('2 years'),
+      '94608000' => t('3 years'),
+      '126144000' => t('4 years'),
+      '157680000' => t('5 years'),
+    ),
+    '#default_value' => variable_get('wipe_provider_users_threshold', 0)
+  );
+
+  return system_settings_form($build);
 }
--- a/modules/ding_provider/ding_provider.admin.inc
+++ b/modules/ding_provider/ding_provider.admin.inc
@@ -38,7 +38,7 @@ function ding_provider_admin_overview() {
       '126144000' => t('4 years'),
       '157680000' => t('5 years'),
     ),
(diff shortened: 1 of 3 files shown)


modules/ding_provider/ding_provider.admin.inc
```diff
@@ -27,5 +27,19 @@ function ding_provider_admin_overview() {
     }
   }
 
-  return $build;
+  $build['wipe_provider_users_threshold'] = array(
```
cableman: I would like this variable to be prefixed with module name (ding_provider_users_wipe_threshold). It would make it easier to debug later on in the DB.

modules/ding_provider/ding_provider.admin.inc
```diff
@@ -27,5 +27,19 @@ function ding_provider_admin_overview() {
...
+    '#title' => t('Delete provider users that are inactive for:'),
+    '#options' => array(
+      '0' => t('None (keep all users)'),
+      '31536000' => t('1 year'),
+      '63072000' => t('2 years'),
+      '94608000' => t('3 years'),
+      '126144000' => t('4 years'),
+      '157680000' => t('5 years'),
+    ),
+    '#default_value' => variable_get('wipe_provider_users_threshold', 0),
```
cableman: This needs to be renamed as well (ding_provider_users_wipe_threshold).

modules/ding_provider/ding_provider.module
```diff
@@ -763,3 +764,36 @@ function ding_provider_entity_info_alter(&$entity_info) {
...
+/**
+ * Implements hook_cron().
+ */
+function ding_provider_cron() {
+  $last_cron_time = variable_get('ding_provider_last_cron', 0);
+
+  // Run daily.
+  if ($last_cron_time < REQUEST_TIME - 86400) {
+    // Remove inactive users.
+    $inactivity_threshold = variable_get('wipe_provider_users_threshold', 0);
```
cableman: This needs to be renamed as well (ding_provider_users_wipe_threshold).

modules/ding_provider/ding_provider.module
```diff
@@ -763,3 +764,36 @@ function ding_provider_entity_info_alter(&$entity_info) {
...
+        ->fields('u', array('uid'))
+        ->condition('u.access', REQUEST_TIME - $inactivity_threshold, '<')
+        ->range(0, 500);
+
+      $result = $query
+        ->execute()
+        ->fetchCol();
+
+      if (!empty($result)) {
+        // TODO: Maybe rework as a queue.
```
cableman: I think that would be a very good idea to implement this as a queue it might be a very large batch that have to be deleted.

cableman:
Reviewed and approved

PR #495, opened 2016-12-28
2039 by Inlead: Ability to translate error codes from alma.

AnatolyMuntean:
http://platform.dandigbib.org/issues/2039


guddo:
@xendk: What do you propose instead? The information we got is that the strings can vary from system to system.

xendk:
http://platform.dandigbib.org/issues/2039

cableman:
Development on the alma provider has been cancelled in DDB CMS

PR #498, opened 2017-01-03
2106: Background on p2 panes

BirkAndMe:
http://platform.dandigbib.org/issues/2106


kasperg:
@BirkAndMe Circle CI fails because there are differences between the compiled versions on the reference platform and your committed code. You can see the differences directly on Circle CI. 

The versions of the SCSS compilation pipeline does not append `false` to `-moz-transition` and `-o-transition` and does not add a `-pie-background` declaration.

The problem could be in the version of `compass-mixins` that you use. [`npm-shrinkwrap` should ensure that you use 0.12.7](https://github.com/ding2/ding2/blob/master/themes/ddbasic/npm-shrinkwrap.json#L7).

image

kasperg:
Closed in favour of #507.

PR #463, opened 2016-12-09
1685: DDBasic V2.0

martinb14:
http://platform.dandigbib.org/issues/1685
Note an updated feature export is not included yet!

PR #501, opened 2017-01-06
2117 by Inlead: Adjusted Event library view.

inlead:
http://platform.dandigbib.org/issues/2117


kasperg:
Changes in this PR is obsolete with the new theme from #463.

PR #502, opened 2017-01-09
2112: Get average rating from openlist

BirkAndMe:
http://platform.dandigbib.org/issues/2112

### DIFF
--- a/modules/p2/ding_entity_rating/ding_entity_rating.module
+++ b/modules/p2/ding_entity_rating/ding_entity_rating.module
@@ -345,14 +345,18 @@ function ding_entity_rating_build_theme_list($rating) {
  *   The average rating of the given entity.
  */
 function ding_entity_rating_get_average($entity) {
-  $query = db_select('field_data_field_weight', 'w');
-  $query->join('eck_ding_type', 'dt', 'w.entity_id = dt.id');
-  $query->join('field_data_field_value', 'v', 'w.entity_id = v.entity_id');
-  $query->condition('v.field_value_value', $entity->ding_entity_id)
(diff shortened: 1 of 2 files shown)


BirkAndMe:
Sure, we'll implement a static cache, setting the rating using a `ding_entity_rating_set_average()` function, instead of using the object as a container.

kasperg:
@BirkAndMe :+1:

modules/p2/ding_entity_rating/ding_entity_rating.module
```diff
@@ -345,14 +345,54 @@ function ding_entity_rating_build_theme_list($rating) {
...
-
-  return $result->fetchField(0);
+  return ding_entity_rating_average($entity);
+}
+
+/**
+ * Set the average rating of a ting object entity.
+ *
+ * @param object $entity
+ *   The ting object entity to remove from the rating list.
```
kasperg: Nitpicking here but this bit of DocBlock seems copy/pasted. Could you please update it?

modules/p2/ding_entity_rating/ding_entity_rating.module
```diff
@@ -345,14 +345,54 @@ function ding_entity_rating_build_theme_list($rating) {
...
+ */
+function ding_entity_rating_set_average($entity, $value) {
+  return ding_entity_rating_average($entity, $value);
+}
+
+/**
+ * Get / Set average rating of a ting object.
+ *
+ * @param object $entity
+ *   The ting object entity to remove from the rating list.
```
kasperg: As above.

PR #503, opened 2017-01-10
2109: Serendipity links

BirkAndMe:
http://platform.dandigbib.org/issues/2109


kasperg:

PR #510, opened 2017-01-12
B2111

BirkAndMe:
http://platform.dandigbib.org/issues/2111


cableman:
Change look fine, but could I get you to run an rebase and squash the commits into one.

See http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

PR #513, opened 2017-01-19
2134: Styled interaction for lazy pane

BirkAndMe:
http://platform.dandigbib.org/issues/2134


modules/p2/ding_interaction/plugins/content_types/interaction_pane.inc
```diff
@@ -67,6 +67,7 @@ function ding_interaction_content_type_render($subtype, $conf, $panel_args, $con
       'help_text' => $help_text,
     ));
     $output->css_class = isset($data['css_class']) ? $data['css_class'] : "default-interaction";
+    $output->css_class .= ' ding-interaction-true-pane';
```
kasperg: The name `ding-interaction-true-pane` is non-obvious to me.

If adding this class is required because we use lazy-panes then please add a comment about it.

PR #514, opened 2017-01-19
Issue 1736: Save pincode in user_creds

pjohans:
Unlike alma and openruth fbs does NOT save pincode in user creds. Thomas may have a way of doing it in a different way, but this change makes infomedia work.

### DIFF
--- a/modules/fbs/includes/fbs.user.inc
+++ b/modules/fbs/includes/fbs.user.inc
@@ -62,6 +62,7 @@ function fbs_user_authenticate($name, $pincode) {
     // We have to save the library card number/CPR number, as the pincode
     // change call uses it.
     $result['creds']['library_card_number'] = $name;
+    $result['creds']['pass'] = $pincode;
 
     $result['user'] = array(
       'mail' => $res->patron->emailAddress,


modules/fbs/includes/fbs.user.inc
```diff
@@ -62,6 +62,7 @@ function fbs_user_authenticate($name, $pincode) {
     // We have to save the library card number/CPR number, as the pincode
     // change call uses it.
     $result['creds']['library_card_number'] = $name;
+    $result['creds']['pass'] = $pincode;
```
kasperg: @xendk: Do you have any input on this?

kasperg: I do not think it is a good idea to store the pincode in the creds. It is not required by the integration, the provider in itself or any provider related functionality like changing password mentioned above. It also means that any attacker getting access to the session storage (database) will have usernames and passwords for all users with a session.

On the other hand I do not have any other solution. If we go with this then please add a comment explaining why this is necessary.

@xendk: Do you have any input on this?

xendk: For starters I think it's bad form for anyone but the provider in use to dig around in $creds and making assumptions about it's structure.

This also lessens security for sites not using infomedia.

I'd rather se a ding_user_get_credentials() that infomedia can call which might throw up a password prompt if the password isn't available.

DDBdrift:
https://platform.dandigbib.org/issues/1736 | Resolved (tag version)

PR #515, opened 2017-01-20
2022: Serendipity frontpage

BirkAndMe:
http://platform.dandigbib.org/issues/2022

### DIFF
--- a/modules/p2/ding_serendipity/plugins/content_types/serendipity_ting_object.inc
+++ b/modules/p2/ding_serendipity/plugins/content_types/serendipity_ting_object.inc
@@ -32,7 +32,7 @@ function ding_serendipity_serendipity_ting_object_admin_info($subtype, $conf, $c
 function ding_serendipity_serendipity_ting_object_content_type_render($subtype, $conf, $panel_args, $context) {
   // Only show on front if serendipity is enabled for the frontpage.
   // See ding_frontpage config at /admin/config/ding/frontend.
-  if (drupal_is_front_page() && !variable_get('ding_frontpage_serendipity') && variable_get('ding_frontpage_promoter') != 'serendipity') {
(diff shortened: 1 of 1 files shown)


modules/p2/ding_serendipity/plugins/content_types/serendipity_ting_object.inc
```diff
@@ -32,7 +32,7 @@ function ding_serendipity_serendipity_ting_object_admin_info($subtype, $conf, $c
 function ding_serendipity_serendipity_ting_object_content_type_render($subtype, $conf, $panel_args, $context) {
   // Only show on front if serendipity is enabled for the frontpage.
   // See ding_frontpage config at /admin/config/ding/frontend.
-  if (drupal_is_front_page() && !variable_get('ding_frontpage_serendipity') && variable_get('ding_frontpage_promoter') != 'serendipity') {
+  if (isset($conf['selected_keys']['frontpage']) && !variable_get('ding_frontpage_serendipity') && variable_get('ding_frontpage_promoter') != 'serendipity') {
```
kasperg: I do not understand why this fixes the issue: "The serendipity carrousel is shown on the frontpage even if it is not activated". As I understand it the carrousel is rendered by this function. If it returns NULL a carrousel is not rendered.

1. The current page is the Drupal frontpage AND
2. The user has not enabled serendipity on the frontpage AND
3. The selected promoter is not serendipity

The boolean logic seems good to me. I cannot see how the changing part 1 so we do not check whether we are on the frontpage changes this.

kasperg: @BirkAndMe has explained that this is caused by panes loaded through lazy_pane are not aware of where they are being inserted. Thus we test for a setting instead.

PR #516, opened 2017-01-20
2138 by Inlead: Remove space and changing logic on update action.

avoykov:
http://platform.dandigbib.org/issues/2138

### DIFF
--- a/modules/ding_place2book/ding_place2book.fields.inc
+++ b/modules/ding_place2book/ding_place2book.fields.inc
@@ -572,7 +572,13 @@ function _ding_place2book_update_p2b_entities($entity, array $settings) {
     }
     catch (Exception $ex) {
       watchdog_exception('ding_place2book', $ex);
-      $items = array();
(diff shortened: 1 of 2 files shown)


modules/ding_place2book/ding_place2book.fields.inc
```diff
@@ -572,7 +572,13 @@ function _ding_place2book_update_p2b_entities($entity, array $settings) {
...
     catch (Exception $ex) {
       watchdog_exception('ding_place2book', $ex);
-      $items = array();
+      $items = array(
+        array(
+          'event_id' => $event_id,
+          'event_maker_id' => $event_maker_id,
+          'synchronize' => 1,
+        ),
+      );
```
kasperg: `$items` is duplicated on L563.

If we always want to return the value then it should be created once. Based on the position of the `try` and where the exception is thrown there is no guarantee that `$event_id` and `$event_maker_id` has been initialized.

PR #520, opened 2017-01-26
1404: Revert place2book upgrade

kasperg:
This change has caused several issues which cannot be fixed quickly.

We want to finish the release so try to revert it.


kasperg:
Superseeded by #523.

PR #521, opened 2017-01-29
2146: Fix default place2book URLs

holt83:
http://platform.dandigbib.org/issues/2146


kasperg:
Place2book configuration has changed significantly since this PR. I will close it as it is now obsolete.

PR #525, opened 2017-02-01
2160: Remove the bookmarks page

BirkAndMe:
http://platform.dandigbib.org/issues/2160


modules/p2/ding_list/ding_list.install
```diff
@@ -37,6 +37,12 @@ function ding_list_install() {
   $mail_settings = variable_get('mail_system', array());
   $mail_settings['ding_list_share'] = "HTMLMailSystem";
   variable_set('mail_system', $mail_settings);
+
+  module_load_include('inc', 'page_manager', 'plugins/tasks/page');
+  $page = page_manager_page_load('user_bookmarks');
+  if (!empty($page)) {
+    page_manager_page_delete($page);
+  }
```
kasperg: Why do we need to delete the page on install?

As the configuration is deleted from the feature export I cannot see how there would be anything to delete when new sites are installed.

modules/p2/ding_list/ding_list.install
```diff
@@ -54,3 +60,14 @@ function ding_list_update_7001() {
...
+
+/**
+ * Remove the bookmarks page, if it's been overridden.
+ */
+function ding_list_update_7002() {
+  module_load_include('inc', 'page_manager', 'plugins/tasks/page');
+  $page = page_manager_page_load('user_bookmarks');
+  if (!empty($page)) {
+    page_manager_page_delete($page);
+  }
```
kasperg: I am unsure whether the `ding_list` module is the right place for this update hook since the page being deleted was defined by `ding_user_frontend`. On the other hand this is the module which makes the other page obsolete.

If this comment has not stirred anything then leave this be.

modules/p2/ding_list/ding_list.install
BirkAndMe: If ding_list is enabled on a site, where the `user_bookmarks` page has been altered, that page is not removed when the feature no longer exists.

So when p2 is enabled on older sites the page might "survive". One could argue that if someone has customized the page, they might not want it deleted. So we can remove this code, I don't feel strongly about it.

kasperg: Ah right - installing the module is not the same as installing the site. Thanks for the explanation.

PR #527, opened 2017-02-02
Issue2164 gatewayf

pjohans:
This commit depends on an updated gatewayf - gatewayf now holds a hostlist with a salt for each host. Also gatewayf returns a checksum from selected attributes which are recreated and compared on login

### DIFF
--- a/modules/ding_gatewayf/ding_gatewayf.module
+++ b/modules/ding_gatewayf/ding_gatewayf.module
@@ -285,9 +285,11 @@ function ding_gatewayf_ding_provider_user() {
 /**
  * Do a check on the checksum given from gatewayf.
  *
- * To protect against malicious attempts to login through gatewayf the response contains a checksum which can be used to verify the data:
+ * To protect against malicious attempts to login through gatewayf
+ * the response contains a checksum which can be used to verify the data:
  *
- * 1.Take specific values from the attributes (schacPersonalUniqueID, eduPersonTargetedID)
+ * 1.Take specific values from the
+ *   attributes (schacPersonalUniqueID, eduPersonTargetedID)
  * 2.Concatenate into a single string
  * 3.Append a shared salt
  * 4.Hash the string using using MD5
--- a/modules/ding_gatewayf/ding_gatewayf.module
+++ b/modules/ding_gatewayf/ding_gatewayf.module
@@ -252,6 +252,7 @@ function ding_gatewayf_callback() {
  *   The generated salt to return.
  *
  * @return string
+ *   The salt.
  */
 function ding_gatewayf_key_from_attributes(array $params, $ret = '') {
   foreach ($params as $key => $value) {
@@ -285,9 +286,10 @@ function ding_gatewayf_ding_provider_user() {
  * Do a check on the checksum given from gatewayf.
  *
  * @param array $attributes
- *   The attributes from gatewayf
+ *   The attributes from gatewayf.
  *
  * @return bool
+ *   Safe or not.
  */
 function ding_gatewayf_security_check(array $attributes) {
   // Check attributes.
--- a/modules/ding_gatewayf/ding_gatewayf.module
+++ b/modules/ding_gatewayf/ding_gatewayf.module
@@ -217,7 +217,6 @@ function ding_gatewayf_callback() {
   $attributes = $_POST;
   ding_gatewayf_security_check($attributes);
 
-
   // Check if the login was successful and the users information was returned.
   if (!empty($attributes['schacPersonalUniqueID'])) {
     $config = _ding_gatewayf_load_configuration();
@@ -249,11 +248,10 @@ function ding_gatewayf_callback() {
 /**
  * Do a check on the checksum given from gatewayf.
  *
- * @param $attributes
- * @return null
+ * @param array $attributes
  */
 function ding_gatewayf_security_check($attributes) {
-  // Check attributes.
+  //Check attributes.
   if (empty($attributes['schacPersonalUniqueID']) || empty($attributes['eduPersonTargetedID'])) {
     // We expect alway to get these two attributes.
     watchdog('ding_gatewayf', 'Neccesary attributes missing', array(), WATCHDOG_ERROR);
@@ -263,7 +261,7 @@ function ding_gatewayf_security_check($attributes) {
 
   $gateway_checksum = isset($attributes['hash'][0]) ? $attributes['hash'][0] : NULL;
   if (!isset($gateway_checksum)) {
-    // This is an error from gatewayf - checksum should always be in attributes.
+    //This is an error from gatewayf - checksum should always be in attributes.
     watchdog('ding_gatewayf', 'No checksum from gatewayf', array(), WATCHDOG_ERROR);
     drupal_set_message('Sorry login failed - contact site administrator', 'warning');
     ding_gatewayf_user_logout();
@@ -271,10 +269,10 @@ function ding_gatewayf_security_check($attributes) {
 
   $fixed_params = array(
     $attributes['schacPersonalUniqueID'][0],
-    $attributes['eduPersonTargetedID'][0]
+    $attributes['eduPersonTargetedID'][0],
   );
   $key = ding_gatewayf_key_from_attributes($fixed_params);
-  // Get salt from settings $conf.
+  //Get salt from settings $conf.
   $salt = variable_get('ding_gatewayf_salt', NULL);
   if (empty($salt)) {
     //This is a configuration error.
@@ -284,7 +282,7 @@ function ding_gatewayf_security_check($attributes) {
   }
   $ding_checksum = md5($key . $salt);
   if ($ding_checksum !== $gateway_checksum) {
-    // Somebody has changed something - abort.
+    //Somebody has changed something - abort.
     watchdog('ding_gatewayf', 'checksum from gatewayf did not match', array(), WATCHDOG_ERROR);
     ding_gatewayf_user_logout();
   }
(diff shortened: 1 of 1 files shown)


GitCop:

 * Commit: bb2d2ca6e4df9190511f0c7ffebc7b247a72ceab

---

GitCop:

 * Commit: bb2d2ca6e4df9190511f0c7ffebc7b247a72ceab

---

modules/ding_gatewayf/ding_gatewayf.module
```diff
@@ -241,6 +246,70 @@ function ding_gatewayf_callback() {
...
+    $attributes['eduPersonTargetedID'][0],
+  );
+  $key = ding_gatewayf_key_from_attributes($fixed_params);
+  //Get salt from settings $conf.
+  $salt = variable_get('ding_gatewayf_salt', NULL);
+  if (empty($salt)) {
+    //This is a configuration error.
+    watchdog('ding_gatewayf', 'No salt has been defined - wayf login is disabled for security', array(), WATCHDOG_ERROR);
+    drupal_set_message('No salt has been defined - wayf login is disabled for security');
+    ding_gatewayf_user_logout();
```
kasperg: Why do we want to log out the user here?

I would think that the assumption is that if the security check fails then the user has not actually been logged in through gatewayf anyway so there is no reason to log out either.

modules/ding_gatewayf/ding_gatewayf.module
```diff
@@ -241,6 +246,70 @@ function ding_gatewayf_callback() {
 }
 
 /**
+ * Do a check on the checksum given from gatewayf.
```
kasperg: Is there any official documentation we can link to about this security check? I mean: How do we know that this implementation is correct.

modules/ding_gatewayf/ding_gatewayf.module
```diff
@@ -241,6 +246,70 @@ function ding_gatewayf_callback() {
...
+  $ding_checksum = md5($key . $salt);
+  if ($ding_checksum !== $gateway_checksum) {
+    //Somebody has changed something - abort.
+    watchdog('ding_gatewayf', 'checksum from gatewayf did not match', array(), WATCHDOG_ERROR);
+    ding_gatewayf_user_logout();
+  }
+}
+
+/**
+ * Make a cachekey based on request parameters.
```
kasperg: As I see it this is not a cache key?

modules/ding_gatewayf/ding_gatewayf.module
```diff
@@ -241,6 +246,70 @@ function ding_gatewayf_callback() {
...
+ * @param array $params
+ * @param string $ret
+ *
+ * @return string
+ */
+function ding_gatewayf_key_from_attributes($params, $ret = '') {
+  foreach ($params as $key => $value) {
+    if (is_array($value)) {
+      //Recursive.
+      $ret = ding_gatewayf_key_from_attributes($value, $ret);
```
kasperg: I do not think it is a good idea for a recursive function to pass an inner part `$value` and a current state `$ret`. It reduces complexity if it works using `$value` only.

modules/fbs/includes/fbs.user.inc
```diff
@@ -62,6 +62,7 @@ function fbs_user_authenticate($name, $pincode) {
     // We have to save the library card number/CPR number, as the pincode
     // change call uses it.
     $result['creds']['library_card_number'] = $name;
+    $result['creds']['pass'] = $pincode;
```
pjohans: i will take it out - its from a previous issue

modules/ding_gatewayf/ding_gatewayf.module
pjohans: we updated the gateway to deliver a checksum and basically duplicated the methods to get a match, so we know (or at least hope) the implementation is correct because we did it ourselves

modules/ding_gatewayf/ding_gatewayf.module
```diff
@@ -241,6 +246,70 @@ function ding_gatewayf_callback() {
...
+ * @param array $attributes
+ */
+function ding_gatewayf_security_check($attributes) {
+  //Check attributes.
+  if (empty($attributes['schacPersonalUniqueID']) || empty($attributes['eduPersonTargetedID'])) {
+    // We expect alway to get these two attributes.
+    watchdog('ding_gatewayf', 'Neccesary attributes missing', array(), WATCHDOG_ERROR);
+    drupal_set_message('Sorry login failed - contact site administrator', 'warning');
+    ding_gatewayf_user_logout();
+  }
```
pjohans: you're right - there is a double check on one of the attributes (schacPersonalUniqueID). I will fix it and move the call to securitycheck

modules/ding_gatewayf/ding_gatewayf.module
```diff
@@ -241,6 +246,70 @@ function ding_gatewayf_callback() {
...
+    $attributes['schacPersonalUniqueID'][0],
+    $attributes['eduPersonTargetedID'][0],
+  );
+  $key = ding_gatewayf_key_from_attributes($fixed_params);
+  //Get salt from settings $conf.
+  $salt = variable_get('ding_gatewayf_salt', NULL);
+  if (empty($salt)) {
+    //This is a configuration error.
+    watchdog('ding_gatewayf', 'No salt has been defined - wayf login is disabled for security', array(), WATCHDOG_ERROR);
+    drupal_set_message('No salt has been defined - wayf login is disabled for security');
```
pjohans: right again

modules/ding_gatewayf/ding_gatewayf.module
pjohans: well at this point the user has been verified by the idp (nemlogin or wayf), and we want the user to be logged out of the idp again. The ding_gatewayf_logout method (almost) does the right call to gatewayf

modules/ding_gatewayf/ding_gatewayf.module
```diff
@@ -241,6 +246,70 @@ function ding_gatewayf_callback() {
...
+    ding_gatewayf_user_logout();
+  }
+  $ding_checksum = md5($key . $salt);
+  if ($ding_checksum !== $gateway_checksum) {
+    //Somebody has changed something - abort.
+    watchdog('ding_gatewayf', 'checksum from gatewayf did not match', array(), WATCHDOG_ERROR);
+    ding_gatewayf_user_logout();
```
pjohans: if the checksum doesn't match something has changed along the way - it could be man in the middle or playback. If that is the case we simply want to abort

modules/ding_gatewayf/ding_gatewayf.module
pjohans: yeah, right again

GitCop:

 * Commit: bb2d2ca6e4df9190511f0c7ffebc7b247a72ceab

---

GitCop:

 * Commit: bb2d2ca6e4df9190511f0c7ffebc7b247a72ceab

---

modules/ding_gatewayf/ding_gatewayf.module
```diff
@@ -255,12 +283,66 @@ function ding_gatewayf_ding_provider_user() {
...
+ *   Safe or not.
+ */
+function ding_gatewayf_security_check(array $attributes) {
+  // Check attributes.
+  if (empty($attributes['schacPersonalUniqueID']) || empty($attributes['eduPersonTargetedID'])) {
+    // We expect alway to get these two attributes.
+    watchdog('ding_gatewayf', 'Neccesary attributes missing', array(), WATCHDOG_ERROR);
+    drupal_set_message('Sorry login failed - contact site administrator', 'warning');
+    return FALSE;
+  }
```
kasperg: I would still think this should be removed. The current implementation will set an error message and return `FALSE`. The calling code checks for a non-`TRUE` response and set the exact same error message.

modules/ding_gatewayf/ding_gatewayf.module
kasperg: In this case you are right. The user has been logged in but the Ding site has not been configured correctly. You are right to do a logout then.

modules/ding_gatewayf/ding_gatewayf.module
kasperg: I cannot see this check has been removed.

modules/ding_gatewayf/ding_gatewayf.module
kasperg: I cannot see that a check has been moved to a `hook_requirements()`?

modules/ding_gatewayf/ding_gatewayf.module
pjohans: done

modules/ding_gatewayf/ding_gatewayf.module
```diff
@@ -241,6 +244,31 @@ function ding_gatewayf_callback() {
...
+ *   The values to extract from.
+ * @param string $ret
+ *   The generated salt to return.
+ *
+ * @return string
+ *   The salt.
+ */
+function ding_gatewayf_key_from_attributes(array $params, $ret = '') {
+  foreach ($params as $key => $value) {
+    /** @var array $value */
```
kasperg: Why is this needed? Given the subsequent if statement it should be clear that `$value` is not always an array.

modules/ding_gatewayf/ding_gatewayf.module
```diff
@@ -241,6 +244,31 @@ function ding_gatewayf_callback() {
 }
 
 /**
+ * Make a salt based on request parameters.
```
kasperg: This docblock is somewhat misleading. This function does not generate a [salt](https://en.wikipedia.org/wiki/Salt_(cryptography)). The salt is the value set as a variable in `settings.php`.

modules/ding_gatewayf/ding_gatewayf.module
kasperg: But the Ding2 community needs to maintain this going forward so we need to get a shared understanding of what the contract of the implementation is. If there is no documentation to reference then please add a summary of what is going on.

Perhaps something in the lines of:

To protect against malicious attempts to login through gatewayf the response contains a checksum which can be used to verify the data: 

1. Take all recieved values except the checksum
2. Concatenate into a single string
3. Append a shared salt
4. Hash the string using using MD5
5. Compare the hash with the checksum

kasperg:
Merged. Thanks @pjohans!

PR #529, opened 2017-02-03
Issue2164 wayf

pjohans:


pjohans:
Implements security check in ding_wayf also .. mostly copy/pastes from ding_gatewayf

modules/ding_wayf/ding_wayf.module
```diff
@@ -428,6 +507,14 @@ function ding_wayf_get_credentials() {
   $saml = ding_wayf_get_saml_connection();
   $saml_attributes = $saml->getAttributes();
 
+  // Check security - we need to do it here, it is the only place where we
+  // have (unspoiled) attributes directly from gatewayf.
+  $status = ding_wayf_security_check($saml_attributes);
+  /** @var bool $status */
```
kasperg: Why is this needed? `ding_wayf_security_check()` already declares that it returns a boolean.

modules/ding_wayf/ding_wayf.module
```diff
@@ -124,6 +124,81 @@ function ding_wayf_init() {
...
+ *   The values to extract from.
+ * @param string $ret
+ *   The generated salt to return.
+ *
+ * @return string
+ *   The salt.
+ */
+function ding_wayf_key_from_attributes(array $params, $ret = '') {
+  foreach ($params as $key => $value) {
+    /** @var array $value */
```
kasperg: Why is this needed? Given the subsequent if statement it should be clear that `$value` is not always an array.

modules/ding_wayf/ding_wayf.module
```diff
@@ -124,6 +124,81 @@ function ding_wayf_init() {
...
+    // Somebody has changed something - abort.
+    watchdog('ding_gatewayf', 'checksum from gatewayf did not match', array(), WATCHDOG_ERROR);
+    return FALSE;
+  }
+
+  return TRUE;
+}
+
+/**
+ * Make a salt based on request parameters.
```
kasperg: This docblock is somewhat misleading. This function does not generate a [salt](https://en.wikipedia.org/wiki/Salt_(cryptography)). The salt is the value set as a variable in `settings.php`.

modules/ding_wayf/ding_wayf.module
```diff
@@ -124,6 +124,81 @@ function ding_wayf_init() {
...
+  }
+
+  $fixed_params = array(
+    $attributes['schacPersonalUniqueID'][0],
+    $attributes['eduPersonTargetedID'][0],
+  );
+
+  $key = ding_wayf_key_from_attributes($fixed_params);
+  // Get salt from settings $conf.
+  $salt = variable_get('ding_gatewayf_salt', NULL);
```
kasperg: It seems odd that we get a variable prefixed with the name of a module (ding_gatewayf) which we probably are not using.

modules/ding_wayf/ding_wayf.module
```diff
@@ -124,6 +124,81 @@ function ding_wayf_init() {
...
+ *
+ * @return bool
+ *   Safe or not.
+ */
+function ding_wayf_security_check(array $attributes) {
+  // Check attributes.
+  if (empty($attributes['schacPersonalUniqueID']) || empty($attributes['eduPersonTargetedID'])) {
+    // We expect alway to get these two attributes.
+    watchdog('ding_gatewayf', 'Neccesary attributes missing', array(), WATCHDOG_ERROR);
+    drupal_set_message('Sorry login failed - contact site administrator', 'warning');
```
kasperg: Why set a message here? We return a boolean and let the callee determine how to handle this.

As I see it a message will be printed on L216.

modules/ding_wayf/ding_wayf.module
pjohans: It is the same variable used in ding_gateway and ding_wayf
that is what Christian and i agreed on - naming the variable

modules/ding_wayf/ding_wayf.module
pjohans: scrutinizer wants it

modules/ding_wayf/ding_wayf.module
pjohans: scrutinizer wants it

DDBdrift:
https://platform.dandigbib.org/issues/2164 | Resolved (tag version)

PR #533, opened 2017-02-10
2170: introduce a type-filter on the news feed

danquah:
This PR introduces an optional filter on news-type to the library news filter, and fixes a bug where the feed was not filtered by library.

Before the PR eg
https://bibliotek.kk.dk/node/24/news/feed would show 20 news-items across all libraries
after the PR
https://bibliotek.kk.dk/node/24/news/feed will now only show news-items for the library node 24

Futhermore the PR makes it possible to filter further down to a specific news type, eg. 
https://bibliotek.kk.dk/node/24/news/feed/11562

Would only show news-items categorized by the news-category term with id 11562 (Servicemeddelser - lokal)


DDBdrift:
https://platform.dandigbib.org/issues/2170 | Closed

PR #534, opened 2017-02-16
2178: Fix search backend radios for small screens

holt83:
http://platform.dandigbib.org/issues/2178

Since the form is added twice, '--2' is appended to the HTML-id and the selector doesn't work. Using panel-pane class in selector targets both forms.


kasperg:
This has been solved elsewhere.

DDBdrift:
https://platform.dandigbib.org/issues/2178 | Resolved

PR #535, opened 2017-02-23
2185 by Inlead: Enable Campaign module per default.

inlead:
http://platform.dandigbib.org/issues/2185

### DIFF
--- a/ding2.install
+++ b/ding2.install
@@ -440,3 +440,10 @@ function ding2_update_7030() {
 function ding2_update_7031() {
   ding2_translation_update();
 }
+
+/**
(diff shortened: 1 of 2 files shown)


ding2.install
```diff
@@ -440,3 +440,10 @@ function ding2_update_7030() {
 function ding2_update_7031() {
   ding2_translation_update();
 }
+
+/**
+ * Enable Campaign module.
+ */
+function ding2_update_7032) {
+  module_enable(array('ding_campaign'));
```
kasperg: There is a typo in the update hook definition.

PR #536, opened 2017-02-23
2184: Change admin_menu fix to use permissions

holt83:
http://platform.dandigbib.org/issues/2184#note-18

### DIFF
--- a/modules/ding_permissions/ding_permissions.module
+++ b/modules/ding_permissions/ding_permissions.module
@@ -993,19 +993,22 @@ function ding_permissions_secure_permissions($role) {
     }
   }
 
+  // Ensure that administrative users only have access to one navigation system.
+  foreach ($permissions as &$role_permissions) {
+    if (in_array('access administration pages', $role_permissions)) {
+      // Roles with this permission should have access to Admin menu.
+      $role_permissions = array_unique(array_merge($role_permissions, array('access administration menu')));
+      $role_permissions = array_diff($role_permissions, array('access toolbar'));
+    }
+    else {
+      // If the role doesn't have 'access administration pages', Admin menu is
+      // useless so we can safely remove any access to it. If they role has
+      // access to Toolbar it will fallback to that.
+      $role_permissions = array_diff($role_permissions, array('access administration menu'));
+    }
+  }
+
   if (isset($permissions[$role])) {
     return $permissions[$role];
-  }
-}
-
-/**
- * Implements hook_preprocess_HOOK().
- */
-function ding_permissions_preprocess_html(&$variables) {
-  if (user_access('access administration menu') && !user_access('access administration pages')) {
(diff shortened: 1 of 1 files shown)


modules/ding_permissions/ding_permissions.module
```diff
@@ -1000,12 +1001,10 @@ function ding_permissions_secure_permissions($role) {
  * Implements hook_preprocess_HOOK().
  */
 function ding_permissions_preprocess_html(&$variables) {
-  global $user;
```
kasperg: It makes sense to use permissions instead of roles but why is this check for `access administration menu AND NOT access administration pages` needed?

I would think the proper way to do this would be to avoid this preprocessing altogether and rely on the permissions already supported by the toolbar and administration modules: If the user has `access administration menu` then show the admin menu. If the user has `access toolbar` then show the toolbar.

holt83: The resoning was that if the user has the 'access administration menu' but not the 'access administration pages' the display of admin_menu will have no top-level categories:https://www.drupal.org/node/1337864#comment-5493320

But your right. The proper way would be to update permissions such that all users **without** 'access administration pages' will have the 'access toolbar' and all users **with** the 'access administration pages' should have the 'access administration menu' and **NOT** 'access toolbar' (or else they will see two toolbars).

We could use the preprocess to ensure that no user will see toolbars, but if we're carefull with permissions it shouldn't be an issue.

Let me know if I should update the PR with this solution.

kasperg: @holt83: I see your point regarding careful use of permissions.

Could we avoid the preprocess altogether and instead implement this logic in `ding_permissions_secure_permissions()` programmatically as is done with `promote to front page option`?

I am thinking something like:

```php
// Ensure that administrative users only have access to one navigation system: Admin menu and not toolbar.
foreach ($permissions as $role => &$role_permissions) {
  if (in_array('access administration pages', $role_permissions)) {
    $role_permissions = array_unique(array_merge($role_permissions, array('access administration menu')));
    $role_permissions = array_diff($role_permissions, ['access toolbar']);
  }
}
```

holt83: Good idea. Didn't think of putting the logic in secure permissions hook.

So if the role has `access administration pages` we make sure it has admin_menu and not toolbar.

We could also make sure that a role without `access administration pages` doesn't have permissions to both. What do you think?

holt83: Updated the PR with the proposed solution plus an additional safety measure.

Tested and works on all roles from administrators to guest_blogger. User 1 still sees two bars though, but I guess that doesn't matter much?

kasperg: Great. I do not think it matters for user 1.

PR #537, opened 2017-02-28
2191 by Inlead: Adjust ding_campaign permissions.

inlead:
http://platform.dandigbib.org/issues/2191


kasperg:
For the record: This PR removes editor permissions to edit/delete *any* campaign and adds the permission for local editors to create/edit/delete own campaigns.

PR #550, opened 2017-03-13
1721: Move filter by holding logic to request class

holt83:
http://platform.dandigbib.org/issues/1721


JrgenGNielsen:
reviewed and approved

PR #552, opened 2017-03-17
1773: Improve holdings display for FBS

holt83:
http://platform.dandigbib.org/issues/1773


JrgenGNielsen:
reviewed and approved

PR #557, opened 2017-03-24
2124: Removed serendipity key overview

BirkAndMe:
http://platform.dandigbib.org/issues/2124


JrgenGNielsen:
rewieved and approved

PR #559, opened 2017-03-24
2120: User lists UX

BirkAndMe:
http://platform.dandigbib.org/issues/2120

### DIFF
--- a/modules/p2/ding_list/include/form.inc
+++ b/modules/p2/ding_list/include/form.inc
@@ -347,7 +347,11 @@ function ding_list_create_list_form_submit($form, &$form_state) {
       return;
     }
 
-    drupal_set_message(theme('ding_list_create_info', array('list' => $list)));
+    $out = array(
+      '#theme' => 'ding_list_create_info',
+      '#list' => $list,
+    );
+    drupal_set_message(drupal_render($out));
   }
   else {
     $list = ding_list_get_list($form_state['existing']);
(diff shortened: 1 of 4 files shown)


modules/p2/ding_list/include/form.inc
```diff
@@ -347,13 +347,7 @@ function ding_list_create_list_form_submit($form, &$form_state) {
...
     }
 
-    drupal_set_message(t(
-        "You've create the list !list.<br/>Read more about how you use the lists !help",
-        array(
-          '!list' => l($list->title, 'list/' . $list->id),
-          '!help' => l(t('Here'), 'help/lists'),
-        )
-      ));
+    drupal_set_message(theme('ding_list_create_info', array('list' => $list)));
```
JrgenGNielsen: It is preferable to replace direct calls to the theme() function with calls to drupal_render() by passing a render array with a #theme key to drupal_render(), which in turn calls theme().

modules/p2/ding_list/theme/theme.inc
```diff
@@ -12,6 +12,31 @@ function theme_ding_list_message($variables) {
...
+  ));
+  $items[] = t('Use the "More" button on materials, to add them to the list.');
+  $items[] = t('You can search after materials !search, that you can add to the list.', array(
+    '!search' => l(t('Here'), 'search'),
+  ));
+  $items[] = t('Read more about how you use the lists !help', array(
+    '!help' => l(t('Here'), 'help/lists'),
+  ));
+
+  $out .= theme('item_list', array('items' => $items));
```
JrgenGNielsen: As above. See also https://api.drupal.org/api/drupal/includes!theme.inc/function/theme/7.x

JrgenGNielsen:
rewieved and approved

PR #560, opened 2017-03-24
2119: Remove loan history sharing

BirkAndMe:
http://platform.dandigbib.org/issues/2119


JrgenGNielsen:
rewieved and approved

PR #561, opened 2017-03-24
2091: Public lists sort order

BirkAndMe:
http://platform.dandigbib.org/issues/2091


JrgenGNielsen:
rewieved and approved

PR #562, opened 2017-03-24
1946: Anonymous more button

BirkAndMe:
http://platform.dandigbib.org/issues/1946

### DIFF
--- a/modules/p2/ding_list/ding_list.module
+++ b/modules/p2/ding_list/ding_list.module
@@ -190,7 +190,69 @@ function ding_list_menu() {
     'page arguments' => array(2),
   ));
 
+  $items['ding-list/login-create'] = array(
+    'title' => 'Login create',
+    'page callback' => array('ding_list_login_create'),
+    'delivery callback' => 'ajax_deliver',
+    'access callback' => TRUE,
+  );
+
+  $items['ding-list/login-attach/%/%'] = array(
+    'title' => 'Login create',
+    'page callback' => array('ding_list_login_attach'),
+    'page arguments' => array(2, 3),
+    'delivery callback' => 'ajax_deliver',
+    'access callback' => TRUE,
+  );
+
   return $items;
+}
+
+/**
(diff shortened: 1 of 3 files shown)


modules/p2/ding_list/ding_list.module
```diff
@@ -190,10 +190,72 @@ function ding_list_menu() {
...
+    'page arguments' => array(2, 3),
+    'delivery callback' => 'ajax_deliver',
+    'access callback' => TRUE,
+  );
+
   return $items;
 }
 
 /**
+ * Handle list attachment for anonuymous users.
```
JrgenGNielsen: typo

modules/p2/ding_list/ding_list.module
```diff
@@ -190,10 +190,72 @@ function ding_list_menu() {
...
+ */
+function ding_list_login_attach($type, $value) {
+  $commands = array();
+
+  if (!user_is_logged_in()) {
+    $commands[] = ajax_command_ding_user_authenticate('');
+  }
```
JrgenGNielsen: you could do
return array('#type' => 'ajax', '#commands' => $commands);
here, and do without the 'else' structure.
It's a matter of taste, though... personally, I like the "early return style of coding". See f.ex. http://blog.timoxley.com/post/47041269194/avoid-else-return-early

modules/p2/ding_list/ding_list.module
```diff
@@ -190,10 +190,72 @@ function ding_list_menu() {
...
+
+/**
+ * Handle list creation for anonymous users.
+ */
+function ding_list_login_create() {
+  $commands = array();
+
+  if (!user_is_logged_in()) {
+    $commands[] = ajax_command_ding_user_authenticate('');
+  }
```
JrgenGNielsen: as line 219

JrgenGNielsen:
I have a couple of comments....

JrgenGNielsen:
rewieved and approved

PR #563, opened 2017-03-24
1939: List share message

BirkAndMe:
http://platform.dandigbib.org/issues/1939


JrgenGNielsen:
rewieved and approved

PR #568, opened 2017-04-05
2267: IE specific css

BirkAndMe:
http://platform.dandigbib.org/issues/2267


JrgenGNielsen:
reviewed and approved

PR #570, opened 2017-04-07
2163: Validate CQL in serendipity

BirkAndMe:
http://platform.dandigbib.org/issues/2163


JrgenGNielsen:
reviewed and approved

PR #572, opened 2017-04-07
2127: User consent response

BirkAndMe:
http://platform.dandigbib.org/issues/2127


kasperg:

PR #573, opened 2017-04-07
2145: revive interest period

xendk:
This revives the interest period field on user profiles, and changes it to store it locally.


cableman:
Reviewed and approved

PR #574, opened 2017-04-10
2258: Ting caching

xendk:
Precache materials on the public_lists page, and attempt to fix ting caching.


JrgenGNielsen:
reviewed and approved

PR #576, opened 2017-04-11
1618: branches blacklist

xendk:
Fixes up #480

### DIFF
--- a/modules/fbs/fbs.module
+++ b/modules/fbs/fbs.module
@@ -259,39 +259,19 @@ function fbs_settings_form() {
     '#description' => t('Allow users to delete their reservations as well as ready for pickup ones.'),
   );
 
-  // Update list of branches.
-  $form['update'] = array(
-    '#type' => 'fieldset',
-    '#title' => t('Update from library system'),
-  );
-
-  $form['update']['fbs_branches_last_updated'] = array(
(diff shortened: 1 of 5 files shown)


xendk:
Scrutinizer har drukket af natpotten. Den indentering den kræver er hul i hovedet.

modules/fbs/fbs.module
```diff
@@ -575,3 +620,33 @@ function _fbs_wayf_login($name, $value = NULL) {
...
+  }
+}
+
+/**
+ * Quote string for use in CQL.
+ *
+ * Adds " around the string and escapes inline ".
+ */
+function _fbs_cql_quote_string($name) {
+  return '"' . preg_replace('/"/', '//"', $name) . '"';
```
holt83: When testing it seems like only \\" is working for escaping. So shouldn't this be something like:
`preg_replace('/"/', '\\\\"', $name)`

xendk: No, two is enough, we need to escape the backslash to make it part of the string. If it was in the regexp, four is needed to escape it in the string and escape it in the regexp, but as this is the replacement string, we don't need that.

But it should be backslashes, not slashes, yes.

holt83: I see :)

JrgenGNielsen:
reviewed and approved

PR #578, opened 2017-04-11
1625: Add FBS to installer

xendk:
Add FBS to installer.


cableman:
Reviewed and approved

PR #579, opened 2017-04-11
1458: Add custom message when not reservable

xendk:
Rather than displaying "0 reservations" on non-reservables, display "Cannot be reserved".

Also removes the count column (redmine 31).

### DIFF
--- a/modules/ding_availability/ding_availability.module
+++ b/modules/ding_availability/ding_availability.module
@@ -192,7 +192,7 @@ function ding_availability_holdings($provider_ids) {
       }
       // If yes, add a column.
       if ($item['ordered_count']) {
-        $header['acquisition'] = t('In acquisition');
+        $header['ordered_count'] = t('In acquisition');
       }
 
       $rows = array();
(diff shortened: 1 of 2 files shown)


JrgenGNielsen:
Good job on the render array - however, calling the theme() function directly is strongly discouraged. (See: https://api.drupal.org/api/drupal/includes%21theme.inc/function/theme/7.x).

modules/ding_availability/ding_availability.module
```diff
@@ -208,23 +211,48 @@ function ding_availability_holdings($provider_ids) {
...
+      }
+
+      if ($item['reservable']) {
+        $variables['reserved_text'] = format_plural($reserved_count,
+                                      'There is 1 user in queue to loan the material.',
+                                      'There are @count users in queue to loan the material.');
+      }
+
+      // Theme the output.
+      $item['html'] = theme('ding_holdings', $variables);
```
JrgenGNielsen: why call theme() here? you can return a render array, and let the template drupal_render() it

xendk: Normally, yes, but this function is also called by `ding_availability_holdings_js()` which simply passes the return value to the browser, so that would require some refactoring.

JrgenGNielsen:
reviewed and approved

PR #580, opened 2017-04-12
813: Validate phone number

xendk:


cableman:
Reviewed and awaiting release

PR #587, opened 2017-04-21
2305: Patch Features against SA-CONTRIB-2016-020

kasperg:
This closes a security hole. The change is also included in release
2.9 but updating Features may also introduce other changes and thus
introduce risk.


cableman:
Reviewed and awaiting release

PR #588, opened 2017-04-21
1460: reserve button on periodicals

xendk:
Hide reservation button on periodicals with issues.


JrgenGNielsen:
reviewed and approved

PR #589, opened 2017-04-21
2306: Update Field Group to 1.5

kasperg:
This fixes a security vulnerability in 1.4.

The new version does not introduce new permissions.

The new version introduces a few additional configuration options.
One of these changes has to do with handling of additional classes. 
Consequently we update all Feature exports of field groups
accordingly. Otherwise we use the defaults.


cableman:

Scrutinizer is behaving very strangely...

PR #590, opened 2017-04-21
2289: Remove unused argument from ding_user_get_creds()

xendk:


cableman:
Reviewed and apporved

PR #594, opened 2017-04-24
2311: Update Entity API to 1.8

kasperg:
Release 1.5 contains [a security vulnerability](https://www.drupal.org/node/2437905).

We want to get the most recent version as Entity API is a module 
which is used by other modules e.g. workflow and we do not want to
restrain out possibilities for using and updating such modules.

There are no changes to configuration, permissions or texts between 
the two versions.

### DIFF
--- a/ding2.make
+++ b/ding2.make
@@ -64,7 +64,7 @@ projects[email][subdir] = "contrib"
 projects[email][version] = "1.3"
 
 projects[entity][subdir] = "contrib"
-projects[entity][version] = "1.5"
+projects[entity][version] = "1.8"
 
 projects[entitycache][subdir] = "contrib"
 projects[entitycache][version] = "1.2"


attritionorg:
@kasperg Do you have a reference to the vulnerability?

kasperg:
@attritionorg: I have updated the PR description with a link.

In general Drupal maintains [a list of security advisories](https://www.drupal.org/security) where this is publicly available. It is also mentioned in [the release notes for each project](https://www.drupal.org/project/entity/releases). What is your background for asking?

cableman:
Reviewed and awaiting release

attritionorg:
@kasperg I help maintain a vulnerability database. We're big on making sure that new entries aren't added for a case like this; instead we update the base entry for the vulnerability to include products affected due to the inclusion of the library code.

PR #595, opened 2017-04-25
2256: Fixed use libcode in popular

BirkAndMe:
http://platform.dandigbib.org/issues/2256


JrgenGNielsen:
reviewed and approved

PR #596, opened 2017-04-25
2317: Update metagtag to 1.21

kasperg:

The new version does not introduce any configuration or permission
changes which require updates.

The new version introduces new strings but they are purely admin
facing.


cableman:
Reviewed and awaiting release

PR #597, opened 2017-04-25
2171: Staff lists

BirkAndMe:
http://platform.dandigbib.org/issues/2171

### DIFF
--- a/modules/p2/ding_list/theme/theme.inc
+++ b/modules/p2/ding_list/theme/theme.inc
@@ -149,6 +149,7 @@ function theme_ding_list_info($variables) {
  * Ding list lists theme.
  */
 function theme_ding_list_lists($variables) {
+  global $user;
   $out = '';
 
   foreach ($variables['groups'] as $group_name => $group) {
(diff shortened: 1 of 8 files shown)


modules/p2/ding_list/theme/theme.inc
```diff
@@ -164,24 +165,28 @@ function theme_ding_list_lists($variables) {
       }
     }
 
+    if (!empty($rows)) {
+      $out .= theme('item_list', array(
```
JrgenGNielsen: Avoid calling theme() directly. It is preferable to replace direct calls to the theme() function with calls to drupal_render() by passing a render array with a #theme key to drupal_render(), which in turn calls theme().
See: https://api.drupal.org/api/drupal/includes!theme.inc/function/theme/7.x

modules/p2/ding_list/theme/theme.inc
```diff
@@ -164,24 +165,28 @@ function theme_ding_list_lists($variables) {
...
+        'title' => $group['title'],
+        'type' => 'ul',
+        'attributes' => array(),
+      ));
+    }
+  }
+
+  // Footer links.
+  if ($variables['uid'] == $user->uid) {
     $out .= theme('item_list', array(
```
JrgenGNielsen: ditto

JrgenGNielsen:
Jeg har en kommentar vedr. brug af theme()

modules/p2/ding_list/theme/theme.inc
```diff
@@ -164,24 +165,33 @@ function theme_ding_list_lists($variables) {
...
+    if (!empty($rows)) {
+      $render = array(
+        '#theme' => 'item_list',
+        '#title' => $group['title'],
+        '#items' => $rows,
+        '#title' => $group['title'],
+        '#type' => 'ul',
+        '#attributes' => array(),
+      );
+      $out .= drupal_render($render);
```
cableman: Calling drupal_render is basically the same as theme().

You theme functions should return a render array and let the theme layer do the rendering. So the front-end has a change to alter in style and content.

modules/p2/ding_list/theme/theme.inc
```diff
@@ -164,24 +165,33 @@ function theme_ding_list_lists($variables) {
...
+      '#theme' => 'item_list',
+      '#items' => array(
+        l(t("Mark all notifications as read"), "ding_message/purge_user_messages", array('attributes' => array('class' => array('button')))),
+        l(t("Create new list"), "list/create", array('attributes' => array('class' => array('button')))),
+      ),
+      '#title' => "",
+      '#type' => 'ul',
+      '#attributes' => array('class' => array('bottom-links')),
+    );
+    $out .= drupal_render($render);
```
cableman: Calling drupal_render is basically the same as theme().

You theme functions should return a render array and let the theme layer do the rendering. So the front-end has a change to alter in style and content.

modules/p2/ding_list/theme/theme.inc
BirkAndMe: It's called from inside a theme_HOOK() function, we could move the creation of the item_list to a preprocessor, but we'd still need to call drupal_render inside the theme_HOOK().

modules/p2/ding_list/theme/theme.inc
BirkAndMe: It's called from inside a theme_HOOK() function, we could move the creation of the item_list to a preprocessor, but we'd still need to call drupal_render inside the theme_HOOK().

PR #598, opened 2017-04-26
2320: Update Views to 3.16

kasperg:
This fixes [multiple security vulnerabilities up until 3.15](https://www.drupal.org/project/views/releases?api_version%5B%5D=103).

We update to 3.16 as 3.15 has some changes which hav subsequently been
reverted to earlier behaviour. We do not want to risk updating the
system due to these changes only to have to change them back in a
subsequent update.

The new version may introduce new configuration options.
We use the defaults here.

The new version does not introduce any new permissions.


cableman:
Reviewed and awaiting release

PR #599, opened 2017-04-26
2321: Update Search API to 1.18

kasperg:
This fixes [a security vulnerability](https://www.drupal.org/node/2710063).

Updating also means we can drop a patch which has been merged into
the new release.

There are no new permissions or configuration changes.

The new version does have new texts in the administrative interface.


cableman:
Reviewed and awaiting release

PR #601, opened 2017-04-26
2324: Update Organic Groups to 2.9

kasperg:

The update also means that we can drop two patches which have been
merged into the newer version.

The new version may introduce new configuration options.
We use the defaults here.

The new version does not introduce any new permissions.


cableman:
Reviewed and awaiting release

PR #602, opened 2017-04-26
2325: Update Organic Groups Menu to 3.0

kasperg:

Updating also means we can drop a patch which has been merged into
the newer version of the module.

There are no new permissions, configuration options or texts in the
newer version.


cableman:
Reviewed and awaiting release

PR #604, opened 2017-04-27
2328: Update Media: YouTube module to 3.0

kasperg:
The current release lies between 2.0-rc4 and 2.0-rc5. Release 3.0 is
out and consequently the current version will no longer receive
security updates.

This also means we can drop a patch since it has been merged in the
newer version of the module.

There are no new tests, permissions or configuration changes in the
newer version.


cableman:
Reviewed and awaiting release

PR #605, opened 2017-04-27
2329: Show video players on site - not as links

kasperg:
oEmbed module takes precedence over YouTube and Vimeo media modules.
It supports YouTube and Vimeo and will thus handle urls from these
sites. Our Media videos are not configured to show videos through
oEmbed and will thus show a link instead.

To fix this we disable oEmbed support from all other sites except
eKurser.

This will also prevent authors from embedding content which we have 
not approved.


JrgenGNielsen:
reviewed and approved

PR #606, opened 2017-05-01
2335: Ensure that stray #weights doesn't mess up ting display

xendk:


cableman:
Reviewed and awaiting release

PR #607, opened 2017-05-02
2330: Update Media module to 2.0

kasperg:
This fixes multiple security vulnerabilities in the Media module.

File Entity needs to be updated as well as thumbnails do not work
with the combination of the new Media module and the old File Entity
module.

We can also drop a patch for Media since the change has been merged
in the newer version of the module.

With the new versions of Media and File Entity elements are now
children of the widget form. Consequently we have to update our logic
for adding the help text.

With the updated Media and File Entity modules the manualcrop module
works. Preview and crop button are shown in the administrative UI
but are aligned vertically - not horizontally.

We need to patch the module to fix that.


cableman:
Reviewed and approved for release

PR #609, opened 2017-05-09
2327 by Inlead: Various permission improvements.

inlead:


modules/ding_permissions/ding_permissions.module
```diff
@@ -1011,3 +1013,14 @@ function ding_permissions_secure_permissions($role) {
     return $permissions[$role];
   }
 }
+
+/**
+ * Implements hook_page_alter().
```
cableman: A comment about why this change in the page is need would be nice.

bozeak: There was a problem with toolbars (core toolbar and admin_menu toolbar) when the user with role "administrator" and uid=1 was logged in, both toolbars was displayed. Through standard permissions system it was not possible to disable core toolbar for administrator role, we were forced to unset this on page build.

cableman: Would it be not be possible to just use one of the toolbars? Simply disabling the one you don't want to use?

guddo: @cableman That's a decision by the business - you can [read more about it in the ticket tracker](http://platform.dandigbib.org/issues/2206).

cableman: Still would like a comment in the code about this, so we also knowns why it's here in the future.

holt83: I don't see it as a big problem that uid=1 has two toolbars. As administrator you can easily just disable one of the modules.

We removed hook_page_alter() in http://platform.dandigbib.org/issues/2184 in favor of handling it in hook_secure_permission. I see no reason to reintroduce this hook just for uid=1. If permissions is set correctly up for all the roles, there shouldn't really be any problem.

cableman:

PR #611, opened 2017-05-17
2369: Fix IE arrows

KristianB14:
http://platform.dandigbib.org/issues/2369


JrgenGNielsen:
reviewed and approved

PR #612, opened 2017-05-17
2370: Openlist prefix and default url

BirkAndMe:
http://platform.dandigbib.org/issues/2370


JrgenGNielsen:
reviewed and approved

PR #613, opened 2017-05-18
#2014 by Inlead: Adjust permissions for the Staff-overview.

guddo:
http://platform.dandigbib.org/issues/2014


kasperg:
Thanks @guddo. Reviewed and approved.

PR #615, opened 2017-05-24
Core: PHP 7 patches and fixes

cableman:


kasperg:
Should we increase the PHP version for the project as well in Circle CI and/or run PHP7 compatibility static code analysis in Scrutinizer? If so perhaps it should analyse the entire codebase.

cableman:
Yes, we should also update the test etc. :-)

This is just the first step in the test process and to have a place to apply a PHP7 patch to local  builds for the PL's.

cableman:
It is sadly not possible to get php7cc to return a non-zero exit code :-(

See https://github.com/sstalle/php7cc/issues/102

kasperg:
According to @xendk it should be possible to wrap the call to `php7cc` in some other shell commands to detect errors in the output and return a proper exit code accordingly.

xendk:
Something like:

```sh
php7cc --except=vendor --level=error --extensions=php,inc,module,install $DRUPAL_SITE_PATH | tee /dev/tty | grep -vq 'File: '
```

circle.yml
```diff
@@ -49,6 +53,8 @@ test:
     - cp -R ./* $DRUPAL_SITE_PATH/profiles/ding2/
     # Install the site using the ding2 profile
     - (cd $DRUPAL_SITE_PATH && drush site-install ding2 --db-url=mysql://ubuntu@127.0.0.1/circle_test -y)
+    # Run PHP7 Compatibility Checker
+    - 'php7cc --except=vendor --level=error --extensions=php,inc,module,install $DRUPAL_SITE_PATH | tee /dev/tty | grep -vq ''File: '''
```
kasperg: Adding a comment about why we invoke php7cc as we do would be good. Consider adding a reference to [the related issue](https://github.com/sstalle/php7cc/issues/102).

drupal.make
```diff
@@ -3,10 +3,11 @@ api = 2
 
 ; Core
 projects[drupal][type] = core
-projects[drupal][version] = 7.54
-projects[drupal][patch][] = http://drupal.org/files/issues/menu-get-item-rebuild-1232346-45.patch
-projects[drupal][patch][] = http://drupal.org/files/ssl-socket-transports-1879970-13.patch
-projects[drupal][patch][] = http://drupal.org/files/issues/translate_role_names-2205581-1.patch
+projects[drupal][version] = 7.56
```
kasperg: This is Drupal Core 7.56 - README mentions 7.55. Which is it? 😄

drupal.make
```diff
@@ -3,10 +3,11 @@ api = 2
...
+projects[drupal][patch][] = "http://drupal.org/files/issues/menu-get-item-rebuild-1232346-45.patch"
+projects[drupal][patch][] = "http://drupal.org/files/ssl-socket-transports-1879970-13.patch"
+projects[drupal][patch][] = "http://www.drupal.org/files/issues/1232416-autocomplete-for-drupal7x53.patch"
+projects[drupal][patch][] = "http://drupal.org/files/issues/translate_role_names-2205581-1.patch"
```
kasperg: This does not seem to include all the patches mentioned in the README e.g. https://www.drupal.org/files/issues/DATE_RFC7231-2877243-26.patch seems to be missing.

General comments not related to this PR:
- https would be nice.
- patches should have comments.

modules/ding_ddbasic/ding_ddbasic.module
```diff
@@ -79,7 +79,35 @@ function ding_ddbasic_form_alter(&$form, &$form_state, $form_id) {
...
+          '#options' => array(
+            // The primary and secondary value is handled in the theme_settings
+            // submit callback, in the ddbasic theme.
+            'primary' => t('Primary'),
+            'secondary' => t('Secondary'),
+            DING_DDBASIC_COLOR_WHITE => t('White'),
+          ),
+          '#size' => '1',
+          '#default_value' => $default_value,
+        ) + (isset($form['color']['palette']['text_link_on_dark']) ? $form['color']['palette']['text_link_on_dark'] : array());
```
kasperg: Why are these changes included in this PR? They do not seem to be related to the changes at hand.

README.md
```diff
@@ -9,7 +9,7 @@ of the [TING concept](http://ting.dk).
 
 # Installation
 This README assumes that you have install a configured your server with a
-working Apache/Nginx, APC, Memcached, PHP 5.6 and Varnish 3.x (optional). The
+working Apache/Nginx, APC, Memcached, PHP 7.0 and Varnish 3.x (optional). The
```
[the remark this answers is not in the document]
cableman: It will work with older versions as well (re-written to 5.6+)

README.md
```diff
@@ -58,12 +58,18 @@ web-services that runs OpenSSL v1.0.x or newer works.
   ~$ wget -qO- http://drupal.org/files/ssl-socket-transports-1879970-13.patch | patch -p1
 ```
 
+This [patch](https://www.drupal.org/node/2877243) ensure that core is PHP 7.0 compliant.
+web-services that runs OpenSSL v1.0.x or newer works.
```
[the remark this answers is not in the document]
cableman: Removed

README.md
```diff
@@ -58,12 +58,18 @@ web-services that runs OpenSSL v1.0.x or newer works.
...
+  ~$ wget -qO- https://www.drupal.org/files/issues/DATE_RFC7231-2877243-26.patch | patch -p1
+```
+
 __Optional__,but recommended patch that ensures that Ajax errors only are
 displayed when not in readystate 4. So when the user presses enter to perform a
 search before auto-complete Ajax is call is completed an error will not be
 displayed.
 ```sh
-  ~$ wget -qO- https://drupal.org/files/issues/autocomplete-1232416-17-7x.patch | patch -p1
+  ~$ wget -qO- http://www.drupal.org/files/issues/autocomplete-1232416-205-7x.patch | patch -p1
```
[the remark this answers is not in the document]
cableman: Have not written https as all the other patches in the readme is not https.

circle.yml
cableman: Added

drupal.make
cableman: Updated

drupal.make
cableman: Patch removed as it's part of the 7.56 release of core.

modules/ding_ddbasic/ding_ddbasic.module
cableman: Yes they are.. as php 5.6 ignores these but PHP 7.0 throws an error that stops PHP execution. So the theme settings page is not accessible when using php 7

kasperg:
Failing the build on PHP7 errors does not work currently.

We are using the following command to check:

```bash
php7cc --except=vendor --level=error --extensions=php,inc,module,install $DRUPAL_SITE_PATH | tee /dev/tty | grep -vq 'File: '
```

I suspect that it fails because the output is multiline. As such it contains other lines which does not match the pattern.

I merged this to allow us to start testing but failing the build even if errors occur should still be an issue.

PR #617, opened 2017-05-24
Don't render empty campaigns 

xendk:
Don't render block content, as it'll never be empty

CTools hides the pane when empty($block->content), but an array with
an empty #markup is never empty.

Also, there's no need for drupal_render as ctools will do that for us.


cableman:
Reviewed and awaiting release

PR #620, opened 2017-05-30
2439: Fix dobbelt encoding in tabroll title

holt83:
http://platform.dandigbib.org/issues/2439

This also corrects hook_theme() to use the correct 'base hook' instead of 'original hook'. Se the following for more info:

https://api.drupal.org/api/views/help!api-default-views.html/7.x-3.x
https://drupal.stackexchange.com/questions/25846/how-to-tell-drupal-to-look-for-templates-in-module-directory
https://www.drupal.org/node/2106635

The problem with not using 'base hook' is that we have to specify processors manually. As we can see if we take a closer look at https://api.drupal.org/api/drupal/includes%21theme.inc/function/_theme_process_registry/7.x, this locks them down and prevents other modules and themes from using variable processors on our theme suggestion.

There seems to be a lot of misinformation about this. For example the Views doc above claims that using 'base hook' to get processors added automatically, is not possible if our module doesn't have a higher weight that module of the base hook. This doesn't seem to be true.

### DIFF
--- a/modules/ding_tabroll/ding_tabroll.module
+++ b/modules/ding_tabroll/ding_tabroll.module
@@ -21,43 +21,39 @@ function ding_tabroll_theme($existing, $type, $theme, $path) {
         'title' => NULL,
       ),
       'template' => 'templates/views-view-unformatted--ding-tabroll',
-      'original hook' => 'views_view_unformatted',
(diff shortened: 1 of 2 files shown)


cableman:
Reviewed and approved

PR #621, opened 2017-05-31
2092: Translations and settings added

martinb14:
http://platform.dandigbib.org/issues/2092


cableman:
Approved and awaiting release

PR #622, opened 2017-05-31
2429: Handle language facet better

martinb14:
http://platform.dandigbib.org/issues/2429


cableman:
Approved and awaiting release

PR #626, opened 2017-06-16
2404 by Inlead: Title is missing on the user status loan page.

bozeak:
http://platform.dandigbib.org/issues/2404


cableman:
Approved and awaiting release.

PR #628, opened 2017-06-20
2223 by Inlead: Caching collections only on ting_get_object stage.

bozeak:
http://platform.dandigbib.org/issues/2223


cableman:
I think this PR is "overriden" by https://github.com/ding2/ding2/pull/1075

PR #630, opened 2017-06-20
2337 by Inlead: Add new condition for ding_availability.

bozeak:
http://platform.dandigbib.org/issues/2337


cableman:
Approved and awaiting release

PR #632, opened 2017-06-22
1599 by Inlead: Various fixes to IPE module.

inlead:
http://platform.dandigbib.org/issues/1599


cableman:
Reviewed and awaiting release

PR #633, opened 2017-06-27
2223: Use static cache when caching collections from search request

holt83:
http://platform.dandigbib.org/issues/2223


cableman:
I think this PR is "overriden" by https://github.com/ding2/ding2/pull/1075

holt83:
Yes, I'll close it down :)

PR #636, opened 2017-06-30
2361 by Inlead: Only show nodes actually in the queue.

inlead:
http://platform.dandigbib.org/issues/2361


cableman:
Reviewed and approved

PR #637, opened 2017-06-30
2473 by Inlead: Lazy load public lists pane.

inlead:
http://platform.dandigbib.org/issues/2473


cableman:
Reviewed and awaiting release

PR #638, opened 2017-06-30
1340 by Inlead: Bigger image export to Kultunaut.

inlead:
http://platform.dandigbib.org/issues/1340


inlead:
Moved to https://github.com/easySuite/ding2/pull/523.

PR #643, opened 2017-07-03
2553: Ensure holdings is shown for all types and sources

holt83:
http://platform.dandigbib.org/issues/2553


cableman:
Reviewed and awaiting release

PR #646, opened 2017-07-05
2479: Set preferred branch on reserve

BirkAndMe:
http://platform.dandigbib.org/issues/2479


cableman:
Reviewed and awaiting release

PR #651, opened 2017-07-14
B2409

BirkAndMe:
http://platform.dandigbib.org/issues/2409

### DIFF
--- a/modules/ding_user/ding_user.admin.inc
+++ b/modules/ding_user/ding_user.admin.inc
@@ -17,3 +17,12 @@ function ding_user_admin_setting_form($form, &$form_state) {
 
   return system_settings_form($form);
 }
+
+/**
+ * Validate user admin form.
+ */
+function ding_user_admin_setting_form_validate($form, &$form_state) {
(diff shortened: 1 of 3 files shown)


modules/ding_user/ding_user.admin.inc
```diff
@@ -17,14 +17,3 @@ function ding_user_admin_setting_form($form, &$form_state) {
...
-
-/**
- * Validate user admin form.
- */
-function ding_user_admin_setting_form_validate($form, &$form_state) {
-  if (!empty($form_state['values']['ding_user_signup_link'])) {
-    if (!(0 === strpos($form_state['values']['ding_user_signup_link'], 'http'))) {
-      form_set_error('ding_user_signup_link', t('User signup link must start with "http"'));
-    }
-  }
```
kasperg: Disabling validation entirely to allow for relative urls seems a bit harsh. How about using [`valid_url()`](https://api.drupal.org/api/drupal/includes%21common.inc/function/valid_url/7.x) instead?

```php
TRUE == valid_url('/gatewayf/registration/information');
```

PR #566, opened 2017-03-28
Issue #2160: Remove flag module or disable bookmark view

cableman:
This PR contains https://github.com/ding2/ding2/pull/548 plus changes if the flag module can not be disabled.

It simply disables the default "bookmark" views tab as well.

PR #655, opened 2017-07-20
Issue 2160: Ensure default flag bookmarks views are killed

holt83:
http://platform.dandigbib.org/issues/2160


kasperg:
This PR tries to solve the same issue as #566.

We should not try to remove the flag module since we now know that it is being used by `ding_mobilesearch`. Consequently we should instead strive to remove the user bookmarks view which we do not want to appear.

I believe this PR handles this issue better than #566 so I will merge this.

PR #656, opened 2017-07-20
2573: Only remove Alma user email address if changed to empty

kasperg:
Currently we remove user email when doing an update if there is no
change to the email form field. This leads to unwanted removal when the
user is updating other data.

To fix this we change the logic such that we only remove the email if
the value to changed to something empty. Currently the value will be
null if the user removes any current email address.


JrgenGNielsen:
reviewed and approved

PR #657, opened 2017-07-21
2468: Fixed JS error for extended search

BirkAndMe:
http://platform.dandigbib.org/issues/2468


kasperg:
Reviewed and approved. Thanks.

Circle CI error is caused by configuration switching between 14.04 and 12.0.4.

PR #659, opened 2017-07-21
2331: Removed type filter on public lists

BirkAndMe:
http://platform.dandigbib.org/issues/2331


cableman:
Reviewed and approved

PR #661, opened 2017-07-26
B2546

KristianB14:
http://platform.dandigbib.org/issues/2546


themes/ddbasic/sass/components/view/view.scss
```diff
@@ -66,11 +66,13 @@
     // Styleguide 2.1.9
 
     // Small grid
+    @at-root .grid-small#{&},
```
kasperg: It is hard to tell what is going on here without any comments but as far as I can tell you are trying to make the `.grid` classes usable without the dependency on Views markup by using [`@at-root`](http://sassbreak.com/getting-back-to-our-roots/).

I think this is the wrong way to go about something as fundamental as a grid system. It convolutes the structure of the SCSS.

If the grid system should be usable in general then it should be refactored as a general implementation which *then* can be used with Views using mixins, extends or some other method.

KristianB14: With @at-root it's possible to add a class to the top parent of your sass hierarchy. We use it here to add the grid-classes to our top class which is ".view". 

.view {
  @at-root .grid-small#{&} {
  }
}

Compiles to
.grid-small.view {
}

The sass is still dependent on Views markup, but the grid-classes "grid-small", "grid-medium" and "grid-large" can now be placed on both the panel containing the view and the view itself.

kasperg: @KristianB14: Thanks for explaining that. I am not well-versed in @at-root - nor in reading the documentation apparently. 

With that in mind the situation is a lot better.

PR #664, opened 2017-07-26
2440: Fixed. Styling of place2book field. Help text

KristianB14:
http://platform.dandigbib.org/issues/2440

### DIFF
--- a/themes/ddbasic/sass/components/node/event.scss
+++ b/themes/ddbasic/sass/components/node/event.scss
@@ -309,26 +309,26 @@
       position: relative;
       width: 100%;
       float: left;
-      .place2book-label {
-        @include font('base-bold');
-        position: absolute;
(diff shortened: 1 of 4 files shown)


themes/ddbasic/sass/components/node/event.scss
```diff
@@ -305,6 +305,32 @@
...
+        top: 0;
+        left: 0;
+        width: 100px;
+
+        // Tablet
+        @include media($tablet) {
+          @include span-columns(2 of 5);
+        }
+      }
+      .ticket-info {
```
kasperg: In the spirit of improved CSS then how about naming `.place2book-label` and `.ticket-info` such that is is clear they are part of the same component and we can reduce nesting e.g. `place2book-tickets-label` and `place2book-tickets-info`?

PR #667, opened 2017-07-27
B2566

BirkAndMe:
http://platform.dandigbib.org/issues/2566

Del a

### DIFF
--- a/modules/p2/ding_entity_rating_loan_history/ding_entity_rating_loan_history.module
+++ b/modules/p2/ding_entity_rating_loan_history/ding_entity_rating_loan_history.module
@@ -50,7 +50,7 @@ function ding_entity_rating_loan_history_translatables() {
 /**
  * Implements hook_ding_interaction_view().
  */
-function ding_entity_rating_loan_history_ding_interaction_view() {
+function ding_entity_rating_loan_history_ding_interaction_view($extra) {
   global $user;
   if (ding_user_is_provider_user($user)) {
     // Get entities to rate based on loan history.
--- a/modules/p2/ding_entity_rating_loan_history/ding_entity_rating_loan_history.module
+++ b/modules/p2/ding_entity_rating_loan_history/ding_entity_rating_loan_history.module
@@ -50,11 +50,11 @@ function ding_entity_rating_loan_history_translatables() {
 /**
  * Implements hook_ding_interaction_view().
  */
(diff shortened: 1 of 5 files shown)


modules/p2/ding_interaction/plugins/content_types/interaction_pane.inc
```diff
@@ -32,12 +32,6 @@ function ding_interaction_admin_info($subtype, $conf, $context) {
  * Implements hook_content_type_render().
  */
 function ding_interaction_content_type_render($subtype, $conf, $panel_args, $context) {
-  // Only show on front if serendipity is enabled for the frontpage.
-  // See ding_frontpage config at /admin/config/ding/frontend.
-  if (drupal_is_front_page() && !variable_get('ding_frontpage_serendipity') && variable_get('ding_frontpage_promoter') != 'serendipity') {
-    return;
-  }
-
```
kasperg: Why has this been removed. It does not seem to be related to the change at hand.

modules/p2/ding_interaction/plugins/content_types/interaction_pane.inc
```diff
@@ -54,7 +48,7 @@ function ding_interaction_content_type_render($subtype, $conf, $panel_args, $con
     }
   }
 
-  $data = ding_interaction_content($module_list, $conf['count'], $extra);
+  $data = ding_interaction_content($module_list, variable_get('ding_interaction_page_count', DING_INTERACTION_DEFAULT_COUNT), $extra);
```
kasperg: If we do not want to support per pane count then we should remove the option to set it from the pane configuration and the pane title - as far as I can tell this is the only other place where it is used.

modules/p2/ding_entity_rating_loan_history/ding_entity_rating_loan_history.module
```diff
@@ -50,11 +50,11 @@ function ding_entity_rating_loan_history_translatables() {
 /**
  * Implements hook_ding_interaction_view().
  */
-function ding_entity_rating_loan_history_ding_interaction_view($count = DING_ENTITY_RATING_LOAN_HISTORY_MAX_AMOUNT) {
+function ding_entity_rating_loan_history_ding_interaction_view() {
```
kasperg: Sidenote but this implementation seems to be missing the `$extra` parameter.

modules/p2/ding_interaction/plugins/content_types/interaction_pane.inc
kasperg: This seems to be a leftover from #717.

PR #669, opened 2017-07-27
2566: B) Fixed styling of elements in modal-content and multi select

KristianB14:
http://platform.dandigbib.org/issues/2566

### DIFF
--- a/themes/ddbasic/sass/base/form.scss
+++ b/themes/ddbasic/sass/base/form.scss
@@ -296,13 +296,13 @@ input[type="radio"] {
     right: 0;
     top: 0;
   }
+}
 
-  &.multiple {
-    height: auto;
-    &::before,
-    &::after {
-      display: none;
-    }
+.select-wrapper-multiple {
+  height: auto;
+  &::before,
+  &::after {
+    display: none;
   }
 }
 
@@ -326,7 +326,7 @@ select {
 }
 
 select[multiple="multiple"] {
-  position: relative;
+  position: static;
   width: 100%;
   height: auto;
   padding: 10px;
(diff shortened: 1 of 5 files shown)


themes/ddbasic/template.php
```diff
@@ -1196,7 +1196,12 @@ function ddbasic_select($variables) {
   element_set_attributes($element, array('id', 'name', 'size'));
   _form_set_class($element, array('form-select'));
 
-  return '<div class="select-wrapper"><select' . drupal_attributes($element['#attributes']) . '>' . form_select_options($element) . '</select></div>';
+  if($variables['element']['#attributes']['multiple'] == 'multiple') {
```
kasperg: A bit of code style issue here. There should be a space between `if` and `(`.

themes/ddbasic/sass/base/form.scss
```diff
@@ -317,6 +325,16 @@ select {
   }
 }
 
+select[multiple="multiple"] {
+  position: relative;
```
kasperg: Why do we need `position: relative` here? As far as I can tell the element itself is not positioned using `top`, `left` etc. and child elements are not positioned either. If we want to reset the position `absolute` from above then we should use `position: static;`.

themes/ddbasic/sass/base/form.scss
```diff
@@ -296,6 +296,14 @@ input[type="radio"] {
     right: 0;
     top: 0;
   }
+
+  &.multiple {
```
kasperg: `multiple` is very generic. In the spirit of improved CSS I think this should be `select-wrapper-multiple`.

themes/ddbasic/template.php
```diff
@@ -1196,7 +1196,12 @@ function ddbasic_select($variables) {
...
+    return '<div class="select-wrapper multiple"><select' . drupal_attributes($element['#attributes']) . '>' . form_select_options($element) . '</select></div>';
```
kasperg: This should be updated if `multiple` class name is changed.

PR #671, opened 2017-07-27
2122: List sharing

BirkAndMe:
http://platform.dandigbib.org/issues/2122

### DIFF
--- a/modules/p2/ding_list/include/form.inc
+++ b/modules/p2/ding_list/include/form.inc
@@ -581,10 +581,12 @@ The link is active for 7 days from today, after which it will be disabled for se
     ),
     '#tree' => TRUE,
   );
+
+  $url = url($uri['path'], array('absolute' => TRUE));
   $form['public']['link'] = array(
-    '#type' => 'textfield',
+    '#type' => 'item',
     '#title' => t('Share link'),
-    '#value' => url($uri['path'], array('absolute' => TRUE)),
+    '#markup' => l($url, $url),
   );
 
   return $form;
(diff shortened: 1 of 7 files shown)


modules/p2/ding_list/include/form.inc
```diff
@@ -568,6 +569,24 @@ The link is active for 7 days from today, after which it will be disabled for se
...
+    '#description' => t('You can find your public list under !link', array('!link' => l(t('Public lists'), 'public-lists'))),
+    '#states' => array(
+      'visible' => array(
+        ':input[name="status"]' => array('value' => 'public'),
+      ),
+    ),
+    '#tree' => TRUE,
+  );
+  $form['public']['link'] = array(
+    '#type' => 'textfield',
```
kasperg: Why is this a textfield? The user is not able to change the value of it. If it is to be a textfield then I would expect it to be disabled. Otherwise use [a `item` type](https://api.drupal.org/api/drupal/developer%21topics%21forms_api_reference.html/7.x#item) or the like.

modules/p2/ding_list/plugins/content_types/private_share.inc
```diff
@@ -51,8 +51,10 @@ function ding_list_private_share_content_type_render($subtype, $conf, $panel_arg
           'attributes' => array(
             'class' => array(
               'sharer-button', 'sharer-' . $community, 'use-ajax',
+              'sharer-noshare',
```
kasperg: As I read it this is adding the `sharer-noshare` class to elements that also have an `sharer-facebook` class. That seems odd - perhaps this warrants some refactoring?

modules/p2/ding_list/include/menu_callbacks.inc
```diff
@@ -172,13 +172,69 @@ function ding_list_private_share_message_callback($list_id) {
 
   $commands = array();
 
+  drupal_add_js(drupal_get_path('module', 'ding_list') . '/js/ding-list-private-sharer.js');
+
+  $query = array();
+
+  if ($_GET['community'] !== FALSE) {
+    $query['community'] = $_GET['community'];
+  }
```
kasperg: Why is `$community` a query parameter and not part of the menu path? Other parts of the code like `window.sharer` seem to require it anyway.

PR #674, opened 2017-07-28
B2545

b14sha:
http://platform.dandigbib.org/issues/2542

### DIFF
--- a/themes/ddbasic/sass/components/class.scss
+++ b/themes/ddbasic/sass/components/class.scss
@@ -691,6 +691,31 @@ ul.tabs.primary {
 }
 
 // ==========================================================================
+// User menu count
+// ==========================================================================
+.menu-item-count {
+  display: inline-block;
+  position: relative;
+  top: -3px;
+  width: 20px;
+  height: 20px;
+  margin-left: 6px;
+  font-size: 10px;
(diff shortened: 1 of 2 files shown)


themes/ddbasic/sass/components/class.scss
```diff
@@ -691,6 +691,31 @@ ul.tabs.primary {
...
+  text-align: center;
+  line-height: 22px;
+  border-radius: 50%;
+  background-color: $color-primary;
+  color: $color-text-on-primary;
+  
+  &.debts {
+    background-color: $red;
+  }
+  &.reservations-ready {
```
kasperg: In the spirit of an improved class architecture I think these class names could be handled better.

If `.menu-item-count` is the base element then we have modifiers for that. If we keep item counts generic then I would think that the modifiers should be so as well. Suggestions for such an approach would be `.menu-item-count-success` and `.menu-item-count-warning`.

If the user menu is the base element then I would change the base class name to:

`.user-menu-item-count` with modifiers `.user-menu-item-count-debts` and `.user-menu-item-count-reservations-ready`.

DDBdrift:
https://platform.dandigbib.org/issues/2542 | Resolved (tag version)

PR #676, opened 2017-07-28
2456: SMS help text moved

b14sha:
http://platform.dandigbib.org/issues/2456

### DIFF
--- a/modules/ding_user/ding_user.module
+++ b/modules/ding_user/ding_user.module
@@ -480,13 +480,6 @@ function ding_user_form_profile2_form_alter(&$form, &$form_state) {
 
   $form['#validate'][] = 'ding_user_profile_update_validate';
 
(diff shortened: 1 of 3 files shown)


modules/ding_user/ding_user.module
```diff
@@ -481,10 +481,10 @@ function ding_user_form_profile2_form_alter(&$form, &$form_state) {
   $form['#validate'][] = 'ding_user_profile_update_validate';
 
   $fee_sms = variable_get('ding_user_fee_sms', t('Notice that there is a fee for receiving a SMS'));
-  $form['ding_user_fee_sms'] = array(
```
kasperg: Why is `fbs` involved here?

`ding_user` is a generic module for handling users and the variable `ding_user_fee_sms` is a generic variable for controlling SMS fee messages across providers.

If this only applies to FBS then it should be handled in the `fbs` module.

PR #679, opened 2017-07-31
2119: Specific loan history list info

BirkAndMe:
http://platform.dandigbib.org/issues/2119


cableman:
Reviewed and approved

PR #682, opened 2017-07-31
2580: Fixed. Text color and link check

KristianB14:
http://platform.dandigbib.org/issues/2580

### DIFF
--- a/themes/ddbasic/sass/components/node/campaign.scss
+++ b/themes/ddbasic/sass/components/node/campaign.scss
@@ -16,6 +16,14 @@
    position: relative;
    height: 100%;
    z-index: 3;
+   color: $color-text-on-primary;
+  }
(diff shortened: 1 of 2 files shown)


themes/ddbasic/sass/components/node/campaign.scss
```diff
@@ -86,34 +89,37 @@
...
-    color: $color-text-on-primary;
-
-  .ding-campaign-text {
-    @include span-columns(12);
-    padding: 40px 30px;
 
-    .ding-campaign-headline {
-      @include span-columns(8);
+    a {
+      color: $color-text-on-primary;
```
kasperg: Could this be moved to the general link styling for ` .ding-campaign a` to reduce nesting? It is also duplicated now between two campaign types.

PR #683, opened 2017-07-31
2547: View-mode select option removed

KristianB14:
http://platform.dandigbib.org/issues/2547


DDBdrift:
https://platform.dandigbib.org/issues/2547 | Resolved (tag version)

PR #684, opened 2017-07-31
2583: Fixed styling for multiple libraries on news

KristianB14:
http://platform.dandigbib.org/issues/2583

### DIFF
--- a/themes/ddbasic/sass/components/node/news.scss
+++ b/themes/ddbasic/sass/components/node/news.scss
@@ -270,12 +270,11 @@
       }
       .field-item {
         display: inline-block;
-        margin-right: 5px;
         &::after {
-          content: ',';
+          content: ', ';
+          white-space: pre;
         }
         &:last-child {
-          margin-right: 0;
           &::after {
             content: '';
           }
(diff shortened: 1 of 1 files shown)


themes/ddbasic/sass/components/node/news.scss
```diff
@@ -268,6 +268,19 @@
         top: 0;
         right: 7px;
       }
+      .field-item {
+        float: left;
```
kasperg: If the intention is to make the libraries appear as a line of text I would think that using `display: inline-block;` would be more appropriate than floats which can act quite quirky.

themes/ddbasic/sass/components/node/news.scss
```diff
@@ -268,6 +268,19 @@
         top: 0;
         right: 7px;
       }
+      .field-item {
+        display: inline-block;
+        margin-right: 5px;
+        &::after {
+          content: ',';
```
kasperg: This inserts a comma with margin to the right to simulate a space. This could be improved/simplified further by [adding an actual space after the comma](https://stackoverflow.com/a/40963855). This would also require less overriding for the last element.

PR #685, opened 2017-07-31
2581: Openlist authkey setting added

martinb14:
http://platform.dandigbib.org/issues/2581

### DIFF
--- a/modules/p2/ting_openlist/ting_openlist.admin.inc
+++ b/modules/p2/ting_openlist/ting_openlist.admin.inc
@@ -44,7 +44,6 @@ function ting_openlist_settings_form($form, &$form_state) {
     '#default_value' => variable_get('ting_openlist_authkey', ''),
   );
 
-
   // Caching settings.
   $form['caching'] = array(
     '#type' => 'fieldset',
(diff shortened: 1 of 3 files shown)


modules/p2/ting_openlist/ting_openlist.module
```diff
@@ -268,7 +268,13 @@ function ting_openlist_get_soap($wsdl = NULL, $trace = FALSE) {
         )
       );
 
+      // Warn admins to obtain authkey before it will be mandatory
+      if (empty(variable_get('ting_openlist_authkey', '')) && user_access('administer site configuration')) {
+        drupal_set_message(t("Insert a valid !authkey for OpenList.",array("!authkey" => l("authkey","/admin/config/ding/provider/ting_openlist"))) ,"warning", FALSE);
```
kasperg: Please implement such a check using [`hook_requirements()`](https://api.drupal.org/api/drupal/modules%21system%21system.api.php/function/hook_requirements/7.x). This would also be a good place to inform the user that the site has proper access to OpenList.

I do not think it is a good thing to print messages from `ting_openlist_get_soap()`. The message will appear on random places throughout the site.

modules/p2/ting_openlist/ting_openlist.admin.inc
```diff
@@ -37,6 +37,14 @@ function ting_openlist_settings_form($form, &$form_state) {
...
   );
 
+  $form['settings']['ting_openlist_authkey'] = array(
+    '#type' => 'textfield',
+    '#title' => t('Openlist authkey'),
+    '#description' => t('The key that authenticates this library code towards OpenList.'),
+    '#default_value' => variable_get('ting_openlist_authkey', ''),
+  );
+
+
```
kasperg: Nitpicking: Why the extra newline here?

PR #687, opened 2017-08-01
2544: Fixed. Fields removed from display

KristianB14:
http://platform.dandigbib.org/issues/2544


modules/fbs/fbs.info
```diff
@@ -9,13 +9,15 @@ dependencies[] = ding_provider
 dependencies[] = ding_reservation
 dependencies[] = ding_user
 dependencies[] = entity
+dependencies[] = fbs
```
kasperg: This will make `fbs` depend on itself. Features might be silly but that does not seem right?

PR #689, opened 2017-08-01
2455: Periodical issues on renewals

BirkAndMe:
http://platform.dandigbib.org/issues/2455

### DIFF
--- a/modules/ding_loan/plugins/content_types/loans.inc
+++ b/modules/ding_loan/plugins/content_types/loans.inc
@@ -289,6 +289,9 @@ function ding_loan_loans_form_submit($form, &$form_state) {
       $title = $entity->getTitle();
     }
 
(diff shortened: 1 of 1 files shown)


modules/ding_loan/plugins/content_types/loans.inc
```diff
@@ -289,6 +289,10 @@ function ding_loan_loans_form_submit($form, &$form_state) {
       $title = $entity->getTitle();
     }
 
+    if (!empty($form_state['values']['items'][$id]->notes) && !($form_state['values']['items'][$id]->ill_status)) {
```
kasperg: Please add a code comment describing why this logic is needed.

The issue deals with periodicals, the code checks for notes and inter library loan status and then only prints notes.

PR #691, opened 2017-08-01
2486: Locking title on lists

BirkAndMe:
http://platform.dandigbib.org/issues/2486


cableman:
Reviewed and approved

PR #693, opened 2017-08-02
2593: Ting openlist SOAP to REST

BirkAndMe:
http://platform.dandigbib.org/issues/2593

### DIFF
--- a/modules/p2/ting_openlist/TingOpenlistRequestException.php
+++ b/modules/p2/ting_openlist/TingOpenlistRequestException.php
@@ -2,20 +2,38 @@
 
 /**
  * @file
- * The DingProviderConsentException class.
+ * The TingOpenlistRequestException class.
  */
 
 /**
- * Default exception.
+ * TingOpenlistRequestException exception.
  */
 class TingOpenlistRequestException extends Exception {
 
   /**
-   * Construct exception.
+   * The object returned by drupal_http_request().
+   * @var object
    */
-  public function __construct($response) {
+  public $response;
+
+  /**
+   * The URL called.
+   * @var string
+   */
+  public $request_url;
+
+  /**
+   * Construct exception.
+   *
+   * @param object $response
+   *   The object returned by drupal_http_request().
+   * @param string $request_url
+   *   The URL called.
+   */
+  public function __construct($response, $request_url) {
     parent::__construct(t('Ting openlist error.'));
     $this->response = $response;
+    $this->response = $request_url;
   }
 
 }
--- a/modules/p2/ting_openlist/TingOpenlistRequestException.php
+++ b/modules/p2/ting_openlist/TingOpenlistRequestException.php
@@ -1,0 +1,21 @@
+<?php
+
+/**
+ * @file
+ * The DingProviderConsentException class.
+ */
+
+/**
+ * Default exception.
+ */
+class TingOpenlistRequestException extends Exception {
+
+  /**
+   * Construct exception.
+   */
+  public function __construct($response) {
+    parent::__construct(t('Ting openlist error.'));
+    $this->response = $response;
+  }
+
+}
--- a/modules/p2/ting_openlist/ting_openlist.module
+++ b/modules/p2/ting_openlist/ting_openlist.module
@@ -5,6 +5,8 @@
  * Provider and configuration layer for connection with the OpenList service.
  */
 
+define('TING_OPENLIST_DEFAULT_DOMAIN', 'https://test.openlist.ddbcms.dk');
+
 /**
  * Implements hook_ding_provider().
  */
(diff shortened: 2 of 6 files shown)


modules/p2/ting_openlist/ting_openlist.module
```diff
@@ -305,62 +257,65 @@ function ting_openlist_get_soap($wsdl = NULL, $trace = FALSE) {
...
+    implode('/', $parts),
+    array('query' => $query)
+  );
+
+  $response = drupal_http_request($url);
+
+  if ($response->code != '200') {
+    watchdog(
+      'ting_openlist',
+      '<p><strong>Function:</strong><pre>@function</pre></p><p><strong>Arguments:</strong><pre>!arguments</pre></p><p><strong>Response:</strong><pre>!response</pre></p><p><strong>Execution time:</strong><pre>@execution_time</pre></p>',
```
kasperg: Function and arguments should be replaced with the actual full URL being  requested. The url should contain both elements and more and has the advantage of being directly callable from a browser.

modules/p2/ting_openlist/ting_openlist.module
```diff
@@ -305,62 +257,65 @@ function ting_openlist_get_soap($wsdl = NULL, $trace = FALSE) {
...
+    watchdog(
+      'ting_openlist',
+      '@function (@execution_time ms)!details',
+      array(
+        '@function' => $function,
+        '@execution_time' => timer_read(__FUNCTION__),
+        '!details' => $details,
+      ),
+      WATCHDOG_INFO
+    );
```
kasperg: See comment above regarding handling of function/arguments. I would also like to see the response of each request logged [as we do for the Ting Client](https://github.com/ding2/ting-client/blob/master/lib/adapter/TingClientRequestAdapter.php#L72).

With this in mind `ting_openlist_log_details` may be unneeded. Consider refactoring logging to a single function.

modules/p2/ting_openlist/ting_openlist.module
```diff
@@ -305,62 +257,65 @@ function ting_openlist_get_soap($wsdl = NULL, $trace = FALSE) {
...
+      array(
+        '@function' => $function,
+        '@execution_time' => timer_read(__FUNCTION__),
+        '!arguments' => htmlentities(var_export($arguments, TRUE)),
+        '!response' => htmlentities(var_export($response, TRUE)),
+      ),
+      WATCHDOG_ERROR
+    );
+
+    throw new Exception('Ting openlist error.');
```
kasperg: I would prefer not to use a generic exception here. PHP is expanding which errors cause exceptions to be thrown. Relying on the generic exception we cannot handle problems related to the OpenList integration specifically. In the SOAP implementation we at least had the [`SoapFault`](http://php.net/manual/en/class.soapfault.php) to go on.

modules/p2/ting_openlist/ting_openlist.module
```diff
@@ -305,62 +257,65 @@ function ting_openlist_get_soap($wsdl = NULL, $trace = FALSE) {
...
-
-      // If logging is enabled, log the function name and execution time.
-      if (variable_get('ting_openlist_log', FALSE)) {
-        $execution_time = microtime(TRUE) - $start_time;
-
-        $details = '';
-        if ($detailed_log || variable_get('ting_openlist_log_details', FALSE)) {
-          $details = "<pre>" . var_export($arguments, TRUE) . '</pre>';
-        }
+  timer_start(__FUNCTION__);
```
kasperg: :+1:

kasperg:
Please also look into the merge conflicts.

modules/p2/ting_openlist/TingOpenlistRequestException.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+/**
+ * @file
+ * The DingProviderConsentException class.
```
kasperg: Please correct the classname here.

modules/p2/ting_openlist/ting_openlist.module
```diff
@@ -240,128 +242,74 @@ function ting_openlist_field_validate($element, &$form_state, $form) {
...
+      '<strong>Request</strong><pre>!link</pre><strong>Execution time</strong><pre>@execution_time ms</pre><strong>Response</strong><pre>!response</pre>',
+      array(
+        '!link' => l($url, $url, array('external' => TRUE)),
+        '@execution_time' => timer_read(__FUNCTION__),
+        '!response' => htmlentities(var_export($response, TRUE)),
+      ),
+      WATCHDOG_ERROR
+    );
+
+    throw new TingOpenlistRequestException($response);
```
kasperg: Now that we have a proper exception to work with remember to update all the openlist related code which previously caught the generic `Exception` to catch this class instead.

[PHP code may throw exceptions for many different reasons](http://php.net/manual/en/language.errors.php7.php) and catching them all can make problems difficult to track down.

modules/p2/ting_openlist/TingOpenlistRequestException.php
```diff
@@ -0,0 +1,21 @@
...
+ * Default exception.
+ */
+class TingOpenlistRequestException extends Exception {
+
+  /**
+   * Construct exception.
+   */
+  public function __construct($response) {
+    parent::__construct(t('Ting openlist error.'));
+    $this->response = $response;
```
kasperg: If `response` is to be a property then please declare it properly.

modules/p2/ting_openlist/TingOpenlistRequestException.php
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Default exception.
+ */
+class TingOpenlistRequestException extends Exception {
+
+  /**
+   * Construct exception.
+   */
+  public function __construct($response) {
```
kasperg: Please add a proper DocBlock for constructor and arguments.

Also I would like to see the request as a property/argument here as well.

kasperg:
Note that I had to fix some code style issues in 1740eae39f7b2096cfe63655320874d989939264 and 79356b7407a1e47c274f9d65e3bfcca432426b5f.

PR #694, opened 2017-08-02
2256: Get ratings from openlist

BirkAndMe:
http://platform.dandigbib.org/issues/2256

### DIFF
--- a/modules/p2/ding_entity_rating/ding_entity_rating.module
+++ b/modules/p2/ding_entity_rating/ding_entity_rating.module
@@ -54,20 +54,25 @@ function ding_entity_rating_get_user_rating_callback() {
   $ratings = array();
 
   foreach (ding_entity_load_multiple($_GET['ids']) as $object) {
+    // Get rated elements (elements without rating are still included, but with
+    // FALSE as value).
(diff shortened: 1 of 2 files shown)


modules/p2/ding_entity_rating/ding_entity_rating.module
```diff
@@ -57,7 +57,17 @@ function ding_entity_rating_get_user_rating_callback() {
...
+    return $v === FALSE;
+  }));
+
+  $missing = ding_provider_invoke('openlist', 'call_module', 'TingObjectRating', 'getRatings', array(
+    $missing,
+  ));
+
+  $ratings = array_filter($ratings);
+
+  echo drupal_json_encode(array('user_ratings' => $ratings, 'averages' => $missing));
```
kasperg: Variable naming and handling makes this code hard to follow. As I see it:

L58: `$ratings` now contain a map from object ids to the users' own rating. Values will be `FALSE` if the user has not rated the object.

L64: `$missing` now contain an array of object ids which the user has not rated yet.

L66: `$missing` now contain a map from object ids to global average rating.

L68: `$ratings` now contain a map from object ids which the user has actually to rated to the user rating. 

At least use a new more appropriately named variable on L64.

PR #696, opened 2017-08-04
2530: periodical null values

xendk:
Fixes handling of both null values and empty strings in FBS periodicals.


xendk:
Apparently fixed..

PR #699, opened 2017-08-07
1625: Resurrect Alma

xendk:
Seems that there's still some demand for the alma provider, so add it
back as an option in the installer.


JrgenGNielsen:
reviewed and approved

PR #701, opened 2017-08-07
2564: Provide links to materials in debt list

holt83:
http://platform.dandigbib.org/issues/2564

Also fixes: http://platform.dandigbib.org/issues/2161

~~Realised ding_provider_build_entity_id() needed some work, so I sneaked in some additions here and updated comments as well.~~

Forget what I wrote above :)

I moved the commits regarding changes to build_entity_id() to another PR: #721, which is specifically about this. This PR should work without these changes.


cableman:
Code looks fine, just need to be rebase and the sass conflict resolved.

modules/fbs/includes/fbs.debt.inc
```diff
@@ -51,24 +52,41 @@ function fbs_debt_list($account) {
...
-      $material_ids = array();
-      foreach ($fee->materials as $material) {
-        $material_ids[] = $material->materialItemNumber;
-        $record_ids[] = $material->recordId;
+    // If the has associated materials; collect some information about them and
+    // try to build a more useful title.
+    $materials= array();
+    foreach ($fee->materials as $material) {
+      $ding_entity_id = $ids[$material->recordId];
+      if ($ding_entity_id && $entity = ding_entity_load($ding_entity_id)) {
```
cableman: Could the entities loaded in line 37 not be used here. The function "ding_entity_load_multiple" returns the loaded entities and it seams to be the same function the code is in.

holt83: Your right they could. No reason to load them again even though they are cached. I'll update the code tomorrow.

holt83: I looked into this and on second thought we should probably just keep it as is.

The problem is that the entities returned from ding_entity_load_multiple() is keyed by their local id (tid) from the database. So we would have to do something like this:

```
$entities = ding_entity_load_multiple(array_filter($ids));
...
foreach ($fee->materials as $material) {
      $ding_entity_id = $ids[$material->recordId];
      $key = array_search(
        $ding_entity_id,
        array_column(get_object_vars($entities), 'ding_entity_id')
      );
      if ($key) {
        $entity = $entities[$key];
        // Do something with the ent
      }
```
Or else we could loop through allt the entities at once and make a new array keyed by ding_entity_id.

In any case, it's more messy than before. And considering that all the entities gets statically cached as a result of the call to ding_entity_load_multiple(), I think the best approach is to just load them one by one with ding_entitye_load().

holt83: I guess another more compact solultion for making the mapping array would be:

```
$entities = ding_entity_load_multiple(array_filter($ids));
$entities_map = array_flip(array_column(get_object_vars($entities), 'ding_entity_id'));
```

Then we could easily get the key to use on $entities arrray:
```
foreach ($fee->materials as $material) {
  $ding_entity_id = $ids[$material->recordId];
  if (isset($entities[$entities_map[$ding_entity_id]]) {
    $entity = $entities[$entities_map[$ding_entity_id]];
  }
```

Let me know if I should update the code with this approach or just leave it like it is now.

holt83:
After looking at this agian in code review, I added a new commit fixing a notice on empty debt lists:
https://github.com/ding2/ding2/pull/701/commits/e99b3bc22eb82127222ca5e9c0a92ee6b03d895d

modules/fbs/includes/fbs.debt.inc
holt83: Forget what i wrote above. I found a more elegant solution and just made it :)

holt83:
This should be ready for another code review now.

Would still like to know if I should do anything about the scrutinizer issue: 
> Class property $material_type should use lowerCamel naming without underscores

This is from the DingProviderDebt class in the file ding_provider.debt.inc which extends DingEntityBase. It contains some other non lowerCamel named properties. Should I rename them all?

cableman:
It's okay with the non lowerCamel named properties. Just we try to streamline it in the code base.

cableman:
Approved and awaiting release

PR #706, opened 2017-08-09
2598: Translate menu links

BirkAndMe:
http://platform.dandigbib.org/issues/2598

### DIFF
--- a/themes/ddbasic/template.php
+++ b/themes/ddbasic/template.php
@@ -460,6 +460,12 @@ function ddbasic_menu_link__menu_tabs_menu($vars) {
 
   $element['#localized_options']['attributes']['class'][] = 'js-topbar-link';
 
+  // Some links are not translated properly, this makes sure these links are
+  // run through the t function.
+  if ($element['#original_link']['title'] == $element['#original_link']['link_title']) {
+    $element['#title'] = t($element['#title']);
+  }
+
   // Add some icons to our top-bar menu. We use system paths to check against.
   switch ($element['#href']) {
     case 'search':


themes/ddbasic/template.php
```diff
@@ -460,6 +460,12 @@ function ddbasic_menu_link__menu_tabs_menu($vars) {
 
   $element['#localized_options']['attributes']['class'][] = 'js-topbar-link';
 
+  // Some links are not translated properly, this makes sure these links are
+  // run through the t function.
+  if ($element['#original_link']['title'] == $element['#original_link']['link_title']) {
+    $element['#title'] = t($element['#title']);
+  }
+
```
kasperg: This seems like the wrong way to go about this. I cannot spot anywhere in the codebase containing the text "Log out" which is not wrapped in `t()`. I do not know what process went through arriving at this solution but without further information about the issue I cannot accept this.

The previous version of DDBasic did not need this.

BirkAndMe: The previous version of DDBasic also had this t() function (https://github.com/ding2/ding2/blob/master/themes/ddbasic/template.php#L744).

And we're doing it because Drupal's [`_menu_item_localize()`](https://api.drupal.org/api/drupal/includes%21menu.inc/function/_menu_item_localize/7.x) function doesn't run the items callback function on [menu_router items with different title and link_title](https://github.com/drupal/drupal/blob/7.x/includes/menu.inc#L695).

In our case it's the `Logout` and `Login` items that are not translated.

- The `Logout` is not translated, because it's original title is ["Log out"](https://github.com/drupal/drupal/blob/7.x/modules/user/user.module#L1688) and the menu item title is "Logout". We could fix this by renaming it to "Log out" instead.
- The `Login` item is not translated because it doesn't use the `t()` function as the title callback but `user_menu_title()`.
So an alternative fix to this would be to alter the user menu, and use our own custom title callback.
Or use the "user/login" path instead of "user" in the menu item (and rename the title to Log in) so it matches the [user/login title](https://github.com/drupal/drupal/blob/7.x/modules/user/user.module#L1657) (this requires a change in the template.php, and I don't know/remember if we use or listen for the "user" path anywhere else.

I went for this method because it was the fastest to implement, and it had been used before, and it only translates items that have not been run through the `t()` function yet (or items that are not translated).

kasperg: Thanks for the elaboration @BirkAndMe. I stand corrected.

PR #708, opened 2017-08-09
2600: Login feedback

BirkAndMe:
http://platform.dandigbib.org/issues/2600

### DIFF
--- a/modules/ding_user/ding_user.module
+++ b/modules/ding_user/ding_user.module
@@ -1032,10 +1032,14 @@ function ding_user_authenticate_form_callback($form, &$form_state) {
   );
 
   if ($success) {
+    // Close the login form popup.
     $response['#commands'][] = ajax_command_ding_popup_close('ding_user');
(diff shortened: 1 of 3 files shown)


modules/ding_user/ding_user.module
```diff
@@ -1032,6 +1032,9 @@ function ding_user_authenticate_form_callback($form, &$form_state) {
   );
 
   if ($success) {
+    $response['#commands'][] = ajax_command_ding_popup_close('ding_user');
+    $response['#commands'][] = ajax_command_ding_popup('ding_user', t('Authenticated'), t('Authenticated, refreshing page.'));
+
     // Close dialog if successful.
     $response['#commands'][] = ajax_command_ding_popup_close('ding_user', TRUE);
```
kasperg: This change seems odd. We close the existing popup, create a new popup with some feedback information and then close it again. I do not know if/how it works in practice but please add a comment explaining why this is necessary.

The fact that this code is injected before the existing comment does not help either.

PR #711, opened 2017-08-10
2585: Check openlist availability on login

BirkAndMe:
http://platform.dandigbib.org/issues/2585

### DIFF
--- a/modules/p2/ding_list/ding_list.module
+++ b/modules/p2/ding_list/ding_list.module
@@ -435,6 +435,11 @@ function ding_list_ctools_plugin_directory($owner, $plugin_type) {
  * loans list.
  */
 function ding_list_user_login(&$edit, $account) {
+  // Don't update users when openlist isn't available.
+  if (ding_provider_implements('openlist', 'is_available') && !ding_provider_invoke('openlist', 'is_available')) {
+    return;
+  }
+
   if (ding_provider_implements('openlist', 'update_user')) {
     ding_provider_invoke('openlist', 'update_user', $account);
 
--- a/modules/p2/ting_openlist/ting_openlist.module
+++ b/modules/p2/ting_openlist/ting_openlist.module
@@ -304,7 +304,7 @@ function ting_openlist_get_soap($wsdl = NULL, $trace = FALSE) {
  * @return mixed
  *   Response from Openlist.
  */
-function ting_openlist_request($function, array $arguments = array(), $detailed_log = FALSE) {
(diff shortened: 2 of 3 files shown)


modules/p2/ding_list/ding_list.module
```diff
@@ -435,6 +435,11 @@ function ding_list_ctools_plugin_directory($owner, $plugin_type) {
  * loans list.
  */
 function ding_list_user_login(&$edit, $account) {
+  // Don't update users when openlist isn't available.
+  if (ding_provider_implements('openlist', 'is_available') && !ding_provider_invoke('openlist', 'is_available')) {
+    return;
+  }
```
kasperg: I wonder if this is the right way to go handle the issue in general.

As I understand it the problem is that exceptions are thrown and logged during installation [which we also see on Circle CI during site-install](https://circleci.com/gh/ding2/ding2/2256) because openlist is not configured.

This change tried to determine if openlist is available specifically for user login but why this specific case and not other attempts to access openlist?

If possible it would be preferable to  determine why user login is called during the installation phase or why  OpenList is accessed with the parameters instead.

modules/p2/ting_openlist/ting_openlist.module
```diff
@@ -304,7 +304,7 @@ function ting_openlist_get_soap($wsdl = NULL, $trace = FALSE) {
  * @return mixed
  *   Response from Openlist.
  */
-function ting_openlist_request($function, array $arguments = array(), $detailed_log = FALSE) {
+function ting_openlist_request($function, array $arguments = array(), $detailed_log = FALSE, $error_log = TRUE) {
```
kasperg: This adds a new parameter `$error_log`. If this approach stays then this should be documented.

modules/p2/ding_list/ding_list.module
martinb14: User login is called during installation, I assume perhaps from drush to aquire root permissions for all further actions.

We could filter for only provider role, but I believe that for many purposes the lists should work for all users including uid=1.

We havn't found a proper way to determine if the PHP session is part of an installation or a normal production call. If we have such a bit we should use that instead. Any ideas?

kasperg: How about [`drupal_installation_attempted`](https://api.drupal.org/api/drupal/includes%21bootstrap.inc/function/drupal_installation_attempted/7.x)?

If we can use this here with a code comment explaining why this is necessary then the new provider hook is no longer needed and we should also be able to leave out the changes to `ting_openlist_request`.

kasperg:
Closed in favour of #763

PR #723, opened 2017-08-14
2597: Show both related news and events

BirkAndMe:
http://platform.dandigbib.org/issues/2597


kasperg:
I recognize that we cannot simplify the alterations as we have conditions against two separate taxonomies.

PR #726, opened 2017-08-15
2551: Fetch covers by PID

holt83:
http://platform.dandigbib.org/issues/2551

This is dependant on the changes from #724


JrgenGNielsen:
reviewed and approved

PR #730, opened 2017-08-22
2610: Refactor disable custom alias solution

holt83:
http://platform.dandigbib.org/issues/2610

Removing just `$form['path']['alias']` gives rise to problems, since the validate handler for the alias is attached to the parent element. 

In the case where a default value for alias is found it gives a fatal error in `path_form_element_validate()` from path.module, preventing administrators from editing libraries.

I suggest taking the easy route and just prevent acces to whole path form element.


holt83:
Closed in favor of #941

PR #731, opened 2017-08-23
2253: Different color for own ratings

BirkAndMe:
http://platform.dandigbib.org/issues/2253

### DIFF
--- a/themes/ddbasic/sass/p2/p2-rating.scss
+++ b/themes/ddbasic/sass/p2/p2-rating.scss
@@ -92,6 +92,17 @@
           color: $color-secondary;
         }
       }
+
+      .material-item.odd &,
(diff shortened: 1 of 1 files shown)


themes/ddbasic/sass/p2/p2-rating.scss
```diff
@@ -92,6 +92,17 @@
...
+      .material-item.odd &,
+      .ting-object.view-mode-teaser &,
+      .ting-object.view-mode-search-result & {
+        &.active,
+        &.submitted {
+          &::after {
+            color: $color-secondary;
+          }
+        }
+      }
```
kasperg: This really needs some refactoring but I guess we will have to let it pass for now.

PR #734, opened 2017-08-25
2676: Media permissions for local

BirkAndMe:
http://platform.dandigbib.org/issues/2676


JrgenGNielsen:
reviewed and approved

PR #737, opened 2017-08-25
2687: Remove masonry classes from frontpage event view

KristianB14:
http://platform.dandigbib.org/issues/2687

### DIFF
--- a/themes/ddbasic/template.php
+++ b/themes/ddbasic/template.php
@@ -258,6 +258,12 @@ function ddbasic_preprocess_views_view_unformatted(&$vars) {
     $first_node = $nodes[$first_key];
     $vars['type_class'] = drupal_html_class($first_node->type);
   }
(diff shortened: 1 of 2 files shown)


themes/ddbasic/template.php
```diff
@@ -258,6 +258,12 @@ function ddbasic_preprocess_views_view_unformatted(&$vars) {
     $first_node = $nodes[$first_key];
     $vars['type_class'] = drupal_html_class($first_node->type);
   }
+
+  // Set no-masonry to true for frontpage event view
+  if ($vars['view']->name == 'ding_event' && $vars['view']->current_display == 'ding_event_list_frontpage') {
+    $vars['no_masonry'] = TRUE;
```
kasperg: The comment here only explains what the code does - not why it is necessary.

PR #747, opened 2017-09-05
Core: Remove version information from non-release branch

kasperg:
Otherwise builds will include obsolete versions of the code.


cableman:
Approved

PR #752, opened 2017-09-06
2713: User-page links moved from lists panel to new panel. New page

KristianB14:
http://platform.dandigbib.org/issues/2713


modules/ding_user_frontend/ding_user_frontend.module
```diff
@@ -330,6 +336,8 @@ function ding_user_frontend_build_menu($path = NULL) {
       }
     }
 
+    unset($menu['user/%/view']);
```
kasperg: Please add a comment explaining why we unset this menu item specifically.

modules/ding_user_frontend/plugins/content_types/user_overview.inc
```diff
@@ -0,0 +1,120 @@
...
+  );
+
+  // Add specials.
+  try {
+    ding_user_get_creds($account);
+    $uri = entity_uri('user', $account);
+
+    // Fines.
+    $count = count(ding_provider_invoke('debt', 'list', $account));
+    if($count > 0) {
```
kasperg: Codestyle here is off.

modules/ding_user_frontend/plugins/content_types/user_overview.inc
```diff
@@ -0,0 +1,120 @@
...
+    }
+    else {
+      unset($list[DING_USER_FRONTEND_LIST_RESERVATIONS_READY]);
+    }
+
+    // All Reservations.
+    $count = count(ding_provider_invoke('reservation', 'list', $account, DING_RESERVATION_NOT_READY));
+    $list[DING_USER_FRONTEND_LIST_RESERVATIONS] = array(
+      'data' =>
+      '<a href="' . url($uri['path'] . '/status-reservations') . '" class="signature-label"><span class="icon"></span>' . t('Reservations') . '</a>' . '<span class="label">' . $count . '</span>',
```
kasperg: This links to the same page as reservations ready. Is that intentional? No anchor or the like is used.

modules/p2/ding_list/ding_list.module
```diff
@@ -1164,3 +1158,20 @@ function ding_list_views_post_execute(&$view) {
     }
   }
 }
+
+function ding_list_user_overview_items_alter(&$rows, $account) {
```
kasperg: As I understand it `$rows` here are quite close to a render array. Could we add this using a Panels hook such as [`hook_panels_pane_content_alter`](http://api.drupalecommerce.org/api/panels/panels%21panels.api.php/7.x) instead of introducing a new hook?

Also: Please remember documentation.

modules/ding_user_frontend/plugins/content_types/user_overview.inc
KristianB14: Yes - it's intentional.

PR #758, opened 2017-09-07
2288 by Inlead: Re-think the translations of interest periods.

bozeak:
http://platform.dandigbib.org/issues/2288

### DIFF
--- a/modules/alma/alma.module
+++ b/modules/alma/alma.module
@@ -47,7 +47,12 @@ function alma_get_interest_periods() {
   $field_info = field_info_field('field_alma_interest_period');
   $interest_periods = isset($field_info['settings']['allowed_values']) ? $field_info['settings']['allowed_values'] : array();
 
-  return $interest_periods;
+  $options = array();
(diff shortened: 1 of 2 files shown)


modules/alma/alma.module
```diff
@@ -47,7 +47,13 @@ function alma_get_interest_periods() {
   $field_info = field_info_field('field_alma_interest_period');
   $interest_periods = isset($field_info['settings']['allowed_values']) ? $field_info['settings']['allowed_values'] : array();
 
-  return $interest_periods;
```
JrgenGNielsen: The $periods values are in danish, so this won't match anything.
f.ex. <option value="30">1 måned</option>
       <option value="60">2 måneder</option>

guddo: @JrgenGNielsen What do you expect us to do now?

JrgenGNielsen: I suggest we drop the string replace and translation for now - if (or when) it's needed, we'll make an issue of it.

PR #765, opened 2017-09-12
2756: Use accessionDate instead of aquisitionDate for new materials

holt83:
http://platform.dandigbib.org/issues/2756

### DIFF
--- a/modules/ting_field_search/ting_field_search.module
+++ b/modules/ting_field_search/ting_field_search.module
@@ -261,23 +261,22 @@ function _ting_field_search_detect_sort($profile) {
  * Helper function to create new materials CQL.
  */
 function _ting_field_search_new_materials_cql($days) {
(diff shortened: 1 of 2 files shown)


modules/ting_field_search/ting_field_search.module
```diff
@@ -261,23 +261,22 @@ function _ting_field_search_detect_sort($profile) {
  * Helper function to create new materials CQL.
  */
 function _ting_field_search_new_materials_cql($days) {
-  // Get the current date and year.
+  // The best we can do at the moment seems to be accessionDate combined with a
+  // filter on years covered by the specified period. accessionDate alone
+  // doesn't guarantee a correct result, but it adds to the precision of the
+  // year filtering.
```
kasperg: :+1: I appreciate the level of commenting here.

PR #766, opened 2017-09-13
2337

KristianB14:
http://platform.dandigbib.org/issues/2337

In order to make reservable info work on various view modes for ting object, we print out another type field. Before we used "ting_details_type". Now we use "ting_type".

The "ting_type" field has a default color styling which we need to overwrite for specific view-modes.

The "ting_type" field is printed out as a unordered list, which is wrongly targeted by "ting_search_carousel.js". The generic "carousel.find('ul').append(tab.content);" is therefor replaced by "carousel.find('ul.rs-carousel-runner').append(tab.content);"

In template.php on line 843 the collection_list view-mode has been removed. This is done because the js for "ding_availability_with_labels" clashes with the js for the reserve button. We know this is not the best way to do it, but we suggest that that a new issue is created if we still want "ding_availability_with_labels" to appear on the collection_list view-mode, as we consider that the corrections for this are profound


themes/ddbasic/sass/components/ting-object/ting-object.scss
```diff
@@ -259,7 +263,7 @@
       a {
         width: 100%;
         float: left;
-        display: block;
+        //display: block;
```
kasperg: No commented out code i VCS please.

modules/ting_search_carousel/js/ting_search_carousel.js
```diff
@@ -208,7 +208,7 @@
           carousel = first_carousel;
           // Add in extra content.
           if (tab.content) {
-            carousel.find('ul').append(tab.content);
+            carousel.find('ul.rs-carousel-runner').append(tab.content);
```
kasperg: This does not seem to be related to the issue at hand?

themes/ddbasic/sass/components/ting-object/ting-object.scss
```diff
@@ -284,6 +288,11 @@
         }
       }
     }
+    .field-name-ting-type {
+      li {
+        color: $color-standard-text;
+      }
+    }
```
kasperg: This change does not seem to be related to the issue at hand?

themes/ddbasic/sass/components/ting-object/ting-object.scss
```diff
@@ -498,6 +507,11 @@
     .action-button.other-formats {
       display: none;
     }
+    .field-name-ting-type {
+      li {
+        color: $color-standard-text;
+      }
+    }
```
kasperg: This change does not seem to be related to the issue at hand?

themes/ddbasic/template.php
```diff
@@ -840,7 +840,7 @@ function ddbasic_process_ting_object(&$vars) {
   // Notice it's only done on the "search_result" view mode.
   if ($vars['elements']['#entity_type'] == 'ting_object' && isset($vars['object']->in_collection)
       && isset($vars['elements']['#view_mode'])
-      && in_array($vars['elements']['#view_mode'], array('search_result', 'collection_list'))) {
+      && in_array($vars['elements']['#view_mode'], array('search_result'))) {
```
kasperg: Is this change related to the issue at hand?

kasperg:
@KristianB14: Thanks for the detailed description. Those details are hard to infer from the changes alone.

PR #768, opened 2017-09-13
2092: Removed module names as group headers in config

martinb14:
http://platform.dandigbib.org/issues/2092#note-17


cableman:
Reviewed and approved

PR #769, opened 2017-09-14
2826 by Inlead: Do not rely on ting_agency.

inlead:
http://platform.dandigbib.org/issues/2826

Since we made `update_hook()` in https://github.com/ding2/ding2/pull/648 there is not need to rely on the `ting_agency` variable.


cableman:
Reviewed and approved

PR #770, opened 2017-09-14
2814 by Inlead: Change default state of CMS Content flag.

inlead:
http://platform.dandigbib.org/issues/2814

Default state of flag changed in the export of each CTs. Should be set on the next feature-revert.
The `dependencies` added to the .info-files are from the default export and was not manually removed.


JrgenGNielsen:
reviewed and approved

PR #773, opened 2017-09-18
b2542: Wrong variable-name fixed

KristianB14:
http://platform.dandigbib.org/issues/2542


cableman:
Reviewed and approved

PR #774, opened 2017-09-18
2171: Staff rate materials

BirkAndMe:
http://platform.dandigbib.org/issues/2171


cableman:
Reviewed and approved

PR #776, opened 2017-09-18
2364 by Inlead: Catch error properly.

inlead:
http://platform.dandigbib.org/issues/2364


cableman:
Reviewed and awaiting release

PR #780, opened 2017-09-19
2838: Ensured profile2 submit handler is available

cableman:
Also fixed notice errors on staff profile page.

See: https://platform.dandigbib.org/issues/2838


kasperg:
Merged manually after fixing Scrutinizer issue.

PR #785, opened 2017-09-21
2843: Pane-title added to related content panel

KristianB14:
https://platform.dandigbib.org/issues/2843


cableman:
Reviewed and approved

PR #786, opened 2017-09-21
2083: Improve timing of drupal_set_messages() check

holt83:
https://platform.dandigbib.org/issues/2083


cableman:
Reviewed and approved

PR #787, opened 2017-09-22
1748: Track when materials is added to lists

holt83:
https://platform.dandigbib.org/issues/1748

### DIFF
--- a/modules/ding_webtrends/js/webtrends_add.js
+++ b/modules/ding_webtrends/js/webtrends_add.js
@@ -14,7 +14,7 @@
       // Array of selectors for links to track via webtrends.
       var webtrendsAdd = [
         '.action-button.reserve-button',
-        '.action-button.bookmark-button',
+        '.ding-list-add-button .use-ajax',
         '.group-holdings-available a'
       ];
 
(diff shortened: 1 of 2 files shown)


modules/ding_webtrends/js/webtrends_add.js
```diff
@@ -14,7 +14,7 @@
       // Array of selectors for links to track via webtrends.
       var webtrendsAdd = [
         '.action-button.reserve-button',
-        '.action-button.bookmark-button',
+        '.ding-list-add-button .use-ajax',
```
kasperg: Why is `.use-ajax` included as a selector here? Whether or not to use AJAX does not seem relevant to the tracking.

holt83: We should only track buttons which uses AJAX. 

The button to add a new list for example, results in page reload and gets tracked the normal way. Using .use-ajax in the selector leaves this button out.

kasperg: But then the element we want should have a more specific class. AJAX is a byproduct.

holt83: I'm not sure I understand: "AJAX is a byproduct". Can you elaborate?

We want to target the links using ajax, so I'm having a hard time seeing the problem here.

What do you suggest for a more specific class?

holt83: I updated the PR to add a more specific class to the links which attaches materials to lists (still checking for use-ajax as a precaution).

PR #794, opened 2017-09-26
2861: Field-group name added as class. Field removed from view-mode

KristianB14:
https://platform.dandigbib.org/issues/2861

The field_group module was updated in this issue https://platform.dandigbib.org/issues/1905

After the update the field-group name was no longer printet out as a class. We have added a pre render to do this.

Furthermore a field has been removed from the teaser view-mode, and the correct class-names are used in the scss.

### DIFF
--- a/modules/ding_ddbasic/ding_ddbasic.module
+++ b/modules/ding_ddbasic/ding_ddbasic.module
@@ -117,3 +117,24 @@ function ding_ddbasic_module_implements_alter(&$implementations, $hook) {
     $implementations['ding_ddbasic'] = $group;
   }
 }
+
+/**
+ * Implements hook_field_group_pre_render().
+ *
+ * This function gives you the oppertunity to create the given
+ * wrapper element that can contain the fields.
+ * In the example beneath, some variables are prepared and used when building the
+ * actual wrapper element. All elements in drupal fapi can be used.
+ *
+ * Note that at this point, the field group has no notion of the fields in it.
+ *
+ * There is also an alternative way of handling this. The default implementation
+ * within field_group calls "field_group_pre_render_<format_type>".
+ * @see field_group_pre_render_fieldset.
+ *
+ * @param Array $elements by address.
+ * @param Object $group The Field group info.
+ */
+function ding_ddbasic_field_group_pre_render(& $element, $group, & $form) {
+  $group->format_settings['instance_settings']['classes'] .= ' ' . drupal_html_class($group->group_name);
+}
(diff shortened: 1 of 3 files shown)


modules/ding_ddbasic/ding_ddbasic.module
```diff
@@ -117,3 +117,24 @@ function ding_ddbasic_module_implements_alter(&$implementations, $hook) {
...
+ * actual wrapper element. All elements in drupal fapi can be used.
+ *
+ * Note that at this point, the field group has no notion of the fields in it.
+ *
+ * There is also an alternative way of handling this. The default implementation
+ * within field_group calls "field_group_pre_render_<format_type>".
+ * @see field_group_pre_render_fieldset.
+ *
+ * @param Array $elements by address.
+ * @param Object $group The Field group info.
```
kasperg: Duplicating the default documentation for [`hook_field_group_pre_render()`](http://api.drupalhelp.net/api/field_group/field_group.api.php/function/hook_field_group_pre_render/7) is unnecessary. Please remove it.

modules/ding_ddbasic/ding_ddbasic.module
```diff
@@ -117,3 +117,24 @@ function ding_ddbasic_module_implements_alter(&$implementations, $hook) {
...
+ */
+function ding_ddbasic_field_group_pre_render(& $element, $group, & $form) {
+  $group->format_settings['instance_settings']['classes'] .= ' field-group-format ' . drupal_html_class($group->group_name);
```
kasperg: If the theme depends on these classes is it possible to add these classes in the theme code instead of in `ding_ddbasic`?

modules/ding_ddbasic/ding_ddbasic.module
```diff
@@ -117,3 +117,24 @@ function ding_ddbasic_module_implements_alter(&$implementations, $hook) {
...
+ * Note that at this point, the field group has no notion of the fields in it.
+ *
+ * There is also an alternative way of handling this. The default implementation
+ * within field_group calls "field_group_pre_render_<format_type>".
+ * @see field_group_pre_render_fieldset.
+ *
+ * @param Array $elements by address.
+ * @param Object $group The Field group info.
+ */
+function ding_ddbasic_field_group_pre_render(& $element, $group, & $form) {
```
kasperg: Why is there a space between `&` and `$` in the paramters here? That is not a style we follow elsewhere.

modules/ding_ddbasic/ding_ddbasic.module
KristianB14: We tried to do this, but for some reason the hook doesn't work from the theme, even though it should.

kasperg: Ok. I guess we have to solve it this way then.

PR #795, opened 2017-09-27
2869: Only published nodes in ting references

BirkAndMe:
https://platform.dandigbib.org/issues/2869


modules/ting_reference/ting_reference.module
```diff
@@ -335,11 +335,22 @@ function ting_reference_field_formatter_view($entity_type, $entity, $field, $ins
...
+          }
+        }
+
+        if (!empty($entities)) {
+          $build = entity_view($entity_type, $entities, 'search_result');
+
+          // Don't show links. Should be an option.
+          unset($build['links']);
+          $element[] = $build;
+        }
```
kasperg: It seems like this problem could be handled by adding a condition to `entity_load()` on L337. That would simplify things.

```php
$entities = entity_load($entity_type, array($entity_id), array('status' => NODE_PUBLISHED));
```

Have you tried this?

PR #798, opened 2017-09-27
2819: fix dibs capture

xendk:
Hardcode DIBS capture URL to avoid moneyloss in the future.


cableman:
Reviewed and approved

PR #799, opened 2017-09-27
2520: Hide IPE actions on frontend

BirkAndMe:
https://platform.dandigbib.org/issues/2520

### DIFF
--- a/themes/ddbasic/sass/components/panel/class-panel.scss
+++ b/themes/ddbasic/sass/components/panel/class-panel.scss
@@ -615,3 +615,12 @@
   max-width: 1260px;
   margin: 0 auto;
 }
+
+.panels-ipe-on {
+  // The lazy_pane module will occasionally include the IPE actions, on the pane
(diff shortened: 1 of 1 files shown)


themes/ddbasic/sass/components/panel/class-panel.scss
```diff
@@ -615,3 +615,12 @@
...
   margin: 0 auto;
 }
+
+.panels-ipe-on {
+  // The lazy_pane module will occasionally include the IPE actions, on the pane
+  // being loaded. This is due to the cache keys lazy_pane work with. There's no
+  // hook for us to fix this problem, which is why we hide these actions here.
+  // The IPE module will set the display inline, when an admin is editing a page
+  // so the backend still works with this display none.
+  display: none;
```
kasperg: I appreciate the level of commenting here.

The proper fix would probably be to patch `lazy_pane` to provide a richer cache id but we cannot do that at the moment. The comment indicates that this has been tested regardless so I think it is OK for now.

PR #805, opened 2017-10-02
2907: Set signup link in update

BirkAndMe:
https://platform.dandigbib.org/issues/2907


modules/ding_gatewayf/modules/ding_gatewayf_registration/ding_gatewayf_registration.install
```diff
@@ -16,6 +16,13 @@ function ding_gatewayf_registration_uninstall() {
  * Implements hook_install().
  */
 function ding_gatewayf_registration_install() {
+  ding_gatewayf_registration_update_7001();
```
kasperg: I am not a fan of referencing update hooks from install hooks. If they should share functionality then please implement a separate function à la `ding_gatewayf_registration_set_detault_link()` and then call it from both `hook_install` and  `hook_update`.

PR #807, opened 2017-10-02
2671 by Inlead: Use drupal element hidden functionality.

inlead:
https://platform.dandigbib.org/issues/2671

### DIFF
--- a/modules/ding_ipe_filter/css/ipe.css
+++ b/modules/ding_ipe_filter/css/ipe.css
@@ -15,6 +15,7 @@ div.ipe-popup {
 div.ipe-popup ul {
   padding-left: 0;
   width: 100%;
+  height: 100%;
   list-style-position: outside;
 }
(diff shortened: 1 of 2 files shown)


modules/ding_ipe_filter/css/ipe.css
```diff
@@ -32,6 +33,6 @@ div.ipe-popup .item-list ul li {
   overflow: visible;
 }
 .ding-ipe-menu-item {
-  width: calc(100% - 18px);
```
kasperg: As far as I can tell changes to `ipe.css` do not seem related to the change description or issue. Are they intentional?

guddo: They solve the problems described in the issue. However not reflected in the commit message.

kasperg: OK. Thanks for the update.

PR #808, opened 2017-10-04
2693 by Inlead: Push event dates as ISO-8601.

inlead:
https://platform.dandigbib.org/issues/2693


modules/ding_mobilesearch/ding_mobilesearch.module
```diff
@@ -143,12 +145,12 @@ function ding_mobilesearch_ding_mobilesearch_node_export_mapping($node) {
       ),
       'created' => array(
         'name' => t('Created'),
-        'value' => $node->created,
```
kasperg: This change is added multiple places. Date/timezone handling is generally complex. How about moving the handling to a central function à la `ding_mobilesearch_format_date($timestamp, $timezone)`?

PR #811, opened 2017-10-04
2908: Reserve button in carousel

BirkAndMe:
https://platform.dandigbib.org/issues/2908

- Use Drupal.behaviors in ding_availability.js, instead of document ready.
- Include Drupal.settings in the ting_search_carousel ajax call.

### DIFF
--- a/modules/ding_availability/js/ding_availability.js
+++ b/modules/ding_availability/js/ding_availability.js
@@ -9,7 +9,7 @@
   // Cache of fetched availability information.
   Drupal.DADB = {};
 
-  Drupal.behaviors.dingAvailabilityAttach = {
+  Drupal.behaviors.dingAvailability = {
     attach: function(context, settings) {
       var ids = [];
       var html_ids = [];
--- a/modules/ding_availability/js/ding_availability.js
+++ b/modules/ding_availability/js/ding_availability.js
@@ -36,12 +36,12 @@
       // holdings in order to determine whether the material is a
       // periodical. This is a bit of a hack, but it's the quickest way
       // of fixing the problem right now.
-      if ($('.reserve-button').size() > 0) {
+      if ($('.reserve-button', context).size() > 0) {
         settings.ding_availability_mode = 'holdings';
       }
 
       $.each(html_ids, function (index, id) {
-        $('#' + id).addClass('pending');
+        $('#' + id, context).addClass('pending');
       });
 
       // Fetch availability.
@@ -69,7 +69,7 @@
             });
           },
           error: function () {
-            $('div.loader').remove();
+            $('div.loader', context).remove();
           }
         });
       }
--- a/modules/ding_availability/js/ding_availability.js
+++ b/modules/ding_availability/js/ding_availability.js
@@ -9,79 +9,80 @@
   // Cache of fetched availability information.
   Drupal.DADB = {};
 
-  $(document).ready(function () {
-    var settings = Drupal.settings;
-    var ids = [];
-    var html_ids = [];
+  Drupal.behaviors.dingAvailabilityAttach = {
+    attach: function(context, settings) {
+      var ids = [];
+      var html_ids = [];
 
-    // Loop through the materials given in the settings and collect
-    // HTML ids and entity_ids.
-    if (settings.hasOwnProperty('ding_availability')) {
-      $.each(settings.ding_availability, function (id, entity_ids) {
-        $.each(entity_ids, function (index, entity_id) {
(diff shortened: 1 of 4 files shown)


modules/ding_availability/js/ding_availability.js
```diff
@@ -9,79 +9,80 @@
...
-        $.each(entity_ids, function (index, entity_id) {
-          if (Drupal.DADB[entity_id] === undefined) {
-            Drupal.DADB[entity_id] = null;
-            ids.push(entity_id);
-            html_ids.push(id);
-          }
-        });
-      });
-    }
+  Drupal.behaviors.dingAvailabilityAttach = {
```
kasperg: Please remove `Attach` to the name of the behavior? `attach` is a separate function within the behavior.

modules/ding_availability/js/ding_availability_labels.js
```diff
@@ -9,7 +9,7 @@
   // Cache of fetched availability information.
   Drupal.DADB = {};
 
-  Drupal.behaviors.dingAvailabilityAttach = {
+  Drupal.behaviors.dingAvailabilityLabelsAttach = {
```
kasperg: See comment about naming above.

modules/ting_search_carousel/ting_search_carousel.pages.inc
```diff
@@ -20,10 +20,13 @@ function ting_search_carousel_result($index, $offset) {
...
   $offset = $chunk['next_offset'];
 
+  $js = drupal_add_js();
+
   // Return JSON output.
   drupal_json_output(array(
     'content' => $content,
     'index' => $index,
     'offset' => $offset,
+    'settings' => drupal_array_merge_deep_array($js['settings']['data']),
```
kasperg: I am not a fan of this approach. It pollutes the concept of settings in the behavior. I wonder why it is needed. That is hard to tell from the pull request and commit comments.

If the rendering of the carousel included data that the frontend code needs to react properly then I would much prefer that this was contained with the rendering and not spread between rendering and settings. Would it be possible to add the required data as a data attribute instead?

modules/ding_availability/js/ding_availability.js
```diff
@@ -9,79 +9,80 @@
...
+                ding_availability_update_availability(id, entity_ids);
+              }
+              else {
+                // Update holding information.
+                ding_availability_update_holdings(id, entity_ids);
+              }
+            });
+          },
+          error: function () {
+            $('div.loader').remove();
```
kasperg: This should probably be limited to the `context` of the behavior.

modules/ding_availability/js/ding_availability.js
```diff
@@ -9,79 +9,80 @@
...
-      $.ajax({
-        dataType: "json",
-        url: path,
-        success: function (data) {
-          $.each(data, function (id, item) {
-            // Update cache.
-            Drupal.DADB[id] = item;
-          });
+      $.each(html_ids, function (index, id) {
+        $('#' + id).addClass('pending');
```
kasperg: This could be limited to the `context` of the behavior, but it might not make sense since we are locating by id anyhow.

modules/ding_availability/js/ding_availability.js
```diff
@@ -9,79 +9,80 @@
...
-    });
+      // If there's any reservation buttons, switch to holdings. We have
+      // periodicals that's a bit off an odd one in that they can both
+      // have issues, which means that the reservation button for the
+      // main object should be disabled, or not have any issues, which
+      // means that it should be left alone. So we need to fetch full
+      // holdings in order to determine whether the material is a
+      // periodical. This is a bit of a hack, but it's the quickest way
+      // of fixing the problem right now.
+      if ($('.reserve-button').size() > 0) {
```
kasperg: This should be limited by `context`.

modules/ting_search_carousel/ting_search_carousel.pages.inc
BirkAndMe: In this specific issue, we needed to include the settings, so the ding_availability settings would be included. Adding that information to a `data-` attribute would require a rewrite of how ding_availability works.
Including all the settings should allow other modules that also utilize the settings array to work on the materials loaded in the carousel.

kasperg:
@cableman Is is PR superseded by #845?

cableman:
@kasperg yes it is :-) #845 also fixes this.

cableman:
This issues related to missing reservation links have also been fixed in core in issue 1770.

PR #816, opened 2017-10-06
2917: More flexible width for view-more links on mobile

KristianB14:
https://platform.dandigbib.org/issues/2917


DDBdrift:
https://platform.dandigbib.org/issues/2917 | Resolved (tag version)

PR #820, opened 2017-10-09
2913: jQuery ui touch punch altered to enable vertical scroll

KristianB14:
https://platform.dandigbib.org/issues/2913


modules/ting_search_carousel/js/jquery.ui.touch-punch.js
```diff
@@ -0,0 +1,203 @@
+/*!
```
kasperg: This is an external dependency. Please handle it using `drush make`.

kasperg: Please do not include third party code in the codebase. They should be downloaded using `drush make`.

modules/ting_search_carousel/js/jquery.ui.touch-punch.js
```diff
@@ -0,0 +1,203 @@
...
+   */
+  function simulateMouseEvent (event, simulatedType) {
+
+    // Ignore multi-touch events
+    if (event.originalEvent.touches.length > 1) {
+      return;
+    }
+
+    // @COMMENTED OUT - line is moved to if-statement below
+    //event.preventDefault();
```
kasperg: If we need to alter third party code then please do it using patches. Not by hacking directly in the source code.

DDBdrift:
https://platform.dandigbib.org/issues/2913 | Closed

PR #824, opened 2017-10-11
2928: Only remove empty holding info

BirkAndMe:
https://platform.dandigbib.org/issues/2928

### DIFF
--- a/modules/ding_availability/js/ding_availability_labels.js
+++ b/modules/ding_availability/js/ding_availability_labels.js
@@ -147,13 +147,9 @@
         group.append(element);
 
         // Remove empty groups.
-        $('.js-available, .js-reservable, .js-unavailable', groups_wrapper).each(function () {
-          var $this = $(this);
-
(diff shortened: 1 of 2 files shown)


modules/ding_availability/js/ding_availability_labels.js
```diff
@@ -147,6 +145,15 @@
...
         group.append(element);
+
+        // Remove empty groups.
+        $('.js-available, .js-reservable, .js-unavailable', groups_wrapper).each(function () {
+          var $this = $(this);
+
+          if ($this.children('.search-result--availability-link').length === 0) {
+            $this.remove();
+          }
+        });
```
kasperg: Could this be simplified a bit á la:

```javascript
$('.js-available, .js-reservable, .js-unavailable', groups_wrapper)
  .not(':has(.search-result--availability-link)')
  .remove();
```

kasperg: Also: Should `.search-result--availability-link` have a `js-` prefix?

PR #614, opened 2017-05-22
Slick for search carousel, second coming

xendk:
Supersedes #271.

Changes JS library for creating carousels to Slick, generalises carousels into ding_carousel, which allows for creating both tabbed and standalone carousels and updates P2 to use ding_carousel.

Also numerous bug fixes.

PR #271, opened 2016-07-05
1770: Slick for search carousel

xendk:
Replaces old carousel library with slick.

Also some bugfixes.

PR #826, opened 2017-10-11
2815: Slick for search carousel, second coming

inlead:
https://platform.dandigbib.org/issues/2815

Supersedes #614, which superseded #271.

Changes JS library for creating carousels to Slick, generalises carousels into ding_carousel, which allows for creating both tabbed and standalone carousels and updates P2 to use ding_carousel.

Also numerous bug fixes.


xendk:
This is rendered irrelevant by #845

kasperg:
Thanks for the heads up @xendk.

PR #832, opened 2017-10-12
2989: Fix fetching of covers for attachment- and local-posts

holt83:
https://platform.dandigbib.org/issues/2989


DDBdrift:
https://platform.dandigbib.org/issues/2989 | Closed

PR #836, opened 2017-10-17
2742: Event list, search in location

BirkAndMe:
https://platform.dandigbib.org/issues/2742


JrgenGNielsen:
reviewed and approved

PR #840, opened 2017-10-17
2699: add ding_get_private_key function

JrgenGNielsen:
https://platform.dandigbib.org/issues/2699

### DIFF
--- a/modules/ding_user/ding_user.install
+++ b/modules/ding_user/ding_user.install
@@ -11,6 +11,7 @@
  */
 function ding_user_install() {
   ding_user_create_entity_cache_table();
+  ding_user_create_private_key();
 }
 
 /**
@@ -27,6 +28,13 @@ function ding_user_create_entity_cache_table() {
     $schema['description'] = 'Cache table used to store profile2 entity records.';
     db_create_table('cache_entity_profile2', $schema);
   }
+}
+
+/**
+ * Creates the ding_user_private_key variable.
+ */
+function ding_user_create_private_key() {
+  variable_set('ding_user_private_key', drupal_get_private_key());
 }
 
 /**
@@ -129,4 +137,11 @@ function ding_user_update_7008(&$sandbox) {
 function ding_user_update_7009(&$sandbox) {
   db_query("DELETE a.* FROM redirect a INNER JOIN ( SELECT concat('user/', u.uid) s FROM users u INNER JOIN authmap a ON u.uid=a.uid ) b ON a.redirect=b.s;")
     ->execute();
+}
+
+/**
+ * Create the ding_user_private_key variable.
+ */
+function ding_user_update_7010(&$sandbox) {
+  ding_user_create_private_key();
 }
--- a/modules/ding_user/ding_user.module
+++ b/modules/ding_user/ding_user.module
@@ -673,12 +673,37 @@ function ding_user_default_authname($name) {
   // exactly the size of the authname field in the database. So we ask hash
   // for binary data and lets _password_base64_encode base64 encode it to 86
   // characters, leaving enough room for our suffix.
-  $hash = hash('sha512', drupal_get_private_key() . $name, TRUE);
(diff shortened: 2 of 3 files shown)


modules/ding_user/ding_user.module
```diff
@@ -682,6 +682,32 @@ function ding_user_default_authname($name) {
...
+    if (!($key = variable_get('drupal_private_key', 0))) {
+      $key = drupal_random_key();
+    }
+    watchdog(
+      'ding_user',
+      'New ding_private_key generated. This may break the link to user\'s lists, if on a production site', 
+      array(),
+      WATCHDOG_WARNING
+    );
+    variable_set('ding_private_key', $key);
```
kasperg: I do not agree with this implementation.

As I understand it then by setting a new key if key retrieval fails then we have in essence recreated the problematic behavior from Drupal Core that caused us to implement this change to begin with as discussed with @christianvandel.

I suggest a solution which:

At runtime:
- If `ding_private_key` cannot be retrieved then cause an error - e.g. by throwing an exception. I know that this will cause an error for the end user but the alternative means that we create new user accounts and thus breaking references to lists, comments etc. Since the error should be a temporary situation I think that is preferable.

At install/upgrade time:
- Set `ding_private_key` to `drupal_private_key`.

modules/ding_user/ding_user.module
```diff
@@ -682,6 +682,32 @@ function ding_user_default_authname($name) {
 }
 
 /**
+ * Private key variable used to generate tokens.
+ *
+ * Apparently the key generated in drupal_get_private_key() may change,
+ * so we'll get it from 'ding_private_key' variable instead.
+ * Preferably set the 'ding_private_key' variable in local.settings.php
```
kasperg: A reference to the issue on drupal.org would be good here.

JrgenGNielsen: @kasperg: I can't seem to find any drupal.org issues regarding it - do you have a link?

modules/ding_user/ding_user.install
```diff
@@ -30,6 +31,18 @@ function ding_user_create_entity_cache_table() {
...
+ * Creates the ding_private_key variable.
+ */
+function ding_user_create_private_key() {
+  // variable may be set in local.settings.php, but not in db.
+  if ($key = variable_get('ding_private_key', 0)) {
+    variable_set('ding_private_key', $key);
+    return;
+  }
+  variable_set('ding_private_key', drupal_get_private_key());
+}
```
kasperg: The logic here seems convoluted with unnecessary if-statements and early return.

```php
function ding_user_create_private_key() {
  // variable may be set in local.settings.php, but not in db.
  $key = variable_get('ding_private_key', drupal_get_private_key()));
  variable_set('ding_private_key', $key);
}

File: modules/ding_user/ding_user.install [hunk 4]
```diff
@@ -30,6 +31,18 @@ function ding_user_create_entity_cache_table() {
 }
 
 /**
+ * Creates the ding_private_key variable.
+ */
+function ding_user_create_private_key() {
+  // variable may be set in local.settings.php, but not in db.
+  if ($key = variable_get('ding_private_key', 0)) {
```
kasperg: Naming here is off.

If this variable is owned by `ding_user` then it must be prefixed with `ding_user` e.g. `'ding_user_private_key'. Otherwise it must be managed by the `ding2` profile.

modules/ding_user/ding_user.module
```diff
@@ -682,6 +682,31 @@ function ding_user_default_authname($name) {
...
+ *
+ * @see https://www.drupal.org/node/487366
+ *
+ * @return string
+ *   The private key.
+ */
+function ding_get_private_key() {
```
kasperg: Function name is if off here as well.

modules/ding_user/ding_user.module
kasperg: :+1: You have added it.

PR #842, opened 2017-10-18
3005 by Inlead: Use libraries api for the js cookie.

NicolaeGrigoriu:
https://platform.dandigbib.org/issues/3005


NicolaeGrigoriu:
There are several scrutinizer issues regarding Cookies library. However, based on the documentation https://github.com/js-cookie/js-cookie#basic-usage it is acceptable to use the way it is now.

JrgenGNielsen:
reviewed and approved

PR #843, opened 2017-10-18
2671 by Inlead: Consider initialized slick carousels.

NicolaeGrigoriu:
https://platform.dandigbib.org/issues/2671

### DIFF
--- a/modules/ting_carousel/js/ting-carousel.js
+++ b/modules/ting_carousel/js/ting-carousel.js
@@ -31,7 +31,7 @@
 
   Drupal.behaviors.ting_carousel = {
     attach: function(context, settings) {
-      $('.ting-carousel, .ting-carousel-field > .field-items', context).each(function (delta, carousel) {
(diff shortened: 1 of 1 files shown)


cableman:
Please not that I'am working on new intergration with slick in https://platform.dandigbib.org/issues/1770

NicolaeGrigoriu:
I think(~~hope~~ :) ) that these changes will not affect the way slick working.

cableman:
They will stream-line the usage of carousels, so we don't use different libraries etc. But only uses slick (version 1.8) the same way across the site and only the library once etc.

cableman:
See for more information https://github.com/ding2/ding2/pull/845

guddo:
This line will most likely affect the IPE handling as well: https://github.com/ding2/ding2/pull/845/files#diff-1d4c6fc7228e5c51aed964a5d51b7b22R247

cableman:
I'am closing this PR as the ting_carousel module have been removed from the codebase.

PR #848, opened 2017-10-20
2875: Fixed - Email + Phone validation and rebuild on submit

b14sha:
https://platform.dandigbib.org/issues/2875

### DIFF
--- a/modules/fbs/fbs.module
+++ b/modules/fbs/fbs.module
@@ -496,11 +496,11 @@ function fbs_form_profile2_form_alter(&$form, &$form_state) {
     '#weight' => 3,
   );
   
-  //Add validation to email
+  // Add validation to email.
   $form['profile_provider_fbs']['field_fbs_mail']['#element_validate'] = array('fbs_form_profile2_mail_validate');
-  //Add validation to phone
+  // Add validation to phone.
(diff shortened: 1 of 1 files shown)


modules/fbs/fbs.module
```diff
@@ -495,6 +495,54 @@ function fbs_form_profile2_form_alter(&$form, &$form_state) {
...
+  $form['#submit'][] = 'fbs_form_profile2_rebuild_submit';
+}
+
+/**
+ * Mail validation for fbs_form_profile2
+ */
+function fbs_form_profile2_mail_validate($element, $form_state) {
+  $email = $form_state['input']['profile_provider_fbs']['field_fbs_mail']['und'][0]['value'];
+  if (!empty($email)) {
+    if (!filter_var($email, FILTER_VALIDATE_EMAIL) && !preg_match('/@.+\./', $email)) {
```
JrgenGNielsen: Why not just use "!filter_var($email, FILTER_VALIDATE_EMAIL)"? 
Adding the "!preg_match('/@.+\./', $email)" expression will allow email adresses like 'joe@example.com ' or 'foo @ bar . bas' to pass.
Btw, it might be a good idea to strip whitespace from start and end - it's a common copy/paste error.

JrgenGNielsen:
reviewed and approved

JrgenGNielsen:
still approved :-)

PR #851, opened 2017-10-24
3017: remove random

xendk:
Removes the random ordering that the datawell doesn't like.

Also cleans up code style of ting.module


modules/ting/ting.module
```diff
@@ -82,7 +82,7 @@ function ting_menu() {
     'position' => 'left',
     'weight' => 20,
     'page callback' => 'system_admin_menu_block_page',
-    'access arguments' => array('access administration pages'),
+    'access arguments' => array('administer ting settings'),
```
cableman: Will this require an change in ding_permissions module as well?

xendk: Added the permission to the administrative roles.

PR #852, opened 2017-10-24
1770: fixes for carousels

cableman:
* Fixed access callback for ajax loading ding lists in carousels
* Ensures that 0-hit search carousel don't make infinite loop in JS. 
* Fixes ratings in carousel items.

See https://platform.dandigbib.org/issues/1770

### DIFF
--- a/modules/ding_carousel/ding_carousel.theme.inc
+++ b/modules/ding_carousel/ding_carousel.theme.inc
@@ -23,6 +23,17 @@ function template_preprocess_ding_carousel(&$vars) {
     );
   }
 
+  // This is an empty carousel, which slick don't handle well (as in it will
+  // make a infinite loop and kill the browser). So e.g. 0 hits search
+  // carousel's are fixed here be simply placing a single placeholder slide.
+  if (empty($vars['items'])) {
+    $vars['items'][] = array(
+      '#type' => 'ding_carousel_item',
+      '#content' => $carousel['#placeholder'],
+      '#placeholder' => TRUE,
+    );
+  }
+
   // Fill up with placeholders if there's not enough items.
   if (count($vars['items']) < $carousel['#placeholders']) {
     while (count($vars['items']) < $carousel['#placeholders']) {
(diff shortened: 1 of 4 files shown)


modules/p2/ding_list/include/functions.inc
```diff
@@ -818,6 +818,11 @@ function ding_list_get_permission_permissions($permission = FALSE) {
 function ding_list_user_has_access($list, $permission = DING_LIST_PERMISSION_VIEW, $account = NULL, $token = FALSE) {
   $access = &drupal_static(__FUNCTION__, array());
 
+  // If used as menu access callback the list have to be loaded.
+  if (is_numeric($list)) {
+    $list = ding_list_get_list($list);
+  }
+
```
kasperg: If we are to utilize the static cache then I would prefer to see this moved below the subsequent access check.

modules/p2/ding_list/include/functions.inc
```diff
@@ -802,8 +802,8 @@ function ding_list_get_permission_permissions($permission = FALSE) {
 /**
  * Check if a user has access to a list, with the given permissions.
  *
- * @param object $list
- *   List entity.
+ * @param mixed $list
```
kasperg: This should be `object|int`.

cableman: fixed

modules/p2/ding_list/include/functions.inc
cableman: Nice catch. Was a very fast PR. :-) Fixed now

modules/ding_carousel/ding_carousel.theme.inc
```diff
@@ -23,6 +23,17 @@ function template_preprocess_ding_carousel(&$vars) {
     );
   }
 
+  // This is an empty carousel, which slick don't handle well (as in it will
+  // make a infinite loop and kill the browser). So e.g. 0 hits search
+  // carousel's are fixed here be simply placing a single placeholder slide.
```
kasperg: :+1: Appreciate the level of commenting here.

PR #859, opened 2017-10-26
1625: Suppress warning if FBS not configured yet

xendk:


JrgenGNielsen:
reviewed and approved

PR #861, opened 2017-10-26
3026: Add name and to profile staff

KristianB14:
https://platform.dandigbib.org/issues/3026

### DIFF
--- a/themes/ddbasic/sass/components/user.scss
+++ b/themes/ddbasic/sass/components/user.scss
@@ -36,7 +36,6 @@
     }
     h1 {
       @include font('display');
-      float: left;
       margin-bottom: 0;
       color: $charcoal;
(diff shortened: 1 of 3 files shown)


themes/ddbasic/sass/components/user.scss
```diff
@@ -36,6 +36,7 @@
     }
     h1 {
       @include font('display');
+      float: left;
```
kasperg: Is it intentional that this float is added?

It does not seem to be related to the issue at hand and it is not mentioned in the commit message either.

KristianB14: No - this was a mistake and has been removed in new commit

kasperg: :+1:

PR #864, opened 2017-10-26
2671: Change tabroll to use drupal behaviors

cableman:
See https://platform.dandigbib.org/issues/2671

### DIFF
--- a/modules/ding_tabroll/js/ding_tabroll.js
+++ b/modules/ding_tabroll/js/ding_tabroll.js
@@ -4,50 +4,51 @@
  * events on the tabs.
  */
 (function ($) {
+  Drupal.behaviors.ding_tabroll = {
+    attach: function (context) {
+      var tabroll = $('.ding-tabroll', context);
+      var tabroll_select = $('.ding-tabroll-select-tabs', context);
 
-  $(document).ready(function($) {
-    var tabroll = $('.ding-tabroll');
-    var tabroll_select = $('.ding-tabroll-select-tabs');
+      // Hack to check if tab have been tab_selected, as unbind event will not work.
+      var tab_selected = false;
 
-    // Hack to check if tab have been tab_selected, as unbind event will not work.
-    var tab_selected = false;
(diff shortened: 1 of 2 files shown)


cableman:
The scrutinizer is about JS and "strict" mode, but i'am not sure if we do that in other places

kasperg:
@cableman: You should `use strict` [the same way we do in many other JavaScript files](https://github.com/ding2/ding2/blob/8b432a677fe439f4c56db951b231cae911287bbf/modules/p2/ding_list/js/ding_list_tabledrag.js#L9).

cableman:
@kasperg yes, I thought so, just wanted your take on it.

modules/ding_tabroll/js/ding_tabroll.js
```diff
@@ -4,50 +4,53 @@
...
+      var tab_selected = false;
+
+      // Check if the tabs lib is loaded before trying to call it.
+      if ($.fn.tabs) {
+        tabroll.tabs({
+          select: function(event, ui) {
+            // Update the mobile navigation drop down.
+            tabroll_select.prop('selectedIndex', ui.index);
+          }
+        }).tabs("rotate", 5000);
```
kasperg: Nitpicking but this also uses double quotes?

cableman: Fixed it :-)

modules/ding_tabroll/js/jquery-ui-tabs-rotate.js
```diff
@@ -1,4 +1,6 @@
 ;(function($){
+	'use strict';
```
kasperg: [This file is third party code](https://github.com/cmcculloh/jQuery-UI-Tabs-Rotate) and should not be modified. If anything it should be removed from the repository, downloaded using Drush Make and loaded through libraries API.

cableman: Yes, I see. Reverted the library. It's outside the scope to move it to libraries right now. :-)

PR #867, opened 2017-10-27
3034: Variable check removed from popular panel

KristianB14:
PR: https://platform.dandigbib.org/issues/3034

### DIFF
--- a/modules/p2/ding_entity_rating/plugins/content_types/popular.inc
+++ b/modules/p2/ding_entity_rating/plugins/content_types/popular.inc
@@ -16,11 +16,6 @@ $plugin = array(
  * Implements hook_content_type_render().
  */
 function ding_entity_rating_popular_content_type_render($subtype, $conf, $panel_args, $context) {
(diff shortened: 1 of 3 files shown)


modules/p2/ding_entity_rating/plugins/content_types/popular.inc
```diff
@@ -16,11 +16,6 @@ $plugin = array(
  * Implements hook_content_type_render().
  */
 function ding_entity_rating_popular_content_type_render($subtype, $conf, $panel_args, $context) {
-  // Only show on front if serendipity is enabled for the frontpage.
-  // See ding_frontpage config at /admin/config/ding/frontend.
-  if (drupal_is_front_page() && !variable_get('ding_entity_rating_popular_on_frontpage', FALSE)) {
-    return;
-  }
```
kasperg: As far as I can tell this is the only place where the variable `ding_entity_rating_popular_on_frontpage` is used.

Since it is now deleted then please clean up after it as well - remove the administrative interface and delete the variable in an update hook.

PR #868, opened 2017-10-30
2376: Hide purge notifications on empty

BirkAndMe:


JrgenGNielsen:
reviewed and approved

PR #872, opened 2017-10-31
B2823: Refactoring of header

KristianB14:
https://platform.dandigbib.org/issues/2823


themes/ddbasic/sass/components/header.scss
```diff
@@ -3,702 +3,451 @@
...
-      @include media($tablet) {
-        width: 100%;
-        position: relative;
-        z-index: 4;
-        float: left;
-        padding: 10px 5%;
-        background-color: $color-primary;
-        box-shadow: $box-shadow;
+    .form-actions {
+      input[type="submit"] {
```
kasperg: Is this nesting necessary?

themes/ddbasic/sass/components/header.scss
```diff
@@ -3,702 +3,451 @@
...
       }
-      // Search form extended
-      .search-form-extended & {
-        @include clearfix;
+
+      .form-actions {
+        width: getColumn(1);
+        float: right;
+
+        input[type="submit"] {
```
kasperg: Is this nesting necessary?

themes/ddbasic/sass/components/header.scss
```diff
@@ -3,702 +3,451 @@
...
-        box-shadow: $box-shadow;
-        background-color: $grey-light;
-        z-index: 5;
+  .pane-user-menu {
+    position: absolute;
+    top: 0;
+    z-index: 51;
+
+    // Tablet
+    @include media($tablet) {
```
kasperg: This use of comments makes little sense. The comment does not provide any value to the subsequent code. 

This pattern seems to go through the majority of the changes for media queries and other elements.

themes/ddbasic/sass/components/header.scss
```diff
@@ -3,702 +3,451 @@
...
-        top: $toolbar-height - 20px;
-      }
-      // Search Form Extended on frontpage
-      .search-form-extended.page-search &,
-      .search-form-extended.front & {
-        top: $topbar-height;
+  // Search form extended with profiles
+  .search-form-extended-with-profiles & {
+    .pane-search-form {
+      .form-item-search-block-form {
```
kasperg: Is this deep nesting necessary?

themes/ddbasic/sass/components/header.scss
```diff
@@ -3,702 +3,451 @@
...
+          margin-top: 10px;
         }
       }
-      // Search form extended front when mobile search is open
-      .search-form-extended.page-search.mobile-search-is-open &,
-      .search-form-extended.front.mobile-search-is-open & {
-        // Tablet
-        @include media($tablet) {
-          box-shadow: $box-shadow;
+      .select-wrapper {
```
kasperg: This looks like a custom class to me. The name is very generic. Is the a reason why we cannot use something like `.search-form-extended-with-profiles-select-wrapper`?

guddo:
With all the changes and the nesting it’s quite hard to identify wether classnames are being renamed - if that is the case, is the business aware that this potentially makes, the +40 libraries that are using, all the custom CSS invalid?

themes/ddbasic/sass/components/header.scss
```diff
@@ -3,702 +3,451 @@
...
+      }
+    }
+  }
+
+  // Search form extended
+  .search-form-extended & {
+    @include transition(height $speed $ease);
+    height: 0;
+
+    .pane-search-form {
```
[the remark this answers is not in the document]
KristianB14: .search-form-extended is a state class. This has been changed to .search-form-is-extended

themes/ddbasic/sass/components/header.scss
KristianB14: All select lists are wrapped in a div with the class .select-wrapper. This is done for general styling. Here we need to overwrite a few of these general stylings.

themes/ddbasic/sass/components/header.scss
```diff
@@ -3,702 +3,451 @@
...
+    @include media($tablet) {
+      @include transition(transform $speed $ease);
+      @include transform(translateY(-100%));
+      display: block;
+      width: 100%;
+      max-height: 100vh;
+      padding-bottom: $header-height;
+      overflow: scroll;
+
+      .mobile-usermenu-is-open & {
```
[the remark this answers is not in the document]
KristianB14: Yes, we need this nesting. Pane-user-menu is used more than one place, so we need the parent selector .site-header-search-and-user-menu.

PR #877, opened 2017-11-01
2927: Fix double admin bar by deciding to go only with admin_menu

holt83:
https://platform.dandigbib.org/issues/2927


holt83:
@kasperg I have now rebased the changes.

I will try to provide better context going forward :)

PR #878, opened 2017-11-02
B2986

BirkAndMe:
https://platform.dandigbib.org/issues/2986

ding_message.js has changed line endings, so everything looks altered, but its only line 47 and 49 that's been changed.


JrgenGNielsen:
reviewed and approved

PR #879, opened 2017-11-02
3039: Format changed for files field

KristianB14:
https://platform.dandigbib.org/issues/3039


JrgenGNielsen:
reviewed and approved

PR #882, opened 2017-11-08
3061: Fix loan lists cache

BirkAndMe:
https://platform.dandigbib.org/issues/3061


cableman:
Reviewed and approved

PR #884, opened 2017-11-10
3060: Set toggle_name

BirkAndMe:
https://platform.dandigbib.org/issues/3060


JrgenGNielsen:
reviewed and approved

PR #885, opened 2017-11-13
3019: Sorting holdsing based on library placement

cableman:
See https://platform.dandigbib.org/issues/3019


cableman:
Rewritten the function documentation and fixed scrutinizer errors.

PR #889, opened 2017-11-16
3073: FBS session cache

cableman:
Added ding_session_cache to FBS as in the old providers. Without this FBS var called 9-12 times to show the user profile page and on every page load after login (6-9 times for logged in users to build header menu).

See https://platform.dandigbib.org/issues/3073

### DIFF
--- a/modules/ding_loan/ding_loan.module
+++ b/modules/ding_loan/ding_loan.module
@@ -50,6 +50,6 @@ function ding_loan_ding_session_cache_defaults() {
   return array(
     'title' => 'Ding loans',
     'enabled' => TRUE,
-    'expire' => 3600,
+    'expire' => 60,
   );
 }
--- a/modules/ding_loan/ding_loan.module
+++ b/modules/ding_loan/ding_loan.module
@@ -48,7 +48,7 @@ function ding_loan_sort_expiry($a, $b) {
  */
 function ding_loan_ding_session_cache_defaults() {
   return array(
-    'titel' => 'Ding loans',
+    'title' => 'Ding loans',
     'enabled' => TRUE,
     'expire' => 3600,
   );
--- a/modules/ding_loan/ding_loan.module
+++ b/modules/ding_loan/ding_loan.module
@@ -50,6 +50,6 @@ function ding_loan_ding_session_cache_defaults() {
   return array(
     'title' => 'Ding loans',
     'enabled' => TRUE,
-    'expire' => 60,
+    'expire' => 3600,
   );
 }
--- a/modules/ding_session_cache/ding_session_cache.module
+++ b/modules/ding_session_cache/ding_session_cache.module
@@ -86,18 +86,21 @@ function ding_session_cache_set($module, $cid, $data) {
  *   Name of the module that the date is stored under (prefix).
  * @param string $cid
  *   Index data is stored under.
+ * @param bool $default
+ *   Default value if cache for the $cid is empty.
  *
  * @return mixed
  *   If data found it's returned else FALSE.
  */
-function ding_session_cache_get($module, $cid) {
+function ding_session_cache_get($module, $cid, $default = FALSE) {
   if (_ding_session_cache_enabled($module)) {
     $cache = cache_get(_ding_session_cache_get_cid($module, $cid), 'cache_ding_session_cache');
     if (isset($cache->expire) && $cache->expire > REQUEST_TIME) {
       return $cache->data;
     }
   }
-  return FALSE;
+
+  return $default;
 }
 
 /**
--- a/modules/ding_session_cache/ding_session_cache.module
+++ b/modules/ding_session_cache/ding_session_cache.module
@@ -38,7 +38,7 @@ function ding_session_cache_form_system_performance_settings_alter(&$form, &$for
       $result = $function();
       $form['ding_session_cache'][$module] = array(
         '#type' => 'fieldset',
-        '#title' => $result['titel'],
+        '#title' => $result['title'],
         'enabled' => array(
           '#type' => 'checkbox',
           '#title' => 'Enabled',
@@ -47,7 +47,7 @@ function ding_session_cache_form_system_performance_settings_alter(&$form, &$for
         'expire' => array(
           '#type' => 'select',
           '#title' => t('Maximum cache lifetime'),
-          '#default_value' => isset($defaults[$module]['expire']) ? $defaults[$module]['expire'] : $result['enabled'],
+          '#default_value' => isset($defaults[$module]['expire']) ? $defaults[$module]['expire'] : $result['expire'],
           '#options' => $period,
         ),
       );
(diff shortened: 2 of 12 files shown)


modules/ding_session_cache/ding_session_cache.module
```diff
@@ -38,7 +38,7 @@ function ding_session_cache_form_system_performance_settings_alter(&$form, &$for
       $result = $function();
       $form['ding_session_cache'][$module] = array(
         '#type' => 'fieldset',
-        '#title' => $result['titel'],
+        '#title' => $result['title'],
```
kasperg: :+1:

modules/ding_session_cache/ding_session_cache.module
```diff
@@ -47,7 +47,7 @@ function ding_session_cache_form_system_performance_settings_alter(&$form, &$for
         'expire' => array(
           '#type' => 'select',
           '#title' => t('Maximum cache lifetime'),
-          '#default_value' => isset($defaults[$module]['expire']) ? $defaults[$module]['expire'] : $result['enabled'],
+          '#default_value' => isset($defaults[$module]['expire']) ? $defaults[$module]['expire'] : $result['expire'],
```
kasperg: :+1:

modules/ding_session_cache/ding_session_cache.module
kasperg: :+1:

modules/fbs/includes/fbs.debt.inc
```diff
@@ -10,71 +10,89 @@
...
-    if (!empty($fee->materials)) {
-      $record_ids = array();
-      $material_ids = array();
-      foreach ($fee->materials as $material) {
-        $material_ids[] = $material->materialItemNumber;
-        $record_ids[] = $material->recordId;
+
+    // If both static cache and session cache failed, try reloading information
+    // from FBS.
+    if (empty($results) || $reset) {
```
kasperg: Does this introduce a problem about distinguishing between an empty cache and an empty debt list?

If the user does not have any debts then the cached result will be an empty array. Here this will be interpreted as an empty cache and thus result in debts being fetched again.

One way to approach this would be to use `NULL` as the default value for empty caches and checking it using `isset()`.

modules/fbs/includes/fbs.loan.inc
```diff
@@ -10,61 +10,82 @@
...
-
-    // If this is a periodical, add in issue data.
-    if (isset($loan->loanDetails->periodical)) {
-      $periodical = $loan->loanDetails->periodical;
-      $vol = $periodical->volume;
-      if (!empty($periodical->volumeNumber)) {
-        $vol .= '.' . $periodical->volumeNumber;
+    // If both static cache and session cache failed, try reloading information
+    // from FBS.
+    if (empty($results) || $reset) {
```
kasperg: See comment above regarding handling of empty results from FBS.

modules/fbs/includes/fbs.reservation.inc
```diff
@@ -12,95 +12,116 @@
...
 
-    // Add in periodical information if it is one.
-    if ($reservation->periodical) {
-      $periodical = $reservation->periodical;
-      $vol = $periodical->volume;
-      if (!empty($periodical->volumeNumber)) {
-        $vol .= '.' . $periodical->volumeNumber;
+    // If both static cache and session cache failed, try reloading information
+    // from FBS.
+    if (empty($results) || $reset) {
```
kasperg: As above.

modules/fbs/includes/fbs.reservation.inc
```diff
@@ -189,6 +215,9 @@ function fbs_reservation_create($account, $record_id, $options) {
  *   Reservations to update.
  * @param array $options
  *   Update options.
+ *
+ * @return bool
+ *   TRUE if no errors else FALSE.
```
kasperg: :+1:

modules/ding_session_cache/ding_session_cache.module
```diff
@@ -86,18 +86,21 @@ function ding_session_cache_set($module, $cid, $data) {
  *   Name of the module that the date is stored under (prefix).
  * @param string $cid
  *   Index data is stored under.
+ * @param bool $default
+ *   Default value if cache for the $cid is empty.
  *
  * @return mixed
  *   If data found it's returned else FALSE.
```
kasperg: This comment should be updated to reflect the added default value.

modules/fbs/fbs.info
```diff
@@ -7,6 +7,7 @@ project = fbs
 dependencies[] = date_popup
 dependencies[] = ding_provider
 dependencies[] = ding_reservation
+dependencies[] = ding_session_cache
```
kasperg: Since we add `ding_session_cache` as a dependency we also need to enable it for existing installations with a `hook_update()`.

modules/ding_loan/ding_loan.module
```diff
@@ -48,8 +48,8 @@ function ding_loan_sort_expiry($a, $b) {
  */
 function ding_loan_ding_session_cache_defaults() {
   return array(
-    'titel' => 'Ding loans',
+    'title' => 'Ding loans',
     'enabled' => TRUE,
-    'expire' => 3600,
+    'expire' => 60,
```
kasperg: This reduces the loan cache from an hour to a minute. That seems pretty drastic - and also unrelated to the change at hand. Why is this?

modules/fbs/fbs.info
cableman: It's already an dependency in ding2.info, so it should be enabled.

> ding2.info:36:dependencies[] = ding_session_cache

modules/fbs/includes/fbs.reservation.inc
cableman: Changed default value to NULL and using is_null to check it.

modules/fbs/includes/fbs.loan.inc
cableman: Changed default value to NULL and using is_null to check it.

modules/fbs/includes/fbs.debt.inc
cableman: This may give problems, so changed default value to NULL and using is_null to check it.

modules/ding_session_cache/ding_session_cache.module
cableman: Right you are :-) I have updated the documentation for return as well.

modules/ding_session_cache/ding_session_cache.module
cableman: Reverted

cableman:
Rebased onto master to make it merge-able.

modules/fbs/fbs.info
kasperg: :+1:

PR #890, opened 2017-11-16
3022: Ajax load user menu to prevent caching in varnish

cableman:
See https://platform.dandigbib.org/issues/3022


themes/ddbasic/sass/components/menu.scss
```diff
@@ -559,3 +559,43 @@ ul.topbar-menu {
     }
   }
 }
+
+// ==========================================================================
+// User mobile menu placeholder - lazy loading
+// ==========================================================================
+.mobile-user-menu .lazy-pane-placeholder {
```
kasperg: This seems to set a new spinner specifically for this element. Is that really necessary?

themes/ddbasic/sass/components/menu.scss
```diff
@@ -559,3 +559,43 @@ ul.topbar-menu {
...
+    left: 0;
+    width: 100%;
+    height: 4px;
+    content: "";
+  }
+
+  .mobile-usermenu-is-open & {
+    display: block;
+  }
+}
```
kasperg: Missing newline.

themes/ddbasic/templates/menu/panels-pane--sub-menu--mobile.tpl.php
```diff
@@ -20,11 +20,12 @@
...
-</section>
+<?php if ($is_lazy_pane_render) { ?>
+  <?php print render($content); ?>
+<?php } else { ?>
+  <section class="navigation-wrapper mobile-user-menu <?php print $classes; ?>">
+    <div class="navigation-inner">
+      <?php print render($content); ?>
+    </div>
+  </section>
+<?php } ?>
```
kasperg: This seems to remove `$admin_links` entirely from the renderend template. Is that intentional? I could not tell from the commit comment.

themes/ddbasic/sass/components/menu.scss
cableman: It's due to the way these spinners is constructed. I did not see any other way to get the spinner here without having to refactor the SASS. Which would be outside the scope of this issue.

themes/ddbasic/sass/components/menu.scss
cableman: Fixed

themes/ddbasic/templates/menu/panels-pane--sub-menu--mobile.tpl.php
cableman: I don't think they would every be set here, but re-add theme to the template to be sure.

PR #891, opened 2017-11-20
3078: Correct title when renewing multiple loans.

tripox:


JrgenGNielsen:
reviewed and approved

PR #896, opened 2017-11-22
3082: Added more link to events in groups

cableman:
See https://platform.dandigbib.org/issues/3082

### DIFF
--- a/modules/ding_event/ding_event.views_default.inc
+++ b/modules/ding_event/ding_event.views_default.inc
@@ -849,12 +849,13 @@ function ding_event_views_default_views() {
   $handler->display->display_options['defaults']['title'] = FALSE;
   $handler->display->display_options['title'] = 'Group events';
   $handler->display->display_options['defaults']['use_more'] = FALSE;
+  $handler->display->display_options['use_more'] = TRUE;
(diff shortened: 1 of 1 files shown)


modules/ding_event/ding_event.views_default.inc
```diff
@@ -849,12 +849,13 @@ function ding_event_views_default_views() {
...
   $handler->display->display_options['defaults']['use_more'] = FALSE;
+  $handler->display->display_options['use_more'] = TRUE;
   $handler->display->display_options['defaults']['link_display'] = FALSE;
   $handler->display->display_options['link_display'] = 'custom_url';
-  $handler->display->display_options['link_url'] = 'arrangementer';
+  $handler->display->display_options['link_url'] = 'temaer/!1/arrangementer';
   $handler->display->display_options['defaults']['pager'] = FALSE;
   $handler->display->display_options['pager']['type'] = 'some';
-  $handler->display->display_options['pager']['options']['items_per_page'] = '5';
+  $handler->display->display_options['pager']['options']['items_per_page'] = '6';
```
kasperg: This brings the number of events to show in line with other views. It is not related to the change at hand. I would appreciate it  is this at least was in a separate commit.

PR #902, opened 2017-11-24
B2480

BirkAndMe:
https://platform.dandigbib.org/issues/2480


modules/p2/ding_list/include/form.inc
```diff
@@ -741,3 +741,52 @@ function ding_list_list_permissions_form_share_callback($form, &$form_state) {
...
+    ),
+    'cancel' => array(
+      '#type' => 'submit',
+      '#value' => t('Cancel'),
+      '#submit' => array('ding_list_reset_user_form_submit_cancel'),
+      '#limit_validation_errors' => array(),
+    ),
+  );
+
+  return $form;
```
kasperg: This form should be replaced by [`confirm_form`](https://api.drupal.org/api/drupal/modules%21system%21system.module/function/confirm_form/7.x) unless there is a specific reason to implement this ourselves.

modules/p2/ding_list/include/form.inc
```diff
@@ -741,3 +741,52 @@ function ding_list_list_permissions_form_share_callback($form, &$form_state) {
...
+ */
+function ding_list_reset_user_form_submit($form, &$form_state) {
+  $account = $form_state['build_info']['args'][0];
+
+  ding_list_reset_user($account);
+
+  ding_list_setup_autocreated($account);
+  if (ding_user_consent_has_consent('loan_history_store')) {
+    ding_list_update_historical_loans($account);
+  }
```
kasperg: Is there a reason why these four lines are not part of `ding_list_reset_user()`?

If we at any other point need to reset a user it would be nice to have a single function to achieve that. One might think that `ding_list_reset_user()` would be sufficient but that is not the case.

modules/p2/ding_list/ding_list_installer/ding_list_installer.pages_default.inc
```diff
@@ -8,6 +8,105 @@
...
 function ding_list_installer_default_page_manager_pages() {
+  $page = new stdClass();
+  $page->disabled = FALSE; /* Edit this to true to make a default page disabled initially */
+  $page->api_version = 1;
+  $page->name = 'ding_list_reset_user';
+  $page->task = 'page';
+  $page->admin_title = 'Ding list reset user';
+  $page->admin_description = '';
+  $page->path = 'user/%user/reset-lists';
+  $page->access = array();
```
kasperg: This seems to warrant some form of access control. As I read the code in the current state then I would be able to reset the lists for any user if I know their user id. That is not good.

There might be permissions for deleting own and others lists further down the line but just being able to see the form here, might be enough to scare some people.

themes/ddbasic/sass/base/form.scss
```diff
@@ -341,6 +341,11 @@ select[multiple="multiple"] {
 
 .form-actions {
   margin: 0;
+
+  input[type='submit'] {
+    display: inline-block;
+    margin-right: 20px;
+  }
```
kasperg: This seems to affect submit buttons on all form actions. Is that intentional?

guddo:
Do you mind changing your the title of this PR?
It's quite annoying not being able to see what the PR is about from the overview.

PR #906, opened 2017-11-28
3091: Fixed 16/9 format on E-resource

b14sha:
https://platform.dandigbib.org/issues/3091


cableman:
Replaced by https://github.com/ding2/ding2/pull/908

PR #907, opened 2017-11-28
3092: Added list image to viewmode no-overlay

b14sha:
https://platform.dandigbib.org/issues/3092


DDBdrift:
https://platform.dandigbib.org/issues/3092 | Unable to reproduce (Closed)

PR #910, opened 2017-11-28
3099: Openlist call module fix

BirkAndMe:
https://platform.dandigbib.org/issues/3099


cableman:
Reviewed and awaiting release

cableman:
Closed because it have been fixed: https://github.com/ding2/ding2/blob/master/modules/p2/ting_openlist/ting_openlist.handler.inc#L242

PR #911, opened 2017-11-29
3103: Allow all image formats on library teaser

b14sha:
https://platform.dandigbib.org/issues/3103


cableman:
Reviewed and awaiting release

PR #912, opened 2017-11-30
3108: Add Slick as a DDBasic library dependency

kasperg:
It is used by [views.js](https://github.com/ding2/ding2/blob/master/themes/ddbasic/scripts/view.js#L282-L321) so we should define and load it as well.

Add Slick as a library and load it for all pages. Loading could probably
be made more specific.


JrgenGNielsen:
reviewed and approved

cableman:
Slick should only be defined in one place: https://github.com/ding2/ding2/blob/master/modules/ding_carousel/ding_carousel.module#L75

Which i think is in ding_carousel.

PR #914, opened 2017-11-30
3063: Overdue loans translation

BirkAndMe:
https://platform.dandigbib.org/issues/3063


JrgenGNielsen:
reviewed and approved

cableman:
The changed file is a translation file from D.O an will be overwritten be l10n_update module.

Also all translations are handle by a group in DDB, so all translations an updates of these are added be the core team at release time.

PR #915, opened 2017-12-03
629: Fix translations of search backends

holt83:
https://platform.dandigbib.org/issues/629

### DIFF
--- a/modules/ting_search/plugins/content_types/search_backends.inc
+++ b/modules/ting_search/plugins/content_types/search_backends.inc
@@ -51,7 +51,7 @@ function ting_search_backend_engines_content_type_render($subtype, $conf, $panel
 
     // Create the option as a link.
     $txt = format_string('Search @backend', array('@backend' => $backend['title']));
-    $title = l($txt, $path, array('query' => $conditions));
+    $title = l(t($txt), $path, array('query' => $conditions));
 
     // Getting the count of search results for each enabled backend.
     switch ($backend['path']) {
--- a/modules/ting_search/plugins/content_types/search_backends.inc
+++ b/modules/ting_search/plugins/content_types/search_backends.inc
@@ -51,7 +51,7 @@ function ting_search_backend_engines_content_type_render($subtype, $conf, $panel
 
     // Create the option as a link.
     $txt = format_string('Search @backend', array('@backend' => $backend['title']));
-    $title = l(t('@txt', array('@txt' => $txt)), $path, array('query' => $conditions));
+    $title = l($txt, $path, array('query' => $conditions));
 
     // Getting the count of search results for each enabled backend.
     switch ($backend['path']) {


modules/ting_search/plugins/content_types/search_backends.inc
```diff
@@ -51,7 +51,7 @@ function ting_search_backend_engines_content_type_render($subtype, $conf, $panel
 
     // Create the option as a link.
     $txt = format_string('Search @backend', array('@backend' => $backend['title']));
-    $title = l(t('@txt', array('@txt' => $txt)), $path, array('query' => $conditions));
+    $title = l($txt, $path, array('query' => $conditions));
```
JrgenGNielsen: Shouldn't this be:
$txt = t("Search @backend", array('@backend' => $backend['title']));
$title = l($txt, $path, array('query' => $conditions));

holt83: You're right.. there's something wrong.

However, I changed it to:
$txt = format_string('Search @backend', array('@backend' => $backend['title']));
$title = l(t($txt), $path, array('query' => $conditions));

Doing it the suggested way, will prevent us from translating the backend['title'], which is returned untranslated from search_get_info(): https://api.drupal.org/api/drupal/modules%21search%21search.api.php/function/hook_search_info/7.x
We can only translate "Search @backend" in this case.

We could pass backend['title'] through a separate call to t(), but I suggest doing it this way, since it will work with the existing translations (it was done this way before).

modules/ting_search/plugins/content_types/search_backends.inc
```diff
@@ -51,7 +51,7 @@ function ting_search_backend_engines_content_type_render($subtype, $conf, $panel
 
     // Create the option as a link.
     $txt = format_string('Search @backend', array('@backend' => $backend['title']));
-    $title = l($txt, $path, array('query' => $conditions));
+    $title = l(t($txt), $path, array('query' => $conditions));
```
JrgenGNielsen: Scrutinizer pukes on t($txt). The Drupal documentation on t() also warns against it - although format_string() fix that.
However, I think the issue lies elsewhere... 
At installation,  "Search @backend" will evaluate to "Search ".
"Search Materialer" and "Search Content" won't be added to the database of strings to be translated until the code is run. 
I suggest we add the strings to the database of strings to be translated, at installation.
So, add t("Search Materialer") and t("Search Content") to hook_install() (and a hook_update() for good measure)

holt83: But the"Search Materialer" and "Search Content" is already in da.po translation file, so wont they get added during installation and update? Do they really need to be in the database first to be imported?

JrgenGNielsen: hmm... I think you're right. I can't find a good way to get around translating variables.

JrgenGNielsen:
reviewed and approved

modules/ting_search/plugins/content_types/search_backends.inc
holt83: This has annoyed me as well several times before. I think that maybe we have misused the translation system a bit in DDB CMS.

PR #918, opened 2017-12-06
3098 by Inlead: Added patch to oembed module.

guddo:
https://platform.dandigbib.org/issues/3098


cableman:
Reviewed and awaiting release

PR #920, opened 2017-12-07
B2875v2

BirkAndMe:
https://platform.dandigbib.org/issues/2875

### DIFF
--- a/modules/ding_base/ding_base.module
+++ b/modules/ding_base/ding_base.module
@@ -685,3 +685,39 @@ function ding_base_lazy_pane_ajax_deliver($page_callback_result) {
 
   return ajax_deliver($page_callback_result);
 }
+
+/**
+ * Validate an email.
+ *
+ * This check uses Drupals valid_email_addess(), and makes sure the domain used
+ * is a valid top level domain.
+ *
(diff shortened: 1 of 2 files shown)


modules/ding_base/ding_base.module
```diff
@@ -685,3 +685,39 @@ function ding_base_lazy_pane_ajax_deliver($page_callback_result) {
...
+    // If the response fails the $tld_list will be empty and we'll simply skip
+    // the domain check.
+    $response = drupal_http_request('http://data.iana.org/TLD/tlds-alpha-by-domain.txt');
+    if ($response->code == 200) {
+      $tld_list = explode("\n", $response->data);
+      cache_set('ding_base:tld_list', $tld_list);
+    }
+  }
+
+  $domain = strtoupper(substr($email, strrpos($email, '.') + 1));
```
cableman: If we look at https://tools.ietf.org/html/rfc5322 an valid mail could be "<cableman@test.dk>" with "<>". So as this is placed at a generic checker in ding_base it might take this into account.

cableman:
Single comment and please fix the scrutinizer notifications.

PR #922, opened 2017-12-07
2331: Hide follow button on own lists

BirkAndMe:
https://platform.dandigbib.org/issues/2331


cableman:
Reviewed and awaiting release

PR #923, opened 2017-12-08
2692: Updated memcache module

BirkAndMe:
https://platform.dandigbib.org/issues/2692

The change on line 231 is a linebreak fix not related to the issue.


cableman:
Reviewed and awaiting release

PR #928, opened 2017-12-12
3076: ding webtrekk fixes

holt83:
https://platform.dandigbib.org/issues/3076

OSS parameters was added twice because of the changes in https://platform.dandigbib.org/issues/629.

Updated the page titles also. Before the page title for Search node, was ding_multiple_search.

Also includes the fix to track empty searches.

### DIFF
--- a/modules/ding_webtrekk/ding_webtrekk.module
+++ b/modules/ding_webtrekk/ding_webtrekk.module
@@ -65,7 +65,7 @@ function ding_webtrekk_panels_pane_content_alter($content, $pane, $args, $contex
   if ($pane->type == 'search_result' && $pane->configuration['type'] == 'ting_search' && strpos(current_path(), 'search/ting') === 0) {
     if ($search_result = drupal_static('ting_search_results', FALSE)) {
       $params = [
-        'pageTitle' => $content->original_title,
+        'pageTitle' => 'Search ting',
         'OSS' => $search_result->search_key,
         'OSSr' => $search_result->numTotalObjects,
       ];
@@ -91,7 +91,7 @@ function ding_webtrekk_views_post_execute(&$view) {
     $params = [
       'internalSearch' => array_shift($view->args),
       'numberSearchResults' => $view->total_rows,
-      'pageTitle' => $view->human_name,
+      'pageTitle' => 'Search content',
     ];
     _ding_webtrekk_set_data_layer_values($params);
   }
--- a/modules/ding_webtrekk/ding_webtrekk.module
+++ b/modules/ding_webtrekk/ding_webtrekk.module
@@ -17,7 +17,7 @@ function ding_webtrekk_menu() {
     'access arguments' => ['administer site configuration'],
     'file' => 'ding_webtrekk.admin.inc',
   ];
-  
+
   return $items;
 }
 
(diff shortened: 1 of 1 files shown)


holt83:
Regarding the "OSS parameters was added twice":

modules/ding_webtrekk/ding_webtrekk.module
```diff
@@ -80,8 +80,7 @@ function ding_webtrekk_panels_pane_content_alter($content, $pane, $args, $contex
  * Track internal search results.
  */
 function ding_webtrekk_views_post_execute(&$view) {
-  $count = $view->total_rows;
-  if ($view->name == 'ding_multiple_search' && $count > 0) {
+  if ($view->name == 'ding_multiple_search' && strpos(current_path(), 'search/node') === 0) {
```
kasperg: Why do we need the current path check here?

modules/ding_webtrekk/ding_webtrekk.module
```diff
@@ -62,10 +62,10 @@ function ding_webtrekk_page_alter(&$page) {
  * Track ting search results.
  */
 function ding_webtrekk_panels_pane_content_alter($content, $pane, $args, $contexts) {
-  if ($pane->type == 'search_result' && $pane->configuration['type'] == 'ting_search') {
+  if ($pane->type == 'search_result' && $pane->configuration['type'] == 'ting_search' && strpos(current_path(), 'search/ting') === 0) {
```
kasperg: The combination of the implemented hook and the dependency on the current path does not seem right. The reliance on panes seems to be unnecessary noise.

How about using another hook instead and just checking whether `drupal_static('ting_search_results', FALSE)` is set? Perhaps `hook_page_alter` is really the way to go. Another approach could be `hook_search_page`

modules/ding_webtrekk/ding_webtrekk.module
holt83: Because else it's added on ting search page also because of https://platform.dandigbib.org/issues/629

modules/ding_webtrekk/ding_webtrekk.module
holt83: I don't know, this was the way it was done in the old module. Just wanted to ensure that it only gets added on the correct page.

holt83: If I remember correctly it was done this way because of:

holt83: The static variable drupal_static('ting_search_results', FALSE) is set in ting_search_search_execute. So maybe it's safe to check for it in hook_page_alter. 

Should I make this update and remove hook_panels_pane_content_alter() all together?

Just have the feeling that I forgotten something and it was made this way for a reason (made it myself, but can't remember tbh)

holt83: Looked into hook_search_page. It seems it's only invoked on the module implementing the search type?

kasperg: @holt83: Then I think we should switch to `hook_page_alter` and check current path. `hook_panels_pane_content_alter()` can be removed entirely.

modules/ding_webtrekk/ding_webtrekk.module
kasperg: OK. Then I would appreciate a comment explaining this.

holt83: Oh yeah sorry, of course. Coming up :)

modules/ding_webtrekk/ding_webtrekk.module
holt83: Ok. I will update PR with this approach

holt83:
@kasperg I updated the PR with the suggested changes.

I tried my best to make helpful comments. Let me know if they don't make sense or if you have suggestions for improvements :)

kasperg:
Approved - but branched from wrong `master` instead of `release-4.1.x`. I cherry picked commits manually.

PR #930, opened 2017-12-13
2376: Clear messages and redirection

BirkAndMe:


cableman:
Reviewed and awaiting release

PR #935, opened 2017-12-19
3159: Fixed opening hours subqueue

BirkAndMe:
https://platform.dandigbib.org/issues/3159

### DIFF
--- a/modules/ding_ddbasic/modules/ding_ddbasic_opening_hours/ding_ddbasic_opening_hours.theme.inc
+++ b/modules/ding_ddbasic/modules/ding_ddbasic_opening_hours/ding_ddbasic_opening_hours.theme.inc
@@ -188,6 +188,9 @@ function template_preprocess_ding_ddbasic_opening_hours_all(&$variables) {
 
   $order = array();
   $queue = nodequeue_load_queue_by_name('ding_library_listing');
+  if ($queue->subqueues < 1) {
+    return;
+  }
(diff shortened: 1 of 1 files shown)


modules/ding_ddbasic/modules/ding_ddbasic_opening_hours/ding_ddbasic_opening_hours.theme.inc
```diff
@@ -188,7 +188,8 @@ function template_preprocess_ding_ddbasic_opening_hours_all(&$variables) {
 
   $order = array();
   $queue = nodequeue_load_queue_by_name('ding_library_listing');
-  foreach (nodequeue_load_nodes($queue->qid, FALSE, 0, 0) as $node) {
+  $subqueue = current(nodequeue_load_subqueues_by_queue($queue->qid));
+  foreach (nodequeue_load_nodes($subqueue->sqid, FALSE, 0, 0) as $node) {
```
kasperg: [`nodequeue_load_subqueues_by_queue`](http://www.drupalcontrib.org/api/drupal/contributions%21nodequeue%21nodequeue.module/function/nodequeue_load_subqueues_by_queue/7) can return an empty array. In that case `$subqueue` will be `FALSE` and the code will fail again.

We should either loop over all subqueues or ensure that we have at least one subqueue before trying to extract the first if there is a specific reason to go for that one.

kasperg:
@BirkAndMe You opted for a different approach compared to what I suggested. I wonder why.

I updated the code and merged this manually.

PR #936, opened 2017-12-19
3115: update code to use ortograf.dbc.dk

JrgenGNielsen:
See https://platform.dandigbib.org/issues/3115

### DIFF
--- a/modules/ding_ting_frontend/ding_ting_frontend.info
+++ b/modules/ding_ting_frontend/ding_ting_frontend.info
@@ -25,7 +25,6 @@ dependencies[] = ting_proxy
 dependencies[] = ting_reference
 dependencies[] = ting_relation
 dependencies[] = ting_search
-dependencies[] = ting_search_autocomplete
 dependencies[] = uuid
 dependencies[] = virtual_field
 features[ctools][] = field_group:field_group:1
--- a/modules/ting_search_ortograf/includes/ting_search_ortograf.admin.inc
+++ b/modules/ting_search_ortograf/includes/ting_search_ortograf.admin.inc
@@ -1,0 +1,68 @@
+<?php
+
+/**
+ * @file
+ * Admin settings for ting search ortograf autocomplete.
+ */
+
+/**
+ * Ting Search Ortograf admin form
+ */
+function ting_search_ortograf_admin() {
+
+  $form['ting_search_ortograf_webservice'] = array(
+    '#type' => 'fieldset',
+    '#title' => t('Autocomplete URL', array(), array('context' => 'ting_search_ortograf')),
+    '#tree' => FALSE,
+    '#collapsible' => TRUE,
+  ); 
+
+  $form['ting_search_ortograf_webservice']['ting_search_ortograf_suggestion_url'] = array(
+    '#type' => 'textfield',
+    '#title' => t('Autocomplete URL', array(), array('context' => 'ting_search_ortograf')),
+    '#description' => t('URL to Autocomplete webservice. F.ex.: https://ortograf.dbc.dk/ortograf/suggest', array(), array('context' => 'ting_search_ortograf')),
+    '#default_value' => variable_get('ting_search_ortograf_suggestion_url', NULL),
+    '#required' => TRUE,
+  );
+
+  $form['ting_search_ortograf_settings'] = array(
+    '#type' => 'fieldset',
+    '#title' => t('Autocomplete settings', array(), array('context' => 'ting_search_ortograf')),
+    '#tree' => FALSE,
+    '#collapsible' => TRUE,
+  );
+
+  $form['ting_search_ortograf_settings']['ting_search_ortograf_rows'] = array(
+    '#type' => 'textfield',
+    '#title' => t('Max suggestions', array(), array('context' => 'ting_search_ortograf')),
+    '#default_value' => variable_get('ting_search_ortograf_rows', 10),
+    '#element_validate' => array('element_validate_integer_positive'), 
+    '#description' => t('Maximum suggestions to return. Default 10', array(), array('context' => 'ting_search_ortograf')),
+  );
+
+  $form['ting_search_ortograf_settings']['ting_search_ortograf_minimum_string'] = array(
+    '#type' => 'textfield',
+    '#title' => t('Minimum string', array(), array('context' => 'ting_search_ortograf')),
+    '#description' => t('Minimum characters the user has to type before the suggestionservice is queried. Default 3', array(), array('context' => 'ting_search_ortograf')),
+    '#default_value' => variable_get('ting_search_ortograf_minimum_string', 3),
+    '#element_validate' => array('element_validate_integer_positive'), 
+  );
+
+  return system_settings_form($form);
+}
+
+
+/**
+ * Validation handler for ting_search_ortograf_admin
+ */
+function ting_search_ortograf_admin_validate($form, &$form_state) {
+  $form_state['values']['ting_search_ortograf_suggestion_url'] = trim($form_state['values']['ting_search_ortograf_suggestion_url']);
+  if (!filter_var($form_state['values']['ting_search_ortograf_suggestion_url'], FILTER_VALIDATE_URL)) {
+    form_set_error(
+      'ting_search_ortograf_suggestion_url',
+      t('Invalid URL: %url',
+        array('%url' => $form_state['values']['ting_search_ortograf_suggestion_url']),
+        array('context' => 'ting_search_ortograf'))
+      );
+  }
+}
--- a/modules/ting_search_ortograf/includes/ting_search_ortograf.ajax.inc
+++ b/modules/ting_search_ortograf/includes/ting_search_ortograf.ajax.inc
@@ -1,0 +1,92 @@
+<?php
+/**
+ * @file
+ * AJAX functions ting search ortograf autocomplete.
+ */
+
+/**
+ * Auto-complete callback for the ting search block.
+ *
+ * The function returns JSON and ending execution.
+ *
+ * @param string $string
+ *   The string to search suggestion with.
+ */
+function ting_search_ortograf_callback($string = '') {
+  $matches = array();
+
+  $query['q'] = check_plain(strtolower($string));
+  $query['type'] = 'all';
+  $query['rows'] = variable_get('ting_search_ortograf_rows', 10);
+
+  $minimum_string = variable_get('ting_search_ortograf_minimum_string', 3);
+  if (drupal_strlen($string) < $minimum_string) {
+    drupal_json_output(array());
+    drupal_exit();
+  }
+
+  // Try to use cache for the search string.
+  $cid = md5(serialize($query));
+  if ($cache = cache_get($cid, 'cache_ting_search_ortograf')) {
+    $matches = $cache->data;
+  }
+  else {
+    $suggestions = ting_search_ortograf_get_suggestions($query);
+    // Maximum space in suggestions dropdown is 62 characters.
+    // Maximum characters for search query string is 256 characters.
+    foreach ($suggestions as $suggestion) {
+      $val = truncate_utf8($suggestion, 62, TRUE, TRUE, 1);
+      $key = truncate_utf8($suggestion, 256, TRUE, FALSE, 1);
+      $matches[$key] = $val;
+    }
+    // Store the matches in cache for faster lookup next time.
+    cache_set($cid, $matches, 'cache_ting_search_ortograf', CACHE_TEMPORARY);
+  }
+
+  drupal_json_output($matches);
+  drupal_exit();
+}
+
+
+/**
+ * SuggestionService - retrieve suggestions with drupal_http_request() 
+ *
+ * @param array $query
+ *   Search suggestion parameters.
+ *
+ * @return array
+ *   Search suggestion results.
+ */
+function ting_search_ortograf_get_suggestions(array $query) {
+  $items = array();
+  $rest_url = variable_get('ting_search_ortograf_suggestion_url', '');
+  $url = url($rest_url, array('query' => $query));
+  $headers = array('Accept' => 'application/json');
+
+  $result = drupal_http_request($url, $headers);
+
+  if (!empty($result->error)) {
+    watchdog(
+      'ting_search_ortograf',
+      "Autocomplete error: %error (%code)\nURL: !url",
+      array(
+        '%error' => $result->error,
+        '%code' => $result->code,
+        '!url' => $url,
+      ),
+      $severity = WATCHDOG_ERROR
+    );
+    return $items;
+  }
+
+  if (!empty($result->data)) {
+    $result_data = drupal_json_decode($result->data);
+    if (!empty($result_data['response']['docs'])) {
+      foreach ($result_data['response']['docs'] as $suggestion) {
+        $items[] = check_plain($suggestion['all'][0]);
+      }
+    }
+  }
+
+  return $items;
+}
--- a/modules/ting_search_ortograf/ting_search_ortograf.drush.inc
+++ b/modules/ting_search_ortograf/ting_search_ortograf.drush.inc
@@ -1,0 +1,15 @@
+<?php
+
+/**
+ * @file
+ * This file implements support for drush into the module.
+ */
+
+/**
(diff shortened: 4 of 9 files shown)


modules/ting_search_ortograf/includes/ting_search_ortograf.admin.inc
```diff
@@ -0,0 +1,92 @@
...
+    '#type' => 'fieldset',
+    '#title' => t('Autocomplete URL', array(), array('context' => 'ting_search_ortograf')),
+    '#tree' => FALSE,
+    '#collapsible' => TRUE,
+  ); 
+
+  $form['ting_search_ortograf_webservice']['ting_search_ortograf_suggestion_url'] = array(
+    '#type' => 'textfield',
+    '#title' => t('Autocomplete URL', array(), array('context' => 'ting_search_ortograf')),
+    '#size' => 73,
```
kasperg: Why set this very specific size?

If we want to control the layout then please use the style sheet.

modules/ting_search_ortograf/includes/ting_search_ortograf.admin.inc
```diff
@@ -0,0 +1,92 @@
...
+      'data' => $path . '/css/ting_search_ortograf_admin_form.css',
+    ),
+  );
+
+  $form['ting_search_ortograf_webservice'] = array(
+    '#type' => 'fieldset',
+    '#title' => t('Autocomplete URL', array(), array('context' => 'ting_search_ortograf')),
+    '#tree' => FALSE,
+    '#collapsible' => TRUE,
+  );
```
kasperg: This seems to be a field set containing a single field with the same title as the fieldset itself. That seems silly.

modules/ting_search_ortograf/includes/ting_search_ortograf.admin.inc
```diff
@@ -0,0 +1,92 @@
...
+    '#type' => 'fieldset',
+    '#title' => t('Autocomplete settings', array(), array('context' => 'ting_search_ortograf')),
+    '#tree' => TRUE,
+    '#collapsible' => TRUE,
+  );
+
+  $form['ting_search_ortograf_settings']['rows'] = array(
+    '#type' => 'textfield',
+    '#title' => t('Max suggestions', array(), array('context' => 'ting_search_ortograf')),
+    '#size' => 26,
```
kasperg: Why set this very specific size?

If we want to control the layout then please use the style sheet.

modules/ting_search_ortograf/includes/ting_search_ortograf.admin.inc
```diff
@@ -0,0 +1,92 @@
...
+    '#default_value' => $default_settings['rows'],
+  );
+
+  $form['ting_search_ortograf_settings']['minimumString'] = array(
+    '#type' => 'textfield',
+    '#title' => t('Minimum string', array(), array('context' => 'ting_search_ortograf')),
+    '#description' => t('Minimum string length to do autocomplete lookup on.', array(), array('context' => 'ting_search_ortograf')),
+    '#size' => 26,
```
kasperg: Why set this very specific size?

If we want to control the layout then please use the style sheet.

modules/ting_search_ortograf/includes/ting_search_ortograf.admin.inc
```diff
@@ -0,0 +1,92 @@
...
+  $form['ting_search_ortograf_help']['description'] = array(
+    '#theme' => 'table',
+    '#rows' => array(
+      array(t('Max suggestions:', array(), array('context' => 'ting_search_ortograf')),
+        t('Maximum suggestions to return. Default 10', array(), array('context' => 'ting_search_ortograf'))),
+      array(t('Minimum string:', array(), array('context' => 'ting_search_ortograf')), 
+        t('Minimum characters the user has to type before the suggestionservice is queried.', array(), array('context' => 'ting_search_ortograf'))),
+    ),
+    '#attributes' => array('class' => array('table-class')),
+    '#empty' => t('Your table is empty'),
```
kasperg: This structure seems to be leftovers from the current autocomplete module. The number of fields and the complexity of their descriptions is greatly reduced. Nice. We should be able to move these texts to `#description` for their respective fields.

modules/ting_search_ortograf/includes/ting_search_ortograf.ajax.inc
```diff
@@ -0,0 +1,77 @@
...
+function ting_search_ortograf_get_suggestions(array $query) {
+  $items = array();
+  $rest_url = variable_get('ting_search_ortograf_suggestion_url', '');
+  $url = url($rest_url, array('query' => $query));
+  $headers = array('Accept' => 'application/json');
+
+  $result = drupal_http_request($url, $headers);
+  if (!empty($result->data)) {
+    $result_data = drupal_json_decode($result->data);
+    if (empty($result->error)) {
```
kasperg: The order of the checks here seem wrong. I would expect that we checked for errors first and if there are none then started working with the data.

modules/ting_search_ortograf/includes/ting_search_ortograf.ajax.inc
```diff
@@ -0,0 +1,77 @@
...
+        foreach ($result_data['response']['docs'] as $suggestion) {
+          $suggestion = check_plain($suggestion['all'][0]);
+          $val = truncate_utf8($suggestion, 62, TRUE, TRUE, 1);
+          $key = truncate_utf8($suggestion, 256, TRUE, FALSE, 1);
+          $items[$key] = $val;
+        }
+      }
+    }
+    else {
+      watchdog('ting_search_ortograf', 'Autocomplete error: ' . $result->error . ' (' . $result->code . ")\nURL: $url", array(), $severity = WATCHDOG_ERROR);
```
kasperg: Please pass variables as a separate variable instead of injecteing them directly into the string.

modules/ting_search_ortograf/includes/ting_search_ortograf.ajax.inc
```diff
@@ -0,0 +1,77 @@
...
+
+  $result = drupal_http_request($url, $headers);
+  if (!empty($result->data)) {
+    $result_data = drupal_json_decode($result->data);
+    if (empty($result->error)) {
+      if (!empty($result_data['response']['docs'])) {
+        foreach ($result_data['response']['docs'] as $suggestion) {
+          $suggestion = check_plain($suggestion['all'][0]);
+          $val = truncate_utf8($suggestion, 62, TRUE, TRUE, 1);
+          $key = truncate_utf8($suggestion, 256, TRUE, FALSE, 1);
```
kasperg: This handling of the suggestions seems very specific to the autocomplete case. I suggest moving that to `ting_search_ortograf_callback()` - preferably with a comment describing why it is needed.

This way we can keep `ting_search_ortograf_get_suggestions` clean and reusable for other cases.

modules/ting_search_ortograf/ting_search_ortograf.module
```diff
@@ -0,0 +1,72 @@
...
+}
+
+
+/**
+ * Get settings, and merge with defaults.
+ *
+ * @return array
+ *   Search suggestion settings.
+ */
+function ting_search_ortograf_settings() {
```
kasperg: This function seems to fall between two chairs:

1. Managing the configuration for the module with fallback to defaults - except that it does not include the url for the service and thus require manual handling in `ting_search_ortograf_admin()`
2. Providing a shorthand for query arguments to the service - except that `minimumString` is not a query argument and thus require manual handling in `ting_search_ortograf_callback()`

I suggest that you decide on one of the two and then keep all the manual handling in either `ting_search_ortograf_admin()` or `ting_search_ortograf_callback()`.

modules/ting_search_ortograf/ting_search_ortograf.module
```diff
@@ -0,0 +1,72 @@
...
+  $settings = array();
```
kasperg: We do not need to set `$settings` here. It is not accessed before it is overwritten on L70.

modules/ting_search_ortograf/includes/ting_search_ortograf.ajax.inc
```diff
@@ -0,0 +1,77 @@
...
+
+  // Try to use cache for the search string.
+  $cid = md5(serialize($query));
+  if ($cache = cache_get($cid, 'cache_ting_search_ortograf')) {
+    $matches = $cache->data;
+  }
+  else {
+    $matches = ting_search_ortograf_get_suggestions($query);
+    // Store the matches in cache for faster lookup next time.
+    cache_set($cid, $matches, 'cache_ting_search_ortograf', CACHE_TEMPORARY);
```
kasperg: Is `CACHE_TEMPORARY` really the right approach here? It seems as if the index we search in is quite stable as it works across all libraries and thus not very prone to changes. A longer cache lifetime would lead to quicker response times which I would consider important for autocompletion

modules/ting_search_ortograf/ting_search_ortograf.module
```diff
@@ -0,0 +1,72 @@
...
+ * Add auto-complete functions to the ting search field, using the Ortograf service.
+ */
+
+/**
+ * Implements hook_menu().
+ */
+function ting_search_ortograf_menu() {
+  $items = array();
+
+  $items['ting/autocomplete'] = array(
```
kasperg: Please add `ortograf ` to the path. It is especially important here as another module uses the exact same menu path.

modules/ting_search_ortograf/ting_search_ortograf.drush.inc
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * @file
+ * This file implements support for drush into the module.
+ */
+
+/**
+ * Implements hook_drush_cache_clear().
+ *
+ * Adds a cache clear option for the recommendation cache.
```
kasperg: Wee bit of copy/paste here?

modules/ting_search_ortograf/COPYING
```diff
@@ -0,0 +1,33 @@
...
+The GPL license doesn't repeal the intellectual properties which
+belonges to those TING partners and the legal person who will enforce
+the rights is DBC on behalf of the TING.Community.
+
+DBC A/S (dbc.dk) is a public owned company working with bibliographic
+services and software development for cultural sector, especially the
+libraries.
+
+TING.Community is a community defined as "An open eco-system for
+digital [cultural] cooperation, sharing and development."
```
kasperg: I do not think this file is needed. We already provide a `COPYING` in the root of our repository.

modules/ting_search_ortograf/includes/ting_search_ortograf.admin.inc
```diff
@@ -0,0 +1,92 @@
+<?php
+
+/**
+ * @file
+ * Admin settings for bibliotek.dk autocomplete.
```
kasperg: Copy/paste here?

modules/ting_search_ortograf/includes/ting_search_ortograf.admin.inc
```diff
@@ -0,0 +1,92 @@
...
+    '#title' => t('Autocomplete settings', array(), array('context' => 'ting_search_ortograf')),
+    '#tree' => TRUE,
+    '#collapsible' => TRUE,
+  );
+
+  $form['ting_search_ortograf_settings']['rows'] = array(
+    '#type' => 'textfield',
+    '#title' => t('Max suggestions', array(), array('context' => 'ting_search_ortograf')),
+    '#size' => 26,
+    '#default_value' => $default_settings['rows'],
```
kasperg: Validation like [`element_validate_integer_positive`](https://api.drupal.org/api/drupal/includes%21form.inc/function/element_validate_integer_positive/7.x) would be nice here.

modules/ting_search_ortograf/includes/ting_search_ortograf.admin.inc
```diff
@@ -0,0 +1,92 @@
...
+  );
+
+  $form['ting_search_ortograf_settings']['minimumString'] = array(
+    '#type' => 'textfield',
+    '#title' => t('Minimum string', array(), array('context' => 'ting_search_ortograf')),
+    '#description' => t('Minimum string length to do autocomplete lookup on.', array(), array('context' => 'ting_search_ortograf')),
+    '#size' => 26,
+    '#default_value' => $default_settings['minimumString'],
```
kasperg: Validation like [`element_validate_integer_positive`](https://api.drupal.org/api/drupal/includes%21form.inc/function/element_validate_integer_positive/7.x) would be nice here.

modules/ting_search_ortograf/css/ting_search_ortograf_admin_form.css
```diff
@@ -0,0 +1,23 @@
+#ting-search-ortograf-admin .form-type-textfield {
```
kasperg: I imagine this stylesheet might be leftovers from `ting_search_autocomplete`. Is it really needed now? We do not have as many fields to manage.

modules/ting_search_ortograf/includes/ting_search_ortograf.ajax.inc
```diff
@@ -0,0 +1,77 @@
+<?php
+/**
+ * @file
+ * AJAX functions bibliotek.dk autocomplete.
```
kasperg: More copy/paste here?

modules/ding_ting_frontend/ding_ting_frontend.info
```diff
@@ -25,7 +25,6 @@ dependencies[] = ting_proxy
 dependencies[] = ting_reference
 dependencies[] = ting_relation
 dependencies[] = ting_search
-dependencies[] = ting_search_autocomplete
```
kasperg: :+1: - but remember that an autocomplete module should still be enabled by default on new installations. Consider enabling `ting_search_autocomplete` as a part of the profile installation.

modules/ting_search_ortograf/includes/ting_search_ortograf.admin.inc
JrgenGNielsen: ...but looks nice

modules/ting_search_ortograf/includes/ting_search_ortograf.ajax.inc
JrgenGNielsen: Autocomplete responses are not entirely stable - new titles are added on a daily basis. 
Cron jobs (and cache clear) are set to run once daily, which sounds about right to me.

kasperg: :+1:

modules/ting_search_ortograf/includes/ting_search_ortograf.admin.inc
```diff
@@ -0,0 +1,68 @@
...
+function ting_search_ortograf_admin_validate($form, &$form_state) {
+  $form_state['values']['ting_search_ortograf_suggestion_url'] = trim($form_state['values']['ting_search_ortograf_suggestion_url']);
+  if (!filter_var($form_state['values']['ting_search_ortograf_suggestion_url'], FILTER_VALIDATE_URL)) {
+    form_set_error(
+      'ting_search_ortograf_suggestion_url',
+      t('Invalid URL: %url',
+        array('%url' => $form_state['values']['ting_search_ortograf_suggestion_url']),
+        array('context' => 'ting_search_ortograf'))
+      );
+  }
```
kasperg: Nicely done.

modules/ting_search_ortograf/includes/ting_search_ortograf.ajax.inc
```diff
@@ -0,0 +1,92 @@
...
+  }
+  else {
+    $suggestions = ting_search_ortograf_get_suggestions($query);
+    // Maximum space in suggestions dropdown is 62 characters.
+    // Maximum characters for search query string is 256 characters.
+    foreach ($suggestions as $suggestion) {
+      $val = truncate_utf8($suggestion, 62, TRUE, TRUE, 1);
+      $key = truncate_utf8($suggestion, 256, TRUE, FALSE, 1);
+      $matches[$key] = $val;
+    }
```
kasperg: :+1: Appreciate the comments.

PR #941, opened 2017-12-21
2610: Fix fatal error when editing libraries

holt83:
http://platform.dandigbib.org/issues/2610

The changes made in https://platform.dandigbib.org/issues/773 gives fatal error when editing a library under some circumstances.

Removing just $form['path']['alias'] gives rise to problems, since the validate handler for the alias is attached to the parent element. When core path module sets a default value for the alias the validate handler will try to access the element we unset, giving rise to a fatal error later on. The reason this doesn't happen often is because there often seems to be a disperancy between the language of the node and the URL-alias. When this is the case, the default is not set and there's no problem. But for some reason, some libraries manage to have a setup, where it finds a default value.

It doesn't seem like a good idea to just remove the validate handler, since it performs validation on alias generated by pathauto pattern.

The easiset solution seemt to be to re-introduce $form['path']['alias'] and disable it both on server and in frontend. We also need to remove the states array, since it will appear like it's possible to re-enable it in frontend if present.

I tested: 
- The correct alias is still added when creating library.
- The correct alias is maintained when updating. Even when manually inserting another in dev-tools.


modules/ding_library/ding_library.module
```diff
@@ -216,7 +216,13 @@ function ding_library_disable_custom_alias($form, &$form_state) {
...
     $form['path']['pathauto']['#attached']['js'][] = drupal_get_path('module', 'ding_library') . '/ding_library.js';
 
-    unset($form['path']['alias']);
+    // We have to keep the alias textfield, since removing it causes a fatal
+    // error in the parent element validate handler, so we just tell the FORM
+    // API it's disabled and make sure it's disabled in frontend.
+    $form['path']['alias']['#disabled'] = TRUE;
+    // For some reason this is necessary.
+    $form['path']['alias']['#attributes'] = array('disabled' => TRUE);
+    unset($form['path']['alias']['#states']);
```
kasperg: Have you tried setting `$form['path']['#access']` to `FALSE` instead of `$form['path']['alias']['#disabled'] = TRUE;`? This way the form element will not appear in the rendered form but it will still be available serverside.

This might also remove the need for modifying `#attributes` and `#states`.

kasperg: I am not very happy about the existing solution with JavaScript used to faking a checkbox. This change only seems to pile on top of the problem making a bad situation worse.

I see from previous changes that setting `#access` to `FALSE` has been tried - also by yourself in PR #730. To summarize the problem with that solution is that if new library nodes are created then they will not have a path alias generated by `pathauto` either.

I tried investigation the solution and found [this issue](https://www.drupal.org/project/pathauto/issues/2110365#comment-9250617) which seems to describe our problem exactly and provide a solution, which would allow us to remove all other workarounds.

Could you test this out?

kasperg: If we have to keep this it is worth noting:

1. `#states` is set by `pathauto`
2. Adding this disabled attribute manually might be necessary because Drupal Core would normal do this by checking for `#disabled` elements before be set this ourselves.
3. This overwrites `#attributes` completely. It should be sufficient to merge in `[ 'disabled' => TRUE]`

holt83: @kasperg Much better solution :)

I just tested it and it works, so I updated the PR with new solution.

holt83:
@kasperg I updated the PR with the proposed solution

modules/ding_library/ding_library.module
kasperg: @holt83 Thanks. Appreciate the effort.

PR #944, opened 2017-12-22
3157: Openlist analysis

BirkAndMe:
https://platform.dandigbib.org/issues/3157

### DIFF
--- a/modules/p2/ting_openlist/include/ting_openlist_analysis.inc
+++ b/modules/p2/ting_openlist/include/ting_openlist_analysis.inc
@@ -10,13 +10,6 @@ function ting_openlist_get_local_analysis() {
   return '<pre>' . htmlentities(var_export($data, TRUE)) . '</pre>';
 }
 
-function ting_openlist_submit_local_analysis() {
-  dpm(ding_provider_invoke('openlist', 'call_module', 'LocalAnalysis', 'sendData', array(ting_openlist_get_local_analysis_data())));
-
-
-  return "Stats are sent.";
-}
-
 function ting_openlist_get_local_analysis_data() {
   $data = array();
 
--- a/modules/p2/ting_openlist/include/ting_openlist_analysis.inc
+++ b/modules/p2/ting_openlist/include/ting_openlist_analysis.inc
@@ -5,45 +5,40 @@
  * Functions for analysis of local OpenList Entities
  */
 
-function ting_openlist_get_local_analysis() {
-  $data = ting_openlist_get_local_analysis_data();
-  return '<pre>' . htmlentities(var_export($data, TRUE)) . '</pre>';
-}
-
-function ting_openlist_get_local_analysis_data() {
-  $data = array();
-
-  // Providers.
+function ting_openlist_local_analysis_data_providers() {
   $query = db_select('users', 'u');
   $query->join('users_roles', 'ur', 'ur.uid = u.uid');
   $query->join('role', 'r', "r.rid = ur.rid AND r.name = 'provider'");
-  $data['providers'] = $query->countQuery()->execute()->fetchField();
+  return $query->countQuery()->execute()->fetchField();
+}
 
-  // Providers openlist id
+function ting_openlist_local_analysis_data_providers_id() {
   $query = db_select('users', 'u');
   $query->join('users_roles', 'ur', 'ur.uid = u.uid');
   $query->join('role', 'r', "r.rid = ur.rid AND r.name = 'provider'");
   $query->condition('openlist_uid', NULL, 'IS NOT');
-  $data['providers_openlist_id'] = $query->countQuery()->execute()->fetchField();
+  return $query->countQuery()->execute()->fetchField();
+}
 
-  // Providers openlist is sync
+function ting_openlist_local_analysis_data_providers_openlist_is_sync() {
   $query = db_select('users', 'u');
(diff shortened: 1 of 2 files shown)


modules/p2/ting_openlist/include/ting_openlist_analysis.inc
```diff
@@ -0,0 +1,97 @@
...
+  $query->addField('s', 'cnt');
+  $query
+    ->groupBy('s.uid')
+    ->groupBy('s.cnt')
+    ->orderBy('cnt', 'DESC')
+    ->range(0, 10);
+  $data['top10_list'] = $query->execute()->fetchAllKeyed();
+
+  // Element counts
+  $squery = db_select('field_data_field_list_objects', 'flo');
```
kasperg: The use of `$squery` is quite confusing.

1. The name does not reveal what the variable represents. I do not understand what the `s` stands for.
2. Through this >7OLOC function the same variable is both defined (L44), modified in new code blocks (L64-65) and overwritten (L78). That is quite different to `$query` where each block starts with a definition and ends with a query.

modules/p2/ting_openlist/include/ting_openlist_analysis.inc
```diff
@@ -0,0 +1,97 @@
...
+}
+
+function ting_openlist_submit_local_analysis() {
+  dpm(ding_provider_invoke('openlist', 'call_module', 'LocalAnalysis', 'sendData', array(ting_openlist_get_local_analysis_data())));
+
+
+  return "Stats are sent.";
+}
+
+function ting_openlist_get_local_analysis_data() {
```
kasperg: Please add documentation of what this function does.

The result of this function is passed directly to OpenList for analysis so it is important that the structure of the return value is described.

modules/p2/ting_openlist/include/ting_openlist_analysis.inc
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * @file
+ * Functions for analysis of local OpenList Entities
+ */
+
+function ting_openlist_get_local_analysis() {
+  $data = ting_openlist_get_local_analysis_data();
+  return '<pre>' . htmlentities(var_export($data, TRUE)) . '</pre>';
+}
```
kasperg: I do not think this function provides much value on its own. It seems directly tied to `ting_openlist_show_local_analysis_form_callback()` so the logic might as well be placed there.

modules/p2/ting_openlist/include/ting_openlist_analysis.inc
```diff
@@ -0,0 +1,150 @@
+<?php
+
+/**
+ * @file
+ * Functions for analysis of local OpenList Entities
+ */
+
+/**
+ * Get providers.
```
kasperg: Please change the documentation to "provider users" instead of "providers". A provider will usually be module/system which integrates with Ding using the `ding_provider` module.

I know that Drupal users created through the provider system also have the role "provider". This is unfortunate but also hard to change. This is why we need to be precise in our documentation.

modules/p2/ting_openlist/include/ting_openlist_analysis.inc
```diff
@@ -0,0 +1,150 @@
...
+  $query->condition('openlist_modified', 0, '>');
+  return $query->countQuery()->execute()->fetchField();
+}
+
+/**
+ * Count the types of lists.
+ *
+ * @return array
+ *   An associative array, where the key is the list_type_X (X being number of
+ *   lists) and the key is number of occurences.
```
kasperg: This sounds odd to me.

As I read it if there are 42 lists of type `books_read` then the array would contain the following key-value pair:

`books_read_42 => 42`

Perhaps I do not understand the inner workings of the OpenList analysis or I misunderstand the comment.

modules/p2/ting_openlist/include/ting_openlist_analysis.inc
```diff
@@ -0,0 +1,150 @@
...
+    ->orderBy('s.cnt');
+  return $query->execute()->fetchAllKeyed();
+}
+
+/**
+ * The top 10 duplicate lists.
+ *
+ * @return array
+ *   Keyed array, where they key is the openlist_uid and value the number of
+ *   list duplicates.
```
kasperg: I am afraid do not understand the documentation of this function either. It says that the function returns "duplicate lists" but it seems to return an array where the keys are OpenList *user* ids.

Could it be that the method returned the top 10 *users* ordered by the number of duplicate lists they have?

modules/p2/ting_openlist/include/ting_openlist_analysis.inc
```diff
@@ -0,0 +1,150 @@
...
+    ->range(0, 10);
+  return $query->execute()->fetchAllKeyed();
+}
+
+/**
+ * Get lists with duplicate elements.
+ *
+ * @return array
+ *   An associative array, where the key tells how many duplicates there is, and
+ *   the value is the number of occurences.
```
kasperg: Again I might be misunderstanding something. The DocBlock says that the function returns "lists with duplicate elements".

As I read it does not return any actual lists but rather the number of times a list duplicates occur. Example: If 10 lists have been duplicated 3 times and, 7 lists have been duplicated 2 times then the method would return

```php
[
  3 => 10,
  2 => 7,
]
```

PR #947, opened 2018-01-02
3059: Force color scheme to custom

BirkAndMe:
https://platform.dandigbib.org/issues/3059


cableman:
Reviewed and awaiting release

PR #951, opened 2018-01-03
3139: Changed event book button color

b14sha:
https://platform.dandigbib.org/issues/3139


cableman:
Reviewed and awaiting release

PR #955, opened 2018-01-03
3192: Prevent notices for invalid categories

kasperg:
If the provided category argument is included in the available
categories we should not check for panes. Doing this will result in
notices and warnings


cableman:
Reviewed and awaiting release

PR #957, opened 2018-01-09
2593: Openlist REST fixes

kasperg:
https://platform.dandigbib.org/issues/2593

@tripox and I have experienced some problems with the introduction of exceptions for OpenList. These should be addressed by this PR.

Primary problems:

1. Class `TingOpenlistRequestException` cannot be found. It must be included.
2. `TingOpenlistRequestException` is always thrown when making requests to OpenList

### DIFF
--- a/modules/p2/ding_list/include/functions.handler.inc
+++ b/modules/p2/ding_list/include/functions.handler.inc
@@ -275,7 +275,8 @@ function ding_list_handle_fields_in_list(&$list, array $fields) {
     foreach ($fields as $field) {
       switch ($field['name']) {
         case 'field_ding_content_tags':
-          $field['value'] = ding_list_handle_taxonomy_terms($field['value'], $field['name']);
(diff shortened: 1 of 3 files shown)


modules/p2/ding_list/include/functions.handler.inc
```diff
@@ -275,7 +275,8 @@ function ding_list_handle_fields_in_list(&$list, array $fields) {
     foreach ($fields as $field) {
       switch ($field['name']) {
         case 'field_ding_content_tags':
-          $field['value'] = ding_list_handle_taxonomy_terms($field['value'], $field['name']);
+          $field['value'] = ding_list_handle_taxonomy_terms([$field['value']],
+            $field['name']);
```
cableman: Thinks it would be better to have this in one line

cableman:
Reviewed and approved

PR #959, opened 2018-01-10
3190: Sort by res-que and title

b14sha:
https://platform.dandigbib.org/issues/3190


cableman:
Reviewed and awaiting release

PR #961, opened 2018-01-10
3090 shelfmark fixes

holt83:
This PR contains some minor corrections i made when looking at the issue in:

https://platform.dandigbib.org/issues/3090

The main issue is fixed in a PR against ting-client: https://github.com/ding2/ting-client/pull/23, as mentioned in the case above.


modules/fbs/includes/fbs.availability.inc
```diff
@@ -296,8 +296,13 @@ function fbs_translate_marc($code, $field) {
...
     $translations = _fbs_get_marc_039a_translations();
   }
-  else if ($field == '039.b') {
-    $translations = _fbs_get_marc_039b_translations();
+  elseif ($field == '039.b') {
+    // This field contains ISO 3166-1 alpha-2 country codes and we can use the
+    // build in list from the Drupal.
+    // See: http://www.kat-format.dk/danMARC2/Danmarc2.22.htm#pgfId=1574621
+    include_once DRUPAL_ROOT . '/includes/locale.inc';
+    $translation = country_get_list();
```
kasperg: :+1: Appreciate this refactoring. 

However it seems this has to be `$translations` - not `$translation` - for this to work?

Also consider noting that [the country codes in the spec](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2) are in upper case while they are in lower case in the MARC data. Consequently we need to uppercase the code. We could also avoid modifying the input and lowercase all keys in `country_get_list()`.

holt83: I fixed the typo.

I already do uppercase the code in the line below or is there something I'm missing?

kasperg: I know.

My suggestion was to modify keys from `country_get_list()` instead of the function parameter.

PR #962, opened 2018-01-12
2742: Combined event filter breaks view

BirkAndMe:
https://platform.dandigbib.org/issues/2742


cableman:
Reviewed and awaiting release

PR #964, opened 2018-01-15
3121: Link field required on ding-campaign

b14sha:
https://platform.dandigbib.org/issues/3121


cableman:
Reviewed and awaiting release

PR #965, opened 2018-01-16
3091: Fixed 4/3 format on E-resource

b14sha:
https://platform.dandigbib.org/issues/3091


modules/ding_frontend/ding_frontend.info
```diff
@@ -3,9 +3,6 @@ description = Global frontend module.
 core = 7.x
 package = Ding!
 project = ding_frontend
-files[] = includes/callback_bundle_filter.inc
-files[] = includes/callback_published_filter.inc
-files[] = includes/callback_user_role_filter.inc
```
cableman: This must not be removed as this will break search for homepage content.

cableman:
Reviewed and awaiting release

PR #870, opened 2017-10-30
B2697v2

b14sha:
https://platform.dandigbib.org/issues/2697

PR #966, opened 2018-01-16
B2697v3

BirkAndMe:
https://platform.dandigbib.org/issues/2697


kasperg:
This is somewhat complex.

The change depends on reapplying #870 which was reverted in #933. On the other hand #1072 has made significant changes to lists.

PR #968, opened 2018-01-16
3224: Added event organizer taxonomy and field

cableman:
Addes taxonomy with event organizers to ding event content type.

See https://platform.dandigbib.org/issues/3224


cableman:
Rebased

DDBdrift:
https://platform.dandigbib.org/issues/3224 | Resolved

PR #692, opened 2017-08-01
2193: Use default admin_views instead of overidden

holt83:
http://platform.dandigbib.org/issues/2193

Moves the commits from this PR: https://github.com/ding2/bpi/pull/8 to consol repo.

PR #970, opened 2018-01-17
2193: Don't disable admin_views node view on BPI install

holt83:
I missed this one in the last in the PR.

As a result the admin_views node view isn't activated on fresh installs.

https://platform.dandigbib.org/issues/2193


kasperg:
Preceeding changes to this PR have been reverted in #692 . This changes must be reintroduced before considering these changes.

PR #973, opened 2018-01-18
3115: Fix autocomplete search

JrgenGNielsen:
Fix search on trailing question mark, quotes.
Update autocomplete dropdown styling.

See: https://platform.dandigbib.org/issues/3115


modules/ting_search_ortograf/includes/ting_search_ortograf.ajax.inc
```diff
@@ -32,11 +32,17 @@ function ting_search_ortograf_callback($string = '') {
...
+    // Maximum space in suggestions dropdown is 60 characters.
     // Maximum characters for search query string is 256 characters.
     foreach ($suggestions as $suggestion) {
-      $val = truncate_utf8($suggestion, 62, TRUE, TRUE, 1);
+      $val = truncate_utf8($suggestion, 60, TRUE, TRUE, 1);
       $key = truncate_utf8($suggestion, 256, TRUE, FALSE, 1);
+       // Trim whitespace and question mark.
+      $key = trim($key, " \t\n\r\0\x0B?");
+      $key = str_replace('"', '', $key);
+      if (strpos($suggestion, ' ')) {
```
cableman: Why is it $suggestion and not $key that are check here?

PR #974, opened 2018-01-22
3236: Search profile filters

cableman:
Extension to field search to allow grouping of profiles and define which should be available on which pages.

See https://platform.dandigbib.org/issues/3236


cableman:
New PR: https://github.com/ding2/ding2/pull/1019

PR #575, opened 2017-04-11
1437: infomedia login

xendk:
Add login popup to infomedia articles.

PR #980, opened 2018-01-26
1437: informedia fixes

xendk:


xendk:
Scrutinizer fails hard again.

kasperg:
Preceeding changes from #575 have been reverted in #984 . They must be reintroduced before the changes in this PR.

kasperg:
This is superceeded by #1148 which should contain all changes related to the issue at hand.

PR #981, opened 2018-01-29
2430: News view changed to exposed filters

b14sha:
https://platform.dandigbib.org/issues/2430


cableman:
Needs rebase to match newest core and is there any styling/tpls that will not be used any more should these be removed?

modules/ding_news/ding_news.pages_default.inc
```diff
@@ -1296,7 +1320,7 @@ node/%node:nid/news
   $display->title_pane = '0';
   $handler->conf['display'] = $display;
   $page->default_handlers[$handler->name] = $handler;
-  $pages['ding_news_library'] = $page;
+  $pages[''] = $page;
```
JrgenGNielsen: do you mean: $pages[] = $page;? (I'm not sure how the array handles a numerical index, though...)

modules/ding_news/ding_news.pages_default.inc
```diff
@@ -994,7 +1018,7 @@ news/%tid:name',
   $display->title_pane = 'new-bb26b1db-7e7e-4c3f-8956-183ed5e31146';
   $handler->conf['display'] = $display;
   $page->default_handlers[$handler->name] = $handler;
-  $pages['ding_news'] = $page;
+  $pages[''] = $page;
```
JrgenGNielsen: see below

modules/ding_news/ding_news.field_group.inc
```diff
@@ -35,7 +35,7 @@ function ding_news_field_group_info() {
       ),
     ),
   );
-  $field_groups['group_ding_news_attachments|node|ding_news|form'] = $field_group;
+  $field_groups[''] = $field_group;
```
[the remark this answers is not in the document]
b14sha: This is an error caused by the features module after a field-group update.

modules/ding_news/ding_news.pages_default.inc
b14sha: This is an error caused by the features module after a field-group update.

modules/ding_news/ding_news.pages_default.inc
b14sha: This is an error caused by the features module after a field-group update.

JrgenGNielsen:
reviewed and approved

PR #984, opened 2018-01-29
Revert "1437: infomedia login"

kasperg:
Reverts ding2/ding2#575


kasperg:
This changes have en reintroduced with #1148

PR #990, opened 2018-01-29
3109: Opening hours expand options

b14sha:
https://platform.dandigbib.org/issues/3109

### DIFF
--- a/themes/ddbasic/scripts/ddbasic.common.js
+++ b/themes/ddbasic/scripts/ddbasic.common.js
@@ -6,6 +6,8 @@
    * Toggle opening hours
    */
   function toggle_opening_hours() {
+    var hasOpeningHours = Drupal.settings.hasOwnProperty('ding_ddbasic_opening_hours');
+
     // Create toggle link
     $('.js-opening-hours-toggle-element').each(function () {
       var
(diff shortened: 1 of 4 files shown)


cableman:
Please look into the Scrutinizer errors!

themes/ddbasic/scripts/ddbasic.common.js
```diff
@@ -20,11 +20,21 @@
         text.push(Drupal.t('Opening hours'));
       }
 
-      $('<a />', {
-        'class' : 'opening-hours-toggle js-opening-hours-toggle js-collapsed collapsed',
-        'href' : Drupal.t('#toggle-opening-hours'),
-        'text' : text.join(', ')
-      }).insertBefore(this);
+      if (Drupal.settings.ding_ddbasic_opening_hours && Drupal.settings.ding_ddbasic_opening_hours.expand_all_libraries) {
```
cableman: A better way to test if properties exits would be to use hasOwnProperty().

themes/ddbasic/scripts/ddbasic.common.js
```diff
@@ -50,8 +60,8 @@
       event.preventDefault();
     });
 
-    // Expand opening hours on library pages.
-    if (Drupal.settings.ding_ddbasic_opening_hours && Drupal.settings.ding_ddbasic_opening_hours.expand_on_library) {
+    // Expand opening hours on first library on library pages.
+    if (Drupal.settings.ding_ddbasic_opening_hours && Drupal.settings.ding_ddbasic_opening_hours.expand_on_first_library) {
```
cableman: A better way to test if properties exits would be to use hasOwnProperty().

themes/ddbasic/scripts/ddbasic.common.js
cableman: You could also preform the test once (as its used more than once in the code) and put it into a variable.

JrgenGNielsen:
Can you fix the Scrutinizer error on hasOpeningHours definition?

JrgenGNielsen:
reviewed and approved

PR #995, opened 2018-01-29
Feature/3199 carousel ajax

JrgenGNielsen:
See: https://platform.dandigbib.org/issues/3199


cableman:
The kode looks fine. But would like to take this to a core meeting as '#attached' is the recommended way or using drupal_add_js.

PR #997, opened 2018-01-30
3263: Ensure we can fetch relation

holt83:
https://platform.dandigbib.org/issues/3263

Background info and analysis is described on the issue-tracker. See link above.

### DIFF
--- a/modules/ting/ting.entities.inc
+++ b/modules/ting/ting.entities.inc
@@ -233,7 +233,12 @@ class TingEntity extends DingEntity {
         }
       }
       foreach ($this->reply->relationsData as $record) {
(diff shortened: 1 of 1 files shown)


cableman:
Please rebase to match newest core.

holt83:
This issue has been resolved with the implementation of SAL.

First, the offending was moved to the opensearch module:
https://github.com/ding2/ding2/blob/master/modules/opensearch/src/OpenSearchTingObject.php#L102 

In this new version it iterates over the newly created relation_objects array, which has just been searched and collected from the well. Compared to the old implementation, that just iterated over the original reply's relationsData:

https://github.com/vejlebib/ding2/blob/393d72dc9139eb9be1a35fd35b6aedba8469fad6/modules/ting/ting.entities.inc#L235 

This ensures that we're not trying to setup non-searchable relations.

cableman:
👍

PR #1000, opened 2018-02-05
3272: Provider intergration docs

kasperg:
https://platform.dandigbib.org/issues/3272

This is an attempt at providing documentation for the provider integration and list requirements for library systems and search systems in case new providers have to be evaluated in regard to compliance with Ding2.

Note that some of the documentation assumes that the search abstraction layer in #982 is merged.

### DIFF
--- a/docs/provider_integrations.md
+++ b/docs/provider_integrations.md
@@ -1,0 +1,176 @@
+# Provider integrations
+
+Ding2 supports integration with other business systems for handling different aspects of a library website such as searching the library catalogue, determining the status of materials and authenticating patrons. Within Ding2 a system handling one or more of these aspects is called a *provider system*.
+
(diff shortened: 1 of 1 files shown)


docs/provider_integrations.md
```diff
@@ -0,0 +1,266 @@
...
+ */
+function ding_reservation_ding_provider_user() {
+  return array(
+    'reservation' => array(
+      'required' => TRUE,
+      'install time setup' => TRUE,
+    ),
+  );
+}
+```
```
cableman: Mayby a comment about the parameters documentation for the two hooks can be found in https://github.com/ding2/ding2/blob/master/modules/ding_provider/ding_provider.api.php

kasperg: :+1: - see b96ec80a7b69c4cffb831f3c7cf4e21006e3238f.

cableman:
Reviewed and awaiting release

PR #1004, opened 2018-02-09
38: Add rest_server as dependency to ding_services

tripox:


JrgenGNielsen:
reviewed and approved

PR #1006, opened 2018-02-13
3287 by Inlead: Changed path to correct patch.

guddo:
https://platform.dandigbib.org/issues/3287


cableman:
Approved and awaiting release

DDBdrift:
https://platform.dandigbib.org/issues/3287 | Resolved (tag version)

PR #1008, opened 2018-02-14
3270 by Inlead: Cardinality violation upon Search.

bozeak:
https://platform.dandigbib.org/issues/3270

### DIFF
--- a/modules/ding_frontend/ding_frontend.module
+++ b/modules/ding_frontend/ding_frontend.module
@@ -248,3 +248,23 @@ function ding_frontend_views_query_alter(&$view, &$query) {
       break;
   }
 }
+
+/**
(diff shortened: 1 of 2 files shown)


cableman:
Maybe this belongs in ding_frontend as it has the dependency to search_api!

bozeak:
This error was generated by calling and execution of "ding_multiple_search" view in search_backends.inc ting_search plugin (https://github.com/ding2/ding2/blob/master/modules/ting_search/plugins/content_types/search_backends.inc#L78) and this was main reason to keep it here.

cableman:
Yes, but that view is stille defined in ding_frontend and not ting_search module

cableman:
Reviewed and awaiting release

PR #1010, opened 2018-02-16
3035 by Inlead: Adjust ding_redia_rss module

bozeak:
https://platform.dandigbib.org/issues/3035


ding2.make
```diff
@@ -417,6 +417,18 @@ projects[views_bulk_operations][version] = "3.3"
...
+projects[views_rss][subdir] = "contrib"
+projects[views_rss][version] = "2.0-rc4"
+
+; This specific checkout is only because of the module is dev branch only.
+projects[views_rss_media][type] = "module"
+projects[views_rss_media][subdir] = "contrib"
+projects[views_rss_media][download][type] = "git"
+projects[views_rss_media][download][url] = "http://git.drupal.org/project/views_rss_media.git"
+projects[views_rss_media][download][revision] = "14f7cc90d41c0186d2356ff528ac316a16eba3fd"
+projects[views_rss_media][patch][] = "https://www.drupal.org/files/issues/strict-warning-2149287-1.patch"
```
cableman: It would be nice with a comment about these patches and what they fixes.

modules/ding_content_rss/ding_content_rss.features.inc
```diff
@@ -0,0 +1,65 @@
...
+    'show_contextual_link' => 0,
+    'i18n' => 0,
+    'module' => 'ding_content_rss',
+    'locked' => array(
+      0 => 'name',
+    ),
+    'api_version' => 2,
+  );
+  return $flags;
+
```
cableman: Nitpicking: Why the empty line after the return statement?

modules/ding_content_rss/ding_content_rss.info
```diff
@@ -0,0 +1,25 @@
+name = Ding Content RSS
+description = Ding Content RSS feeds
+core = 7.x
+package = Ding!
+version = 7.x-4.1.0
```
cableman: Normally we don't have the version tag in DDB CMS custom modules.

modules/ding_content_rss/ding_content_rss.install
```diff
@@ -0,0 +1,37 @@
...
+ * Implements hook_install().
+ */
+function ding_content_rss_install() {
+  db_update('system')
+    ->fields(array('weight' => 10))
+    ->condition('name', 'ding_content_rss', '=')
+    ->execute();
+}
+
+/**
```
cableman: Normally new modules to the DDB CMS don't have update hooks. But is this module already used in production on some site? If so I think we need to leave them here.

modules/ding_content_rss/ding_content_rss.module
```diff
@@ -0,0 +1,216 @@
...
+ */
+function ding_content_rss_field_formatter_settings_form($field, $instance, $view_mode, $form, &$form_state) {
+  $display = $instance['display'][$view_mode];
+  $settings = $display['settings'];
+  $form = array();
+  $form['fromto'] = array(
+    '#title' => t('Display:'),
+    '#type' => 'select',
+    '#options' => array(
+      'both' => t('Both Start and End dates'),
```
cableman: I might be missing something here, but https://github.com/ding2/ding2/pull/1010/files#diff-4518520246760e0297f29fbeb0664724R54 only runs strtotime on a single value?

modules/ding_content_rss/ding_content_rss.views_default.inc
```diff
@@ -0,0 +1,768 @@
+<?php
```
cableman: Is the views export done with the newest DDB CMS core version as features have been upgraded in core. Just to prevent the view from being marked as overridden all the time :-)

modules/ding_content_rss/views/ding_content_rss_handler_library_id.inc
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * Class ding_content_rss_handler_library_id.
+ */
+// @codingStandardsIgnoreStart
+class ding_content_rss_handler_library_id extends views_handler_field {
+
+  /**
+   * {@inheritdoc}
+   */
+  function option_definition() {
```
cableman: This function don't seam to do anything but call the parent object, so it could be omitted.

modules/ding_content_rss/views/ding_content_rss_handler_library_id.inc
```diff
@@ -0,0 +1,45 @@
...
+    $options = parent::option_definition();
+    return $options;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  function options_form(&$form, &$form_state) {
```
cableman: This function don't seam to do anything but call the parent object, so it could be omitted.

modules/ding_content_rss/views/ding_content_rss_handler_library_id.inc
```diff
@@ -0,0 +1,45 @@
...
+    parent::options_form($form, $form_state);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  function query() {}
```
cableman: Empty function? It would be nice with a commit about why is empty?

modules/ding_content_rss/views/ding_content_rss_handler_library_id.inc
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * @file
+ * Node's "Library ID" custom field handler.
+ */
+
+/**
+ * Class ding_content_rss_handler_library_id.
+ */
+// @codingStandardsIgnoreStart
```
cableman: I think that this "@codingStandardsIgnoreStart" and the "@codingStandardsIgnoreEnd" should be removed or the case why not should be given?

modules/ding_place2book/ding_place2book.module
```diff
@@ -1049,3 +1124,12 @@ function ding_place2book_theme() {
   );
   return $items;
 }
+
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function ding_place2book_preprocess_node(&$variables) {
```
cableman: A single comment about why this "access" false is here would be nice.

modules/ding_content_rss/ding_content_rss.install
guddo: Yes, they are already in use on some sites.

modules/ding_content_rss/ding_content_rss.info
guddo: This will be overriden on the next release, with the script you are using to create versions.

modules/ding_content_rss/ding_content_rss.views_default.inc
guddo: Not sure. What is the newest version? Can you refer to a ref/sha?

modules/ding_content_rss/views/ding_content_rss_handler_library_id.inc
bozeak: Those are used here because PHPCS utility is recognizing non camel case case definition as a non-standard and failing Scrutinizer check. Those annotations are allowing us to skip any code standards checking on this piece of code.

modules/ding_place2book/ding_place2book.module
bozeak: See https://github.com/ding2/ding2/pull/1010/commits/f0fe3bfb10e7bde6e501b55c1028a54d668458f1#diff-8b99672f6dd58da79fd6b49feeca23adR1133

modules/ding_content_rss/ding_content_rss.views_default.inc
cableman: DDB CMS master branch is always the last release :-)

modules/ding_content_rss/ding_content_rss.install
cableman: 👍

modules/ding_content_rss/views/ding_content_rss_handler_library_id.inc
cableman: Yes, please remove them.

The guideline says that you should comment, in the PR, if scrutinizer throws errors that the core team should take into account when reviewing.

Please see https://github.com/ding2/ding2/blob/master/CONTRIBUTING.md point 10 and 11.

modules/ding_content_rss/ding_content_rss.info
```diff
@@ -0,0 +1,24 @@
...
+package = Ding!
+project = ding_content_rss
+dependencies[] = cache_actions
+dependencies[] = ctools
+dependencies[] = entity
+dependencies[] = flag
+dependencies[] = rules
+dependencies[] = views
+dependencies[] = views_rss
+dependencies[] = views_rss_media
```
cableman: We might also here need to have a dependencies to ding_news and ding_event?

ding2.make
cableman: 👍

modules/ding_place2book/ding_place2book.module
cableman: 👍

modules/ding_content_rss/ding_content_rss.info
bozeak: Dependencies to ding_news and ding_event  was added.

modules/ding_content_rss/views/ding_content_rss_handler_library_id.inc
bozeak: Removed annotations.

PR #1011, opened 2018-02-19
3270 by Inlead: Extending list of escaped symbols.

bozeak:
https://platform.dandigbib.org/issues/3270


cableman:
Reviewed and awaiting release

PR #1013, opened 2018-02-20
3265: Slider for facet date

BirkAndMe:

The solution is specific for the date facet, and not generalized for other OR filters.


holt83:
If this is supposed to work with ting_field_search, it would nice if the logic that sets up the slider could be factored out into it's own function.

PR #1018, opened 2018-02-22
3315: Update Drupal Core to 7.57

kasperg:
https://platform.dandigbib.org/issues/3315


cableman:
Reviewed and approved

PR #1019, opened 2018-02-22
3236: Added search field profile filters

cableman:
Extension to field search to allow grouping of profiles and define which should be available on which pages.

See https://platform.dandigbib.org/issues/3236

### DIFF
--- a/modules/ting_field_search/plugins/ting_field_search_ctools_export_ui.inc
+++ b/modules/ting_field_search/plugins/ting_field_search_ctools_export_ui.inc
@@ -1,8 +1,10 @@
 <?php
 
 /**
+ * @file
  * Defines the Ting field search profile export UI.
  */
+
 $plugin = array(
   'schema' => 'ting_field_search_profile',
   'access' => 'administer site configuration',
@@ -22,7 +24,7 @@ $plugin = array(
 
   'form' => array(
     'settings' => 'ting_field_search_profile_export_form',
-    'submit' => 'ting_field_search_profile_export_form_submit'
+    'submit' => 'ting_field_search_profile_export_form_submit',
   ),
 
   'handler' => array(
@@ -86,7 +88,7 @@ function ting_field_search_profile_export_form(&$form, &$form_state) {
   $form['config']['search_request'] = array(
     '#type' => 'fieldset',
     '#title' => t('Search request'),
-    '#description' => '<strong>' .  t('Settings related to the search request') . '</strong>',
+    '#description' => '<strong>' . t('Settings related to the search request') . '</strong>',
     '#collapsible' => TRUE,
     '#collapsed' => TRUE,
     '#weight' => -15,
@@ -96,7 +98,7 @@ function ting_field_search_profile_export_form(&$form, &$form_state) {
     '#title' => t('Library code'),
     '#description' => t('Use an alternative library code. Leave empty to use the <strong>%standard</strong> code configured at <a href="!url"> Ting search settings</a>.', array(
       '%standard' => variable_get('ting_agency', ''),
-      '!url' => url('admin/config/ting/settings')
+      '!url' => url('admin/config/ting/settings'),
     )),
   );
   $form['config']['search_request']['well_profile'] = array(
@@ -112,7 +114,7 @@ function ting_field_search_profile_export_form(&$form, &$form_state) {
     '#type' => 'textfield',
     '#title' => t('Query'),
     '#description' => t('Enter a base search query that is appended to every search request, when using this profile.'),
-    '#maxlength' => 256,
+    '#maxlength' => 1024,
     '#default_value' => $profile->config['search_request']['query'],
     '#size' => 100,
   );
@@ -130,7 +132,7 @@ function ting_field_search_profile_export_form(&$form, &$form_state) {
     '#default_value' => $profile->config['search_request']['allow_empty'],
   );
 
-  // User interaction
+  // User interaction.
   $form['config']['user_interaction'] = array(
     '#type' => 'fieldset',
     '#title' => t('User interaction'),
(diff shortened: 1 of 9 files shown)


modules/ting_field_search/plugins/ting_field_search_ctools_export_ui.inc
```diff
@@ -158,11 +160,61 @@ function ting_field_search_profile_export_form(&$form, &$form_state) {
     '#default_value' => $profile->config['user_interaction']['alter_links'],
   );
 
+  $form['config']['user_interaction']['grouping'] = array(
+    '#type' => 'textfield',
+    '#title' => t('Profile grouping'),
+    '#description' => t('Which group of search profiles should be shown when this profile is used.'),
```
kasperg: I do not understand what this description means regarding groups of profiles to be shown when this profile is used. I cannot tell how a profile is added to a group.

modules/ting_field_search/plugins/ting_field_search_ctools_export_ui.inc
```diff
@@ -158,11 +160,61 @@ function ting_field_search_profile_export_form(&$form, &$form_state) {
...
+    '#default_value' => !empty($profile->config['user_interaction']['grouping']) ? $profile->config['user_interaction']['grouping'] : 'default',
+  );
+  $form['config']['user_interaction']['visibility'] = array(
+    '#type' => 'radios',
+    '#title' => t('Show block on specific pages'),
```
kasperg: What does "Block" refer to here? Perhaps a wee bit of copy/paste?

modules/ting_field_search/ting_field_search.module
```diff
@@ -432,39 +433,202 @@ function ting_field_search_ting_search_settings_submit($form, &$form_state) {
...
+    }
+
+    $visibility = $profile->config['user_interaction']['visibility'] == 'expect' ? 0 : 1;
+    $pages = $profile->config['user_interaction']['pages'];
+    $path = drupal_strtolower(drupal_get_path_alias($_GET['q']));
+
+    // We have an optimization hack to speed up search (by-pass panels 2x
+    // times loading). So all searches are submitted to "/search-blank",
+    // which need to be added to pages to ensure no form validation errors.
+    // If visibility is "expect" it should not be added to the pages list.
```
kasperg: :+1:

modules/ting_field_search/plugins/ting_field_search_ctools_export_ui.inc
```diff
@@ -158,11 +160,61 @@ function ting_field_search_profile_export_form(&$form, &$form_state) {
...
+    '#options' => array(
+      'all' => t('Show on all pages.'),
+      'expect' => t('All pages except those listed'),
```
kasperg: `expect` = `except` - nitpicking but since this is used as a key value later on it should be addressed. Consider constants instead.

modules/ting_field_search/ting_field_search.module
```diff
@@ -432,39 +433,202 @@ function ting_field_search_ting_search_settings_submit($form, &$form_state) {
...
  */
-function ting_field_search_form_search_block_form_alter(&$form, &$form_state) {
-  if (!variable_get('ting_field_search_search_style', FALSE)) {
-    return;
+function ting_field_search_preprocess_html(&$vars) {
+  $profiles = ting_field_search_profiles_load_all();
+
+  foreach ($profiles as $profile) {
+    // Expended search box not checked go to the next profile.
+    if (!array_key_exists('search_box', $profile->config['user_interaction']) || !$profile->config['user_interaction']['search_box']) {
```
kasperg: Could this be replaced with 

```php
if (!empty($profile->config['user_interaction']['search_box'])) {
```

modules/ting_field_search/ting_field_search.module
```diff
@@ -432,39 +433,202 @@ function ting_field_search_ting_search_settings_submit($form, &$form_state) {
...
+  // If there's an active profile and it is stick; set it as default value.
+  if ($default_profile = ting_field_search_get_active_profile()) {
+    $default_profile = $default_profile->config['user_interaction']['sticky'] ? $default_profile->name : '';
+  }
+
+  if (empty($default_profile)) {
+    $profiles = ting_field_search_profiles_load_all();
+    foreach ($profiles as $profile) {
+      // Expended search box not checked go to the next profile.
+      if (!array_key_exists('default_profile', $profile->config['user_interaction']) || !$profile->config['user_interaction']['default_profile']) {
```
kasperg: Some copy/paste in the comment? Also see suggestion regarding value check above.

modules/ting_field_search/ting_field_search.module
```diff
@@ -432,39 +433,202 @@ function ting_field_search_ting_search_settings_submit($form, &$form_state) {
...
+        continue;
+      }
+
+      // If the profile has visibility all always expand the search field.
+      if ($profile->config['user_interaction']['visibility'] === 'all') {
```
kasperg: Do we need to check for the existence of the `visibility` entry here?

modules/ting_field_search/ting_field_search.module
```diff
@@ -432,39 +433,202 @@ function ting_field_search_ting_search_settings_submit($form, &$form_state) {
...
+      $page_match = drupal_match_path($path, $pages);
+      if ($path != $_GET['q']) {
+        $page_match = $page_match || drupal_match_path($_GET['q'], $pages);
+      }
+
+      // When $visibility has a value of 0, the profile is displayed on
+      // all pages except those listed in $pages. When set to 1, it
+      // is displayed only on those pages listed in $pages.
+      if (!($visibility xor $page_match)) {
+        $default_profile = $profile->name;
```
kasperg: This seems to be almost the same code as in `ting_field_search_preprocess_html()`. It is also quite complex. Please refactor.

modules/ting_field_search/ting_field_search.module
```diff
@@ -432,39 +433,202 @@ function ting_field_search_ting_search_settings_submit($form, &$form_state) {
...
+        // If visibility is "expect" it should not be added to the pages list.
+        if ($visibility) {
+          $pages .= "\r\nsearch-blank";
+        }
+
+        // Compare the lowercase internal and lowercase path alias (if any).
+        $page_match = drupal_match_path($path, $pages);
+        if ($path != $_GET['q']) {
+          $page_match = $page_match || drupal_match_path($_GET['q'], $pages);
+        }
```
kasperg: Less copy/paste - more filtering.

modules/ting_field_search/ting_field_search.module
```diff
@@ -618,29 +782,33 @@ function ting_field_search_profiles_load($name) {
...
  *
- * @return array $profiles.
+ * @return array
  *   An array of profile objects keyed bt profile machine name.
  */
 function ting_field_search_profiles_load_all($reset = FALSE) {
-  return ting_field_search_profiles_load_mutliple();
+  $profiles = &drupal_static(__FUNCTION__);
+  if (!isset($profiles) || $reset) {
+    $profiles = ting_field_search_profiles_load_mutliple();
```
kasperg: Typo in `mutliple` but perhaps this matters less here.

modules/ting_field_search/ting_field_search.module
```diff
@@ -823,29 +995,33 @@ function ting_field_search_facet_save($facet) {
...
 /**
  * Delete facet settings associated with a profile.
+ *
+ * @param string $pid
+ *   The id of the profile to delete.
+ * @param string $facet
+ *   Facet to to used as condition for the deletions.
  */
-function ting_field_search_facets_delete($pid, $facet = NULL) {
+function ting_field_search_facets_delete($pid, $facet = '') {
```
kasperg: Why change the default value here?

modules/ting_field_search/ting_field_search.ting_field_search_default_profiles.inc
```diff
@@ -927,6 +957,43 @@ function ting_field_search_ting_field_search_default_profiles() {
...
+    'facets' => array(
+      'use_facets' => 0,
+      'number_of_facets' => 25,
+      'facet_count' => 5,
+      'term_count' => 5,
+    ),
+  );
+  $profile->weight = 11;
+  $profile->facets = array();
+  $export['tegneserier'] = $profile;
```
kasperg: Is it intentional that this profile has been added?

themes/ddbasic/template.php
```diff
@@ -34,8 +34,11 @@ function ddbasic_preprocess_html(&$vars) {
       $vars['classes_array'][] = 'search-form-extended';
       $vars['classes_array'][] = 'show-secondary-menu';
 
-      if (menu_get_item()['path'] === 'search/ting/%') {
-        $vars['classes_array'][] = 'extended-search-is-open';
+      $path = menu_get_item()['path'];
+      switch ($path) {
+        case 'search/ting/%';
+        case 'ding_frontpage':
```
kasperg: Adding `ding_frontpage` seems to be a change which is not related to the issue at hand. Is that intentional?

cableman: Yes, as I remember this fixes som issues where you have selected the search field to default be open.

modules/ting_field_search/ting_field_search.ting_field_search_default_profiles.inc
cableman: It's back in feb. so i really don't remember who request this!

modules/ting_field_search/plugins/ting_field_search_ctools_export_ui.inc
cableman: Simply grouped by the string/name entered in the field.

modules/ting_field_search/plugins/ting_field_search_ctools_export_ui.inc
cableman: Fixed :-)

modules/ting_field_search/plugins/ting_field_search_ctools_export_ui.inc
cableman: Fixed

modules/ting_field_search/ting_field_search.module
cableman: It have been changed from your suggestion, based on internal code review to the current form.

modules/ting_field_search/ting_field_search.module
cableman: The comment from above her as well.

modules/ting_field_search/ting_field_search.module
cableman: This is taken from drupal core, so think this is fine.

modules/ting_field_search/ting_field_search.module
cableman: ?

modules/ting_field_search/ting_field_search.module
cableman: Fixed

modules/ting_field_search/ting_field_search.module
cableman: As all facets are strings, so having NULL as default don't makes sens. The function documentation declare $facet a string.

modules/ting_field_search/ting_field_search.module
cableman: I don't remeber why, but a lot of testing (by GBA) has gone into this and there was a lot og issue with frontpage vs. other pages etc....

modules/ting_field_search/ting_field_search.module
kasperg: As far as I can tell [the Block module in Drupal Core where this is taken from has this once](https://github.com/drupal/drupal/blob/7.x/modules/block/block.module#L819-L831) and it does not include an optimization hack requiring 4 lines of comments so I do not think it is the same.

See aakbcms/patches#1 for a suggestion for how to refactor.

modules/ting_field_search/ting_field_search.module
kasperg: My comment refers to the situation above.

modules/ting_field_search/ting_field_search.module
cableman: Changes accepted

PR #1020, opened 2018-02-23
3233: Material in public lists

BirkAndMe:
https://platform.dandigbib.org/issues/3233


JrgenGNielsen:
CircleCI fejler: 
Error: Call to undefined function ctools_include() in page_manager_page_load() (line 384 of /home/ubuntu/build/profiles/ding2/modules/contrib/ctools/page_manager/plugins/tasks/page.inc).
Drush command terminated abnormally due to an unrecoverable error.

https://github.com/ding2/ding2/pull/1020/files#diff-94c44b3d1893a203c328bb754fabbdcfR33 sætter `required=1`. 
Det ser ud til at det kan skabe problemer: Se evt. https://github.com/ding2/ding2/blob/286cfc39c4b7aba62a08f19aa1b6903e6433ce75/modules/fbs/fbs.info#L21-L27

PR #1021, opened 2018-02-28
731: Removed sorting on AcquisitionDate

cableman:
This removes AcquisitionDate from 4.2.x releases of ddb CMS.

It should not be merged as core has moved on.

See https://platform.dandigbib.org/issues/731


DDBdrift:
https://platform.dandigbib.org/issues/731 | Resolved (tag version)

PR #1024, opened 2018-03-01
2685: Enable opensearch module as search provider on existing sites

kasperg:
https://platform.dandigbib.org/issues/2685

The current codebase enables the new opensearch module as the default
search provider on new sites during installation. However nothing is
done to ensure that a search provider is active on existing sites.

Consequently we implement an update hook which enables the opensearch
module if no other search provider is active.


cableman:
Reviewed and approved

PR #1025, opened 2018-03-01
3356: Return null if OpenSearch cannot retrieve collection for id

kasperg:
https://platform.dandigbib.org/issues/3356

This prevents fatal errors opensearch_get_object can return multiple
types including NULL. OpenSearchTingObjectCollection only accepts a
TingClientObjectCollection.

Returning NULL from the search provider collection_load method should
not cause problems. Callers already check for empty results.

Update DocBlocks accordingly.


cableman:
Reviewed and awaiting release

PR #1026, opened 2018-03-01
3349: Fixed event view labels and submit

b14sha:
https://platform.dandigbib.org/issues/3349


DDBdrift:
https://platform.dandigbib.org/issues/3349 | Resolved (tag version)

PR #1027, opened 2018-03-01
3339: Only use trailing wildcard for reference queries

kasperg:
https://platform.dandigbib.org/issues/3339

Fuzzy search with add wildcards characters before and after the search
string. This makes opensearch reply much slower and requests time out.

Do not use fuzzy search and instead add a trailing wildcard. This
also matches previous behavior.

Reference queries where the only place where this functionality was
used. Remove it entirely to simplify things a bit before usage and
confusion whether it should be used or not spreads.


cableman:
Reviewed and awaiting release

PR #1028, opened 2018-03-01
3338: Move infomedia administration fields to new opensearch form

kasperg:
https://platform.dandigbib.org/issues/3338

The module is called ting_infomedia but requests to the infomedia
service uses the client library also used to access Opensearch.

Consequently we update the form alter for the module to target the
opensearch admininistration form.

We also solidify the dependency on Opensearch by adding the module as
a dependency for ting_infomedia. In theory it should be possible to 
decouple the two services down the line if needed.


cableman:
Reviewed and awaiting release

PR #1029, opened 2018-03-02
3352: Fix author and subject links when not using facets

kasperg:
https://platform.dandigbib.org/issues/3352

The current implementation included url decoding of generated links.
This was added to support from field values to urls containing query
parameters which are double encoded - primarily facets. This causes 
problems when the url does not contain query parameters as these links
are not double encoded.

This change introduces ? as a separator between the search string and 
query parameters. Handling is moved to a single function and 
surrounding code is cleaned up.


cableman:
Reviewed and awaiting release

PR #1030, opened 2018-03-05
2685: Fix retrieval of primary object for collection in Connie

kasperg:
An invalid property will always be empty.

This should also fix the tests.


cableman:
Reviewed and awaiting release

PR #1034, opened 2018-03-07
3320: Fixed p2-list covered by header

b14sha:
https://platform.dandigbib.org/issues/3320


cableman:
Reviewed and awaiting release

PR #1035, opened 2018-03-08
3381: Remember to include ui.widget

xendk:
See https://platform.dandigbib.org/issues/3381


cableman:
Reviewed and awaiting release

PR #1037, opened 2018-03-08
3187: Ajax/lazy load ting reference objects

cableman:
On some pages many (20+) of reference to data well object can lead to page timeouts if note cached yet.

This PR simple ajax/lazy loads the reference onto the page in the same manner as ding_carousel do and thereby giving the user a better perceived load speed. 

See https://platform.dandigbib.org/issues/3187


holt83:
Clearly a much better solution than the current!

I thought about this before, and would've thought that it could also be handled by preloading all the entities to the cache with a single request to the well using hook_field_prepare_view(). (much like we do on the status pages).

cableman:
The big issue here is when cache is empty, so using hook_field_prepare_view would that not still slow down the "first" load of the page?

holt83:
No, because the idea is the preload them all in one big rec.id instead of loading all individually with a getObject request.

cableman:
Yes and using that trick will slow down the first page load after a cache clear. It's faster than requesting them one by one... but it's still to slow to load 30+ object from the data-well into cache in on request than the user should wait for a page view.

Therefore the _perceive_ page load time (not the actual) is faster by ajax loading them one by one. When the objects are cached (which they will be using ether load method) the time will basic be the same for the next views of the page.

holt83:
Of course, I get what you mean ;)

Still, the page load time should not exceed the typical search request page load time, which I think is very acceptable by now. Also, it should take up fewer resources, since less request are made.

Perhaps both methods could be combined in a future performance optimization..

modules/ting_reference/js/ting_reference_ajax.js
```diff
@@ -0,0 +1,48 @@
...
+ */
+(function ($) {
+  "use strict";
+
+  Drupal.behaviors.ting_reference_ajax = {
+    attach: function (context) {
+      $('.ting-reference-item', context).once('ting-reference-item', function () {
+        var elm = $(this);
+        var entity_id = elm.data('entity-id');
+        var view_mode = elm.data('view-mode');
```
kasperg: :+1: for the use of data attributes here

modules/ting_reference/js/ting_reference_ajax.js
```diff
@@ -0,0 +1,48 @@
...
+              for (var i in matches) {
+                var match = matches[i];
+                var id = match.substring(match.indexOf(':') + 1);
+                match = match.replace('reservation', 'availability').replace(':', '');
+                Drupal.settings.ding_availability[match] = [ id ];
+              }
+            }
+
+            // Ensure that behaviors are attached to the new content.
+            Drupal.attachBehaviors($('.ting-reference-item'));
```
kasperg: It seems as if we only need to attach behaviors to the newly inserted element - not all elements with the `ting-reference-item` class. Also consider using a `js-` prefix according to our frontend style guide.

modules/ting_reference/templates/ting_reference_item.tpl.php
```diff
@@ -0,0 +1,17 @@
...
+ * @file
+ * Default ting reference item template.
+ *
+ * Available variables:
+ *  - $content: Item content.
+ *  - $entity_id: The id of the entity referenced.
+ *  - $view_mode: The view mode for the reference.
+ */
+?>
+<span class="<?php print $classes; ?>" data-entity-id="<?php print $entity_id; ?>" data-view-mode="<?php print $view_mode; ?>">
```
kasperg: Perhaps it is just me being oldschool but nesting `div` inside `span` does not seem right. Is there a reason for not just using `div`s?

themes/ddbasic/sass/components/field.scss
```diff
@@ -300,3 +300,49 @@ div.search-field-in-content--message {
...
+    margin-right: 18px;
+  }
+
+  // Mobile
+  @include media($mobile) {
+    width: 218px;
+    margin-right: 18px;
+  }
+
+  .placeholder {
```
kasperg: [I think I have seen this before](https://github.com/ding2/ding2/blob/af8922fc3ff45e6344e07daa43cdf51b10ed5830/themes/ddbasic/sass/components/ding-carousel.scss#L136) and it seems generally applicable. Consider refactoring into a mixin or the like.

modules/ting_reference/ting_reference.module
```diff
@@ -403,6 +416,29 @@ function ting_reference_field_formatter_view($entity_type, $entity, $field, $ins
...
+  // Load the ting objects found.
+  $entity = entity_load('ting_object', array($entity_id));
+  $ting_object = ting_object_view(reset($entity), $view_mode);
+
+  $build = array(
+    'id' => $entity_id,
+    'content' => render($ting_object),
+  );
+
+  drupal_json_output($build);
```
kasperg: This seems to be a very generic approach to rendering ting objects asynchronously. Could this and associated logic be moved to the `ting` module so it could be used more widely in the future?

modules/ting_reference/js/ting_reference_ajax.js
```diff
@@ -0,0 +1,48 @@
...
+        var entity_id = elm.data('entity-id');
+        var view_mode = elm.data('view-mode');
+        $.ajax({
+          type: 'get',
+          url : Drupal.settings.basePath + 'ting_reference/ajax/' + entity_id + '/' + view_mode,
+          dataType : 'json',
+          success : function (data) {
+            elm.html(data.content);
+
+            // This ensures that ting objects loaded via ajax in the carousel's gets
```
kasperg: This comment references carousels.

modules/ting_reference/ting_reference.module
```diff
@@ -192,16 +213,11 @@ function _ting_reference_field_update($entity_type, $entity, &$items) {
...
-
 /**
  * Field_update for cron.
  */
 function _ting_reference_field_update_cron($entity_type, $entity, &$items) {
 
-  // Get the entity's id.
-  $entity_ids = entity_extract_ids($entity_type, $entity);
-  $entity_id = array_shift($entity_ids);
-
```
kasperg: :+1:

themes/ddbasic/sass/components/field.scss
```diff
@@ -300,3 +300,49 @@ div.search-field-in-content--message {
...
+        top: 100%;
+        left: 0;
+        width: 100%;
+        height: 100%;
+        background: url('../../images/spinner.png') no-repeat;
+        background-size: 100% auto;
+      }
+    }
+  }
+}
```
kasperg: Missing newline.

modules/ting_reference/js/ting_reference_ajax.js
```diff
@@ -0,0 +1,48 @@
...
+            // reservations buttons displayed if available. So basically it finds
+            // the material ids and coverts them into ding_availability format and
+            // updates the settings, which is this used when behaviors are attached
+            // below. This is a hack, but the alternative was to re-write
+            // ding_availability.
```
kasperg: ... and now we are repeating the code 😕 .

modules/ting_reference/js/ting_reference_ajax.js
cableman: Fixed

modules/ting_reference/js/ting_reference_ajax.js
cableman: Yes, we are as the comment says to avoid this ding_availability has to be rewritten to get around not having to do this and we don't have a ding_js_utils library.

modules/ting_reference/js/ting_reference_ajax.js
cableman: Fixed; so it only attach to the current element.

modules/ting_reference/templates/ting_reference_item.tpl.php
cableman: Changed to "div".

modules/ting_reference/ting_reference.module
cableman: Moved into helper function.

themes/ddbasic/sass/components/field.scss
cableman: That might be the case but I can't see any mixin's in other than the contrib folder in the theme?

themes/ddbasic/sass/components/field.scss
cableman: Fixed

PR #1040, opened 2018-03-09
3325: Move periodical detection to search system

kasperg:
The current detection of periodicals based on types is not suitable
across search providers which may use different languages.

The current values are tied to OpenSearch and the Danish municipal 
libraries. Consequently we move the detection to the opensearch module.


cableman:
Reviewed and awaiting release

PR #1041, opened 2018-03-09
3335 by Inlead: Add Core patch to fix tablescroll.

guddo:
https://platform.dandigbib.org/issues/3335


JrgenGNielsen:
reviewed and approved

PR #1042, opened 2018-03-12
3346: Initialize OpenSearch relations to empty list

kasperg:
[Relations for the object will only be set if they can be retrieved](https://github.com/ding2/ding2/compare/master...reload:feature/3346-fix-relation-notice?expand=1#diff-cd656e119bf1220b60ff65335fffe0feL69).
Consequently we have to initialize the property to an empty array to
avoid returning NULL if relations cannot be retrieved.


cableman:
Reviewed and awaiting release

PR #1043, opened 2018-03-12
3352: Fix notice when search string does not contain query

kasperg:
https://platform.dandigbib.org/issues/3352

preg_split will only return a single entry causing list to issue a
notice.


cableman:
Reviewed and awaiting release

PR #1007, opened 2018-02-14
1744 by Inlead: Moving ting_search_carousel into panels.

bozeak:
http://platform.dandigbib.org/issues/1744

PR #1044, opened 2018-03-12
3036 by Inlead: Adjust ding_redia_variables module.

bozeak:
https://platform.dandigbib.org/issues/3036


modules/ding_variables/ding_variables.info
```diff
@@ -0,0 +1,6 @@
+name = Ding Variables
+description = Exposed custom variables as a JSON format.
+core = 7.x
+package = Ding!
+version = 7.x-3.0.5
```
cableman: Normally we don't have version tags inside the custom module files.

modules/ding_variables/ding_variables.module
```diff
@@ -0,0 +1,85 @@
...
+}
+
+/**
+ * Provide variables to the json output.
+ *
+ * @return array
+ *   List of all variables to the output.
+ */
+function ding_variables_display() {
+  // TODO: Find a way to remove hardcode.
```
cableman: Should this "todo" be fixed or removed?

modules/ding_variables/ding_variables.module
```diff
@@ -0,0 +1,85 @@
...
+    'ting_search_profile',
+    'ting_search_results_per_page',
+    'ting_search_result_message_limit',
+    'ting_search_url',
+    'ting_sort_default',
+    'ting_ranking_custom',
+    'ting_ranking_fields',
+    'ting_boost_fields',
+    'ting_agency',
+  );
```
cableman: Please note that within the 4.3.x the SAL release these variables have been moved into the opensearch module and I think most of them simply can be renamed to opensearch_*.

@kasperg can all these simply be renamed?

kasperg: Yes. Check [`opensearch_migrate_settings()`](https://github.com/ding2/ding2/blob/master/modules/opensearch/opensearch.install#L18) for new names.

One notable exception is `ting_agency` - [that is bound to change later on though](https://platform.dandigbib.org/issues/3322).

modules/ding_variables/ding_variables.module
```diff
@@ -0,0 +1,85 @@
...
+ * @return array
+ *   List of carousels pane configurations.
+ */
+function _ding_variables_get_carousel_configs($carousel) {
+  $query = db_select('panels_pane', 'pp');
+  $result = $query
+    ->fields('pp', array('subtype', 'configuration'))
+    ->condition('subtype', $carousel)
+    ->execute()
+    ->fetchAll();
```
kasperg: This seems to assume that #1007 is merged. Is that true?

modules/ding_variables/ding_variables.info
kasperg: Also there are no dependencies mentioned here. That does not seem right.

ding2.install
```diff
@@ -564,3 +564,10 @@ function ding2_update_7045() {
 function ding2_update_7046() {
   ding2_translation_update();
 }
+
+/**
+ * Enable ding_variables module.
+ */
+function ding2_update_7047() {
+  module_enable(array('ding_variables'));
```
kasperg: This probably only needs to be enabled for libraries using `opensearch` - especially if `opensearch` is added as a dependency.

modules/ding_variables/ding_variables.module
bozeak: As I know we are still waiting for this to be merged.

cableman: An that PR is firste set to next release and therefor we can't use this before that. I have contact DDB in regards to this issue.

guddo: Ok, let us know if we need to refactor to current release.

cableman: Will do, awaiting Gitte and Rolf

cableman: @guddo after talk with Gitte: We would like the carousels export to use the current data structure as the changes to the workflow (moving them into panes) is not set for this release. But the app has to be released before that release. 

Short version: Yes, please refactor.

cableman:
Reviewed and awaiting release

PR #1048, opened 2018-03-14
3037 by Inlead: Adjust easyddb_smartbanner module.

bozeak:
https://platform.dandigbib.org/issues/3037


modules/ding_smartbanner/ding_smartbanner.module
```diff
@@ -0,0 +1,170 @@
...
+ * Implements hook_init().
+ */
+function ding_smartbanner_init() {
+  $configs = variable_get('smartbanner_config', array());
+
+  drupal_add_css(libraries_get_path('smart-app-banner') . '/smart-app-banner.css');
+  drupal_add_js(libraries_get_path('smart-app-banner') . '/smart-app-banner.js', array('type' => 'file', 'scope' => 'footer'));
+  drupal_add_js(DING_SMARTBANNER_MODULE_PATH . '/js/ding_smartbanner.js', array('type' => 'file', 'scope' => 'footer'));
+
+  $icon_path = drupal_get_path('module', 'ding_smartbanner') . '/icon.png';
```
NicolaeGrigoriu: DING_SMARTBANNER_MODULE_PATH is defined and can be used here

modules/ding_smartbanner/ding_smartbanner.admin.inc
```diff
@@ -0,0 +1,293 @@
...
+
+  $form['smartbanner']['base_config']['app_store_language'] = array(
+    '#type' => 'select',
+    '#title' => t('iTunes store language'),
+    '#description' => t('Select language for iTunes store'),
+    '#options' => array(
+      'en' => t('En'),
+      'dk' => t('Dk'),
+    ),
+    '#default_value' => !empty($configs['base_config']['app_store_language']) ? $configs['base_config']['app_store_language'] : 'en',
```
cableman: Default language for DDB CMS is DK, should this default to 'DK'?

modules/ding_smartbanner/ding_smartbanner.info
```diff
@@ -0,0 +1,6 @@
+name = Ding Smartbanner
+description = Module detects mobile device and shows popup with link to native application.
+core = 7.x
+version = 7.x-4.1.0
```
cableman: Normally we don't have the version in the info files as that would require it to be changes on new releases.

modules/ding_smartbanner/ding_smartbanner.install
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * @file
+ * Install and uninstall ding_smartbanner.
+ */
+
+/**
+ * Implements hook_install().
+ */
+function ding_smartbanner_install() {
+  $smartbanner_config = array(
```
cableman: By setting all this default values during installation. One could omit all the "empty" checks in the administration form as they always will exists.

modules/ding_smartbanner/ding_smartbanner.install
```diff
@@ -0,0 +1,57 @@
...
+    'base_config' => array(
+      'days_hidden' => '15',
+      'days_reminder' => '90',
+      'app_store_language' => 'en',
```
cableman: Think the installed language should be 'dk' as this is the default DDB CMS language and the app descriptions below is in danish.

modules/ding_smartbanner/ding_smartbanner.module
```diff
@@ -0,0 +1,170 @@
...
+ * Implements hook_init().
+ */
+function ding_smartbanner_init() {
+  $configs = variable_get('smartbanner_config', array());
+
+  drupal_add_css(libraries_get_path('smart-app-banner') . '/smart-app-banner.css');
+  drupal_add_js(libraries_get_path('smart-app-banner') . '/smart-app-banner.js', array('type' => 'file', 'scope' => 'footer'));
+  drupal_add_js(DING_SMARTBANNER_MODULE_PATH . '/js/ding_smartbanner.js', array('type' => 'file', 'scope' => 'footer'));
+
+  $icon_path = DING_SMARTBANNER_MODULE_PATH . '/icon.png';
```
cableman: Maybe move the default provided icon into "image" subfolder inside the module.

cableman: The default icon commit with this module is that really the library app's default icon?

modules/ding_smartbanner/ding_smartbanner.info
guddo: This will be overriden when you release new version, with the script you use. I assume?

cableman: No, modules don't have version. We only have versions in the profile in ding2.info

modules/ding_smartbanner/ding_smartbanner.install
bozeak: "Empty" check should be present because there are non-mandatory fields. In fact mandatory fields are the one which are described in 'base_config' section of config array, the rest of settings are allowed to be empty.

cableman: 👍

cableman:
Reviewed and awaiting release

PR #1049, opened 2018-03-15
3397: Fixed missing interest period

b14sha:
https://platform.dandigbib.org/issues/3397

### DIFF
--- a/modules/fbs/fbs.info
+++ b/modules/fbs/fbs.info
@@ -12,14 +12,24 @@ dependencies[] = ding_user
 dependencies[] = entity
 dependencies[] = features
 dependencies[] = libraries
+dependencies[] = list
+dependencies[] = options
 dependencies[] = profile2
 dependencies[] = text
 dependencies[] = virtual_field
 dependencies[] = xautoload
+# Features will mark any module that implements hook_field_info() as required.
+# This causes the module to be enabled automatically during site-install, which
+# in the case of a Ding-provider is a problem.
+# To prevent "required = 1" line below has been commented out. When a future
+# feature-export reintroduce the line make sure to ignore it and keep this
+# block of comments and the out-commented line intact.
+# required = 1
 configure = admin/config/ding/provider/fbs
 explanation = Field type(s) in use - see <a href="/admin/reports/fields">Field list</a>
 features[features_api][] = api:2
 features[field_base][] = field_fbs_address
+features[field_base][] = field_fbs_interest_period
 features[field_base][] = field_fbs_mail
 features[field_base][] = field_fbs_mail_notification
 features[field_base][] = field_fbs_name
@@ -28,6 +38,7 @@ features[field_base][] = field_fbs_phone
 features[field_base][] = field_fbs_phone_notification
 features[field_base][] = field_fbs_preferred_branch
 features[field_instance][] = profile2-provider_fbs-field_fbs_address
+features[field_instance][] = profile2-provider_fbs-field_fbs_interest_period
 features[field_instance][] = profile2-provider_fbs-field_fbs_mail
 features[field_instance][] = profile2-provider_fbs-field_fbs_mail_notification
 features[field_instance][] = profile2-provider_fbs-field_fbs_name
@@ -48,11 +59,3 @@ files[] = includes/classes/FBSLogDrupal.php
 files[] = includes/classes/FBSLogInterface.php
 files[] = includes/classes/FBSLogStdout.php
 files[] = includes/classes/FBSMisconfiguredException.php
-mtime = 1499431707
-# Features will mark any module that implements hook_field_info() as required.
-# This causes the module to be enabled automatically during site-install, which
-# in the case of a Ding-provider is a problem.
-# To prevent "required = 1" line below has been commented out. When a future
-# feature-export reintroduce the line make sure to ignore it and keep this
-# block of comments and the out-commented line intact.
-# required = 1
--- a/modules/fbs/includes/fbs.reservation.inc
+++ b/modules/fbs/includes/fbs.reservation.inc
@@ -326,10 +326,14 @@ function fbs_reservation_branch_name($branch_id) {
  *   The default options.
  */
 function fbs_reservation_default_options($account) {
+  // Get values.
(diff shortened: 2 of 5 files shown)


modules/fbs/fbs.info
```diff
@@ -10,16 +10,27 @@ dependencies[] = ding_reservation
 dependencies[] = ding_session_cache
 dependencies[] = ding_user
 dependencies[] = entity
+dependencies[] = fbs
```
cableman: It might raise some issues when FBS is dependent on FBS?

modules/fbs/fbs.info
```diff
@@ -48,11 +60,4 @@ files[] = includes/classes/FBSLogDrupal.php
...
 files[] = includes/classes/FBSMisconfiguredException.php
-mtime = 1499431707
-# Features will mark any module that implements hook_field_info() as required.
-# This causes the module to be enabled automatically during site-install, which
-# in the case of a Ding-provider is a problem.
-# To prevent "required = 1" line below has been commented out. When a future
-# feature-export reintroduce the line make sure to ignore it and keep this
-# block of comments and the out-commented line intact.
-# required = 1
+required = 1
```
cableman: Please **_read_** the comment and remove this required line.

modules/fbs/includes/fbs.reservation.inc
```diff
@@ -326,10 +326,24 @@ function fbs_reservation_branch_name($branch_id) {
...
+  // Use a metadata wrapper to access the data.
+  $wrapper = entity_metadata_wrapper('profile2', $profile);
+
+  // Get values.
+  $interest_period = $wrapper->field_fbs_interest_period->value();
+
+  if (empty($interest_period)) {
+    $interest_period = variable_get('fbs_default_interest_period', 180);
+  }
+
```
cableman: It seams that the code above is during the same as fbs_reservation_default_interest_period() so could that be reused?

modules/fbs/includes/fbs.reservation.inc
```diff
@@ -377,11 +395,26 @@ function fbs_reservation_options_submit($account, $values) {
...
+    $profile = ding_user_provider_profile($user);
+  }
+
+  // Get interest period from the profile.
+  $wrapper = entity_metadata_wrapper('profile2', $profile);
+  $value = $wrapper->field_fbs_interest_period->value();
+
+  // If no period is selected, try getting default value.
+  if (is_null($value) || empty($value)) {
+    $value = variable_get('alma_default_interest_period', 180);
```
cableman: I this you meant to use FBS and not ALMA her?

modules/fbs/includes/fbs.reservation.inc
```diff
@@ -326,10 +326,24 @@ function fbs_reservation_branch_name($branch_id) {
...
+  // Use a metadata wrapper to access the data.
+  $wrapper = entity_metadata_wrapper('profile2', $profile);
+
+  // Get values.
+  $interest_period = $wrapper->field_fbs_interest_period->value();
+
+  if (empty($interest_period)) {
+    $interest_period = fbs_reservation_default_interest_period();
+  }
+
```
cableman: The lines above that looks in the profile for the periode, looks like do the same "fbs_reservation_default_interest_period($account) would do before returning the default periode.

So why do this above when the function do the same?

b14sha: Good point...

modules/fbs/includes/fbs.reservation.inc
```diff
@@ -326,10 +326,20 @@ function fbs_reservation_branch_name($branch_id) {
...
  */
 function fbs_reservation_default_options($account) {
+  // Load profile2 profile.
+  $profile = ding_user_provider_profile($account);
+
+  // Use a metadata wrapper to access the data.
+  $wrapper = entity_metadata_wrapper('profile2', $profile);
+
+  // Get values.
+  $interest_period = fbs_reservation_default_interest_period($profile);
```
cableman: Change the function to take in a $account and only load the profile inside this function.

modules/fbs/includes/fbs.reservation.inc
```diff
@@ -357,9 +361,13 @@ function fbs_reservation_options($account, $default_branch = NULL) {
     $preferred_branch = $default_branch;
   }
 
+  $profile = ding_user_provider_profile($account);
```
cableman: So close :-)

Why load the profile here, when the $profile is only used inside fbs_reservation_default_interest_period();

So why not: 

`$interest_period = fbs_reservation_default_interest_period($account);`

modules/fbs/includes/fbs.reservation.inc
```diff
@@ -377,11 +385,26 @@ function fbs_reservation_options_submit($account, $values) {
...
+ * Implements hook_default_interest_period().
  *
- * @return int
- *   Default interest period for user.
+ * Tries to find the users default interest period based on the users profile2
+ * provider profile.
  */
-function fbs_reservation_default_interest_period() {
-  return variable_get('fbs_default_interest_period', 180);
+function fbs_reservation_default_interest_period($profile = NULL) {
```
cableman: Then this function could be changed to:

```
function fbs_reservation_default_interest_period($account = NULL) {
  $account = is_null($account) ? $GLOBALS['user'] : $account;
  $profile = ding_user_provider_profile($user);

  // Get interest period from the profile.
  $wrapper = entity_metadata_wrapper('profile2', $profile);
  $value = $wrapper->field_fbs_interest_period->value();

  // If no period is selected, try getting default value.
  if (is_null($value) || empty($value)) {
    $value = variable_get('fbs_default_interest_period', 180);
  }
}
```
Or one could use `user_uid_optional_load` if the whole user object was required and not `$GLOBALS['user']`

modules/fbs/includes/fbs.reservation.inc
kasperg: I agree. Passing the profile to `fbs_reservation_default_interest_period` does not make any sense when the function expects an account as the argument and will load the profile based on the passed account (the same as this line does) on L395.

PR #1051, opened 2018-03-15
3399: White arrow on buttons

b14sha:
https://platform.dandigbib.org/issues/3399


cableman:
Reviewed and awaiting release

PR #1052, opened 2018-03-15
2008: Fixed translation for Fee Date

b14sha:
https://platform.dandigbib.org/issues/2008


cableman:

PR #1056, opened 2018-03-15
3408: Added support clone of current search result

cableman:
Enable support for getting a non-referenced version of the current search result and request.

See https://platform.dandigbib.org/issues/3408

### DIFF
--- a/modules/ting_search/ting_search.module
+++ b/modules/ting_search/ting_search.module
@@ -452,20 +452,26 @@ function ting_search_search_execute($keys = NULL, $conditions = NULL) {
 }
 
 /**
- * Retrive the result for a ting_search query performed in this request.
+ * Retrieve the result for a ting_search query performed in this request.
+ *
+ * If you need to change the search query or/and result and don't want the
+ * original store objects to change use the clone parameter to get an new object
+ * and not a reference to the static objects stored.
  *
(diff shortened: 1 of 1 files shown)


kasperg:
Would it not be sufficient to use the `clone` keyword e.g.

```php
$new_requesy = clone ting_search_current_results()->getSearchRequest();
```

cableman:
Updated the code to using "clone" :-) Have been during to much JS where the serialize trick is the faste why to get a clone.

kasperg:
@cableman: Why add a method argument for this at all?

I do not think it is any easier compared  to just using `clone` when you need to clone an object. In this case cloning the search result is actually not enough as the clone will retain references to object properties. That is key here as it is the search request with we want to modify.

To make my opinion clear. I think it would be best to ignore this PR altogether and just call 

```php
$result = ting_search_current_results();
$request = clone $result->getSearchRequest();
```

Instead of 

```php
$result = ting_search_current_results(null, TRUE);
$request = $result->getSearchRequest();
```

It avoids adding arguments to `ting_search_current_results()` and the explicit `clone` is clearer than `TRUE`.

cableman:
I see your point. The only reason for doing it this way was to helper other using the "ting_search_current_results" function so they don't make the same mistake as me and using time one debug why the frontend changes the last search query etc.

PR #1061, opened 2018-03-19
3293: Fixed height on material warning

b14sha:
https://platform.dandigbib.org/issues/3293


cableman:
Reviewed and awaiting release

PR #1062, opened 2018-03-20
3147: Remove Behat test artifacts from build

xendk:
See https://platform.dandigbib.org/issues/3147


cableman:
Reviewed and awaiting release

PR #1063, opened 2018-03-20
3349: Fixed event view labels and submit

b14sha:
https://platform.dandigbib.org/issues/3349


cableman:
Code looks good. It would be nice with a comment about the changes i the tpl from the standard tpl file.

cableman:
Reviewed and awaiting release

PR #1064, opened 2018-03-20
3105: Varnish 4 vcl file

cableman:
This PR adds a configuration file for varnish 4.0 that can be used with DDB CMS.

See https://platform.dandigbib.org/issues/3105

### DIFF
--- a/ding2_v4.vcl
+++ b/ding2_v4.vcl
@@ -162,11 +162,6 @@ sub vcl_backend_response {
     set beresp.ttl = 10m;
   }
 
-  if (beresp.http.x-drupal-cache == "HIT") {
-    unset beresp.http.Cache-Control;
-    set beresp.http.Cache-Control = "public,max-age=21600";
-  }
-
   # Don't allow static files to set cookies.
   # (?i) denotes case insensitive in PCRE (perl compatible regular expressions).
   # This list of extensions appears twice, once here and again in vcl_recv so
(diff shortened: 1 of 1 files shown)


ding2_v4.vcl
```diff
@@ -0,0 +1,183 @@
...
+  # We need this to cache 404s, 301s, 500s. Otherwise, depending on backend but
+  # definitely in Drupal's case these responses are not cacheable by default.
+  if (beresp.status == 404 || beresp.status == 301 || beresp.status == 500) {
+    set beresp.ttl = 10m;
+  }
+
+  if (beresp.http.x-drupal-cache == "HIT") {
+    unset beresp.http.Cache-Control;
+    set beresp.http.Cache-Control = "public,max-age=21600";
+  }
```
kasperg: This always seems to cache pages which are also in Drupals cache with a hardcoded max-age. Why is this?

cableman:
Right, added the vlc_error function.

ding2_v4.vcl
kasperg: @cableman: Where do you see this in [the existing `ding2.vcl`](https://github.com/ding2/ding2/blob/master/ding2.vcl)?

cableman: Right removed it... can't remember why!

kasperg:
I have rebuilt this multiple times. The build keeps failing with something unrelated to the change at hand. Consequently I will merge this despite the error.

PR #1065, opened 2018-03-20
3402: Fix submit handler for availability source and type config

kasperg:
https://platform.dandigbib.org/issues/3402

The previous approach had been renamed and refactored.

Update surrounding documentation as well.


cableman:
Reviewed and awaiting release

PR #1069, opened 2018-03-23
3350: Styling of search results

KristianB14:
https://platform.dandigbib.org/issues/3350


JrgenGNielsen:
It looks like all the [xxx].features.field_base.inc files are machine-generated...
They all omit the 'foreign keys' key in the $field_bases array.
Can you verify that this is intended, and the element is obsolete?

KristianB14:
The features module has been updated, which has resulted in these changes (plus a lot of other changes). It seems the 'foreign key' is omitted instead of exporting a default setting. Everything works as intended.

JrgenGNielsen:
reviewed and approved

PR #1078, opened 2018-04-09
3438: Styling of Opening hours notice

KristianB14:
https://platform.dandigbib.org/issues/3438


cableman:
Reviewed and awaiting release

PR #1080, opened 2018-04-19
3388: Removed empty profile in fulltext request

cableman:
Removed the setProfile from ting fulltext which it then set in the client. It fixes the issue.

### DIFF
--- a/modules/opensearch/includes/opensearch.search.inc
+++ b/modules/opensearch/includes/opensearch.search.inc
@@ -110,6 +110,8 @@ function opensearch_search_collection_load($id) {
  * @return \Ting\Search\TingSearchResultInterface
  *   The search result.
  *
+ * @throws \TingClientException
+ *   If an search error occurs.
  * @throws \Ting\Search\SearchProviderException
  *   If an error occurs during search.
  */
--- a/modules/ting_fulltext/ting_fulltext.module
+++ b/modules/ting_fulltext/ting_fulltext.module
@@ -47,7 +47,6 @@ function ting_fulltext_object_load($object_id) {
 
     $request = opensearch_get_request_factory()->getObjectRequest();
     $request->setObjectId($object_id);
-    $request->setProfile(variable_get('opensearch_search_profile', ''));
     $request->setFormat('docbook');
     $request->setObjectFormat('docbook');
     $request->setOutputType('xml');


cableman:
The scrutinizer error is actually fixed in #1075

modules/opensearch/includes/opensearch.search.inc
```diff
@@ -110,6 +110,8 @@ function opensearch_search_collection_load($id) {
  * @return \Ting\Search\TingSearchResultInterface
  *   The search result.
  *
+ * @throws \TingClientException
+ *   If an search error occurs.
```
kasperg: I do not think such an exception should be thrown from the provider - or at least declared.

The caller should not be aware of what provider is being used and thus not be aware of providerspecific exceptions. 

If we want to handle this properly then we should wrap the location where this exception can be thrown in a `try/catch` and then wrap the `TingClientException` in a `SearchProviderException`.

modules/ting_fulltext/ting_fulltext.module
```diff
@@ -47,7 +47,6 @@ function ting_fulltext_object_load($object_id) {
 
     $request = opensearch_get_request_factory()->getObjectRequest();
     $request->setObjectId($object_id);
-    $request->setProfile(variable_get('opensearch_search_profile', ''));
```
kasperg: Feeling stupid here but do you mind mentioning where this is set?

I can see it in [`opensearch_get_object()`](https://github.com/ding2/ding2/blob/039bcf756df300ba3c833a2781145f115bbf02a8/modules/opensearch/opensearch.client.inc#L65), [`opensearch_get_objects()`](https://github.com/ding2/ding2/blob/039bcf756df300ba3c833a2781145f115bbf02a8/modules/opensearch/opensearch.client.inc#L174) and [`opensearch_do_search()`](https://github.com/ding2/ding2/blob/039bcf756df300ba3c833a2781145f115bbf02a8/modules/opensearch/opensearch.client.inc#L270) but not in general.

If the current approach without a check is a problem then it seems as it [`opensearch_get_object_marcxchange`](https://github.com/ding2/ding2/blob/039bcf756df300ba3c833a2781145f115bbf02a8/modules/opensearch/opensearch.client.inc#L349) also suffers from this. It has not been updated?

modules/opensearch/includes/opensearch.search.inc
cableman: Okay, just filled it in as the IDE was complaining that the function could throw it.

cableman: So this may be outside the scope of this PR.

modules/ting_fulltext/ting_fulltext.module
cableman: Yes, it's set in opensearch_do_search and when the new cache in opensearch gets into core it will always be set.

But for now we just testet that removing it in this situation fixed the issue

modules/opensearch/includes/opensearch.search.inc
kasperg: Then I think it should be removed.

cableman: But removing the comment don't change the fact that this function can throw the exception. So in-complete documentation should not hide that fact.

holt83:
Steen clearly mentioned that it was the agencyid missing that was the problem: https://platform.dandigbib.org/issues/3388#note-2

He also mentions that the format parameter is obsolete and can be removed.

PR #1092, opened 2018-05-07
3066: Use image title as image caption for specific image fields

KristianB14:
https://platform.dandigbib.org/issues/3066


JrgenGNielsen:
reviewed and approved

PR #1099, opened 2018-05-23
3537: Added check that profile exists for interest period

cableman:
When using NemID to create a new library user the profile don't exists. So when trying to get default interest periode and this is the process the periode should use the default value.

See https://platform.dandigbib.org/issues/3537

### DIFF
--- a/modules/fbs/includes/fbs.reservation.inc
+++ b/modules/fbs/includes/fbs.reservation.inc
@@ -389,18 +389,23 @@ function fbs_reservation_options_submit($account, $values) {
  * Tries to find the users default interest period based on the users profile2
  * provider profile.
  */
- function fbs_reservation_default_interest_period($account = NULL) {
+function fbs_reservation_default_interest_period($account = NULL) {
   $account = is_null($account) ? $GLOBALS['user'] : $account;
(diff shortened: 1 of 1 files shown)


modules/fbs/includes/fbs.reservation.inc
```diff
@@ -389,18 +389,23 @@ function fbs_reservation_options_submit($account, $values) {
  * Tries to find the users default interest period based on the users profile2
  * provider profile.
  */
- function fbs_reservation_default_interest_period($account = NULL) {
+function fbs_reservation_default_interest_period($account = NULL) {
   $account = is_null($account) ? $GLOBALS['user'] : $account;
   $profile = ding_user_provider_profile($account);
```
kasperg: It would be nice to make it explicit in [the DocBlock for `ding_user_provider_profile()`](https://github.com/ding2/ding2/blob/e55599a6860bdab0783904abb926af07973108fe/modules/ding_user/ding_user.module#L578-L579) that it may return `NULL`.

kasperg: I added this in 858daca.

PR #1100, opened 2018-05-24
3572 by Inlead: Remove requirement for app_variables enable.

inlead:
https://platform.dandigbib.org/issues/3572

### DIFF
--- a/modules/ting_search/ting_search.install
+++ b/modules/ting_search/ting_search.install
@@ -9,5 +9,5 @@
  * Removing unused 'ting_search_result_message_limit' variable.
  */
 function ting_search_update_7001() {
-  variable_delete('ting_search_result_message_limit');
+  variable_del('ting_search_result_message_limit');
 }


ding2.install
```diff
@@ -576,9 +576,7 @@ function ding2_update_7047() {
  * Enable ding_app_variables module.
  */
 function ding2_update_7048() {
-  if (module_exists('opensearch')) {
-    module_enable(array('ding_app_variables'));
-  }
+  module_enable(array('ding_app_variables'));
```
kasperg: This will enable `ding_app_variables` and it's dependencies on all sites - even if they do not have Opensearch.

Sites decided whether to enable `opensearch` or not in `ting_update_7014()` which was a part of the system since 7.x-4.3.0. If we can risk running the update hook on earlier sites, then we should declare the dependency on that hook using `hook_update_dependencies()`.

kasperg: I took the liberty of adding a change for this in 33ec233.

PR #1102, opened 2018-05-24
 3581 by Inlead: Removed ding_mobilesearch from sources.

inlead:
~~https://platform.dandigbib.org/issues/3565~~
https://platform.dandigbib.org/issues/3581


inlead:
Requires: https://github.com/ding2/ding2/pull/1101

PR #1105, opened 2018-05-28
3388: Added agency to fulltext request

cableman:
When switching to datawell 5.0 the agency is required to make search and the format is obsolete.

### DIFF
--- a/modules/opensearch/opensearch.client.inc
+++ b/modules/opensearch/opensearch.client.inc
@@ -376,7 +376,7 @@ function opensearch_execute_cache($request) {
 
   // Handle fulltext vs. dkabm caching of object.
   $type = OPENSEARCH_CACHE_TING_OBJECT;
-  if ($parms['format'] == 'docbook') {
+  if ($parms['objectFormat'] == 'docbook') {
     $type = OPENSEARCH_CACHE_TING_OBJECT_FULLTEXT;
   }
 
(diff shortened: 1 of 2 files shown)


JrgenGNielsen:
reviewed and accepted

modules/opensearch/opensearch.client.inc
```diff
@@ -376,7 +376,7 @@ function opensearch_execute_cache($request) {
 
   // Handle fulltext vs. dkabm caching of object.
   $type = OPENSEARCH_CACHE_TING_OBJECT;
-  if ($parms['format'] == 'docbook') {
+  if ($parms['objectFormat'] == 'docbook') {
```
kasperg: This change is not mentioned in commit or PR message. I thought this was an unrelated change. The only way I can tell this is needed is because I reread the Redmine issue.

Please remember to document this here as well.

modules/ting_fulltext/ting_fulltext.module
```diff
@@ -48,7 +48,9 @@ function ting_fulltext_object_load($object_id) {
     $request = opensearch_get_request_factory()->getObjectRequest();
     $request->setObjectId($object_id);
     $request->setProfile(variable_get('opensearch_search_profile', ''));
-    $request->setFormat('docbook');
```
kasperg: Same as above.

PR #1106, opened 2018-05-28
2223: Simplified opensearch cache (Test - do not merge - II)

cableman:
This should not be merged.

This is the same PR as #1075 with the make file patched to apply the required patch to ting-client to make the tests run.


cableman:
Not need anymore as code have been merged etc.

PR #1109, opened 2018-05-30
3572: Enable ding_app_variables on existing Opensearch sites

kasperg:
https://platform.dandigbib.org/issues/3572

Previous changes were made to an update hook which has already been
run on production and thus did not have any effect. Consequently we
rerun the hook.


JrgenGNielsen:
reviewed and approved

PR #1110, opened 2018-06-01
3592: Openlist id

BirkAndMe:
https://platform.dandigbib.org/issues/3592

### DIFF
--- a/modules/ding_user/ding_user.module
+++ b/modules/ding_user/ding_user.module
@@ -1102,6 +1102,23 @@ function ding_user_save_creds($auth_res) {
 }
 
 /**
+ * Get a unique id from the provider.
+ */
+function ding_user_provider_id($account = NULL) {
+  global $user;
+
+  if ($account === NULL) {
+    $accout = $user;
+  }
+
+  if (ding_provider_implements('user', 'get_provider_id')) {
+    return ding_provider_invoke('user', 'get_provider_id', $account);
+  }
+
+  return NULL;
+}
+
+/**
  * Returns user login name of the current user, if available.
  *
  * Checks if provider has a get_name function, or has set name in creds.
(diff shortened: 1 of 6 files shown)


kasperg:
#370 has a similar issue (although I am personally not that fond of that solution).

modules/ding_user/ding_user.module
```diff
@@ -1101,6 +1101,23 @@ function ding_user_save_creds($auth_res) {
...
+ */
+function ding_user_provider_id($account = NULL) {
+  global $user;
+
+  if ($account === NULL) {
+    $accout = $user;
+  }
+
+  if (ding_provider_implements('user', 'get_provider_id')) {
+    return ding_provider_invoke('user', 'get_provider_id', $account);
```
kasperg: :+1:

kasperg:
Please implement it for all providers.

I would prefer to be something that developers have to take into account instead of providing a default provider id value which oddly enough is the same default as is used for the name.

BirkAndMe:
I've implemented the function in the 3 providers, but it simply returns the name in each function, I don't know enough about the 3 providers to figure out an identifier. I can look into it, if there's any documentation for the providers and services I can test up against.

I would assume alma is fine, because it used `name` previously, but I don't know about connie and openruth.

kasperg:
@BirkAndMe Thanks for the update.

I appreciate your comment. That's understandable. The current codebase does use `$creds['name']` for all these providers so while this might not fix the situation for all providers it does provide the proper tools for implementing a fix.

ubcudvikler:
There is an error in [ding_user.module:1111](https://github.com/ding2/ding2/commit/8c38619a544733c393ba10c8e8a8efa02f961ec7#diff-31d3885fdca52f895ca91bc19eb76ebfR1111) where it says "$accout = $user;", it should be "$account = $user;"

PR #1111, opened 2018-06-04
3191: better renewal feedback

holt83:
https://platform.dandigbib.org/issues/3191

In this PR I try to provide some better renewal feedback on the loan list.

renewalStatus is returned from FBS API in two cases:

1. When a loan is renewed
2. When a patrons loans is fetched

Until now we have only used the renewalStatus for feedback when the patron renews a loan.

With the changes in this PR the renewalStatus will also be used to provide useful information when the user is viewing the loan list. More information about what exactly is shown on redmine (https://platform.dandigbib.org/issues/3191#note-40).

As we don't have any information about the "renewal-history" of the materials, I have also used the ding_session_cache to remember which loans was renewed in the current session. This should fix the confusing "The material can not be renewed"-message, that was show right after renewal. Instead, it now shows "The material was renewed" for the rest of the session.

I have tried my best to keep all this in the FBS-provider layer, but I have extended the "ding provider loan API" with a new renewalStatus, that providers have to set when building the loan list.

### DIFF
--- a/modules/ding_loan/plugins/content_types/loans.inc
+++ b/modules/ding_loan/plugins/content_types/loans.inc
@@ -274,7 +274,7 @@ function ding_loan_loans_form($form, &$form_state, $account, $loans = array()) {
             'class' => 'messages information',
           );
         }
-        // Inter library loan
+        // Inter library loan.
         elseif ($loan->renewalStatus == DingProviderLoan::STATUS_RENEWAL_REQUESTED) {
           $item['#material_message'] = array(
             'message' => t('An inter-library loan renewal was requested'),
--- a/modules/ding_loan/plugins/content_types/loans.inc
+++ b/modules/ding_loan/plugins/content_types/loans.inc
@@ -168,6 +168,8 @@ function ding_loan_loans_form($form, &$form_state, $account, $loans = array()) {
       $form['loans'][$gid]['title']['#prefix'] = '<a class="anchor" id="anchor-' . $group_type . '"></a>' . $form['loans'][$gid]['title']['#prefix'];
     }
 
+    $renewed_ids = ding_session_cache_get('ding_loan', 'renewed_ids', array());
+
     foreach ($group['loans'] as $loan) {
       // The default title from loan object.
       $title = $loan->display_name;
@@ -255,7 +257,14 @@ function ding_loan_loans_form($form, &$form_state, $account, $loans = array()) {
       }
 
       // If not renewable set message.
-      if (empty($loan->renewable)) {
+      if (in_array($loan->id, $renewed_ids)) {
+        $disabled_count++;
+        $item['#material_message'] = array(
+          'message' => t('This material was renewed'),
+          'class' => 'messages warning',
+        );
+      }
+      elseif (empty($loan->renewable)) {
         $disabled_count++;
         $item['#material_message'] = array(
           'message' => t('This material can not be renewed'),
@@ -332,7 +341,7 @@ function ding_loan_loans_form_submit($form, &$form_state) {
     DingProviderLoan::STATUS_RENEWAL_RESERVED => t('The material is reserved by another loaner'),
   );
 
-  $clear_cache = FALSE;
+  $renewed_ids = array();
 
   $result = ding_provider_invoke('loan', 'renew', $form_state['values']['account'], $ids);
   foreach ($result as $id => $status) {
@@ -362,7 +371,7 @@ function ding_loan_loans_form_submit($form, &$form_state) {
       case DingProviderLoan::STATUS_RENEWED:
       case DingProviderLoan::STATUS_RENEWAL_REQUESTED:
         drupal_set_message(t('"@title renewed"', array('@title' => $title)));
-        $clear_cache = TRUE;
+        $renewed_ids[] = $id;
         break;
 
       case DingProviderLoan::STATUS_NOT_RENEWED:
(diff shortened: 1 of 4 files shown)


modules/fbs/includes/fbs.loan.inc
```diff
@@ -55,6 +55,32 @@ function fbs_loan_list($account, $reset = FALSE) {
...
+        // Try to provide some meaningful renewal feedback.
+        if (empty($loan->isRenewable)) {
+          $renewed_ids = array();
+          if (module_exists('ding_session_cache')) {
+            $renewed_ids = ding_session_cache_get('fbs', 'renewed_ids', array());
+          }
+
+          // We maintain a list of renewed loans in this session, and if this
+          // loan is here we will give it the renewed status for the rest of the
+          // session.
```
kasperg: Why do we want to set which ids was renewed specifically for the FBS provider? I would think this approach would work across providers.

modules/fbs/includes/fbs.loan.inc
```diff
@@ -120,22 +151,37 @@ function fbs_loan_renew($account, $loans) {
...
   }
 
   return $result;
 }
+
+/**
+ * Helper function to extract renewalStatus from the FBS loan and convert it to
+ * to Ding provider renewal status constants.
+ *
+ * Used when renewing and showing loan list.
```
kasperg: Note the Scrutinizer feedback.

Remember to provide a full DocBlock including parameter and return types.

I do not think where a function is used is very helpful. It quickly gets obsolete and most editors will help you find it pretty quickly.

modules/ding_provider/ding_provider.loan.inc
```diff
@@ -106,6 +106,7 @@ class DingProviderLoan extends DingEntityBase {
       'loan_date',
       'expiry',
       'renewable',
+      'renewalStatus',
```
kasperg: Why is this in camelCase when the rest of the properties use snake_case?

modules/ding_loan/plugins/content_types/loans.inc
```diff
@@ -254,13 +254,47 @@ function ding_loan_loans_form($form, &$form_state, $account, $loans = array()) {
...
+          $item['#material_message'] = array(
+            'message' => t('This material was borrowed today'),
+            'class' => 'messages information',
+          );
+        }
+        else {
+          $item['#material_message'] = array(
+            'message' => $error_messages[$loan->renewalStatus],
+            'class' => 'messages warning',
+          );
```
kasperg: I am not too fond of this approach using `$error_messages` as an array.

1. The array is defined pretty far away from where it is used.
2. This checks for `$loan->renewalStatus` but there are already two checks on `$loan->renewalStatus`. I guess they cannot be included here because they use a different class?
3. There is no default message

I wouuld prefer if `$error_messages` and replaced with 2 more checks á la `elseif ($loan->renewalStatus == DingProviderLoan::STATUS_RENEWAL_NUM_EXCEEDED)` with a final default `else` using `This material can not be renewed`.

modules/fbs/includes/fbs.loan.inc
holt83: I was trying to solve a specific problem in FBS-provider, but you're right; this could also be useful in other providers. I'll try to move it up in the layers.

modules/fbs/includes/fbs.loan.inc
holt83: I'll provide a full DocBlock and remove the "used when" comment.

modules/ding_provider/ding_provider.loan.inc
holt83: It should be snake_case. I'll change it.

modules/ding_loan/plugins/content_types/loans.inc
holt83: I'll remove array and add checks and default.

I'm not sure about question in 2. Will look into it.

holt83: On further inspection: I guess I was trying to separate the error_messages from the status messages. But I can see it's a bit confusing, so I'll just remove the array and add checks instead.

holt83:
I updated the PR with suggested changes.

Regarding removing renewed_ids to ding_loan: I have just removed the setting of renewed_ids to ding_loan. It's still up to the providers to use this information when providing renewal_status. 

I think this is the best way to handle this:

1. The provider may want to set a special message based on loan_type (normal or ill). This information is currently not available in ding_loan.
2. The provider may not need the renewed_ids functionality at all and may have even better information about when a loan was renewed from the library-system. This way they can decide whether to use it or not.

modules/ding_loan/plugins/content_types/loans.inc
```diff
@@ -254,13 +254,64 @@ function ding_loan_loans_form($form, &$form_state, $account, $loans = array()) {
...
       }
 
-      // If not renewable set message.
+      // This is the same array as in loan form submit further down, but we
+      // might want different feedback messages on the list.
+      $error_messages = array(
+        DingProviderLoan::STATUS_NOT_RENEWED => t('This material can not be renewed'),
+        DingProviderLoan::STATUS_RENEWAL_NUM_EXCEEDED => t('The material can not be renewed because maximum number of renewals reached'),
+        DingProviderLoan::STATUS_RENEWAL_RESERVED => t('The material is reserved by another loaner and can not be renewed'),
+      );
```
kasperg: This array can be removed now, right?

kasperg:
@holt83: Thanks for the update. I agree with your decisions here.

I have one small change - otherwise this should be good to merge.

PR #1112, opened 2018-06-06
3144: Show to date on ding_event full

b14sha:
https://platform.dandigbib.org/issues/3144


JrgenGNielsen:
reviewed and approved

PR #1114, opened 2018-06-07
3163: Fixed date time interval ding_event

b14sha:
https://platform.dandigbib.org/issues/3163


JrgenGNielsen:
reviewed and approved

PR #1117, opened 2018-06-08
3110: Library list dependant on nodequeue

b14sha:
https://platform.dandigbib.org/issues/3310


JrgenGNielsen:
reviewed and approved

PR #1118, opened 2018-06-10
Core: Revert #3333 changes from release 4.3.x

kasperg:
Release 4.3.2 (and rc2 + rc3) includes commits related to issue [#3333](https://platform.dandigbib.org/issues/3333) from PRs #1072 #1097 and #1108. These should not have been apart of the release as they are targeted for release 4.3.3 instead.

They have been merged erroneously as they are ancestors of other commits targeted for release 4.3.2 because these fixes have originated on branches which were created from `master` after these PRs had been merged into master.

To fix this we revert the merge commits from above PRs. No other tickets have been planned for 4.3.3 so there are no other commits which have been accidentally merged.

To doubke check the approach I have checked the what remains in the log been `release-4.3.x` and the commit where it split from `master`excluding [all tickets planned for release 4.3.2]https://platform.dandigbib.org/versions/88. This results in 3333 commits only.

`git log --format=oneline  5334ead..release-4.3.x --no-merges --invert-grep --grep=3388 --grep=3464 --grep=3466 --grep=3534 --grep=3536 --grep=3537 --grep=3538 --grep=3565 --grep=3566 --grep=3568 --grep=3572`


guddo:
@kasperg Are you aware that these changes is now not a part of `master`? They got lost between `7.x-4.3.4` and `7.x-4.4.0-rc1`. Is that on purpose?

PR #1119, opened 2018-06-11
3302: Added ding_event place/room and organizers

b14sha:
https://platform.dandigbib.org/issues/3302


themes/ddbasic/template.node.php
```diff
@@ -248,8 +282,32 @@ function ddbasic_preprocess__node__ding_event(&$variables) {
...
+              if (!empty($value['url'])) {
+                $organizers_out .= ', ' . l(
+                  $value['title'],
+                  $value['url'],
+                  array('attributes' => array('class' => 'ding-event-organizer-link'))
+                );
+              }
+              else {
+                $organizers_out .= ', ' . $value['title'];
+              }
```
kasperg: See comment about using formatters for displaying these.

PR #1121, opened 2018-06-13
3463: Reset anonymous ratings if not logged in

BirkAndMe:
https://platform.dandigbib.org/issues/3463


modules/p2/ding_entity_rating/js/ding_entity_rating.js
```diff
@@ -68,6 +69,16 @@
       this._trigger( "change" );
     },
 
+    /**
+     * Reset the stars.
+     */
+    reset: function () {
+      console.log(this, 'reset');
```
kasperg: No `console.log` in VCS please. Removed in 18cf2b1.

PR #1123, opened 2018-06-14
3265: Date slider

BirkAndMe:


modules/ding_facetbrowser/ding_facetbrowser.module
```diff
@@ -112,14 +113,138 @@ function ding_facetbrowser_libraries_info() {
...
+        unset($count_search['conditions']['facets'][$delta]);
+      }
+    }
+  }
+
+  // Only run the counter, if it's different from the search already
+  // performed.
+  if ($count_search['run'] === TRUE) {
+    $count_search['conditions']['result_id'] = 'ding_facetbrowser_count_search';
+    search_data($count_search['searchkey'], 'ting_search', $count_search['conditions']);
```
kasperg: From a performance perspective it is unfortunate that this issues a new search - especially if the search engine is slow.

I understand that we want to use the full spectrum. One suggestion could be to store the minimum and maximum values as hidden entries in the form if such values have not been set already. This would ensure that the values persist between requests without having to rerun the search.

modules/ting_search/ting_search.module
```diff
@@ -398,6 +400,7 @@ function ting_search_search_execute($keys = NULL, $conditions = NULL) {
 
       $query = $query->withPage($page + 1)->withCount($results_per_page);
       $search_result = $query->execute();
+      $search_result->conditions = $conditions;
```
kasperg: This bolts an property to an object which is not defined for the class. This is a practice which should be avoided. 

As I understand it we do this to be able to determine which facets were set and with what value. This allows the facet browser to determine if a year facet has been selected. However this data should already be available in `$search_result->getSearchRequest()->getFilters()`.

I see a couple of options for how to deal with this:

1. Pass the filter objects instead of the conditions in `ting_search_ding_facetbrowser`.
2. Change `ding_facetbrowser` implementations to pass the entire `TingSearchResultInterface` object. This already have facets, search key and filters available.
3. Extract the conditions from the URL instead of from arguments in `ding_facetbrowser_form()`. This would follow the current pattern in the current implementation as seen in `_ding_facetbrowser_is_facet_set()`. It is not very elegant though.

If you go with 1 or 2 there is bonus points for also addressing the current implementation in 3 😅 .

modules/ding_facetbrowser/ding_facetbrowser.module
```diff
@@ -90,14 +91,152 @@ function ding_facetbrowser_ctools_plugin_api($module, $api) {
...
+
+  // Only run the counter, if it's different from the search already
+  // performed.
+  if ($run_search === TRUE) {
+    $cache_key = 'ding_facetbrowser:yearspan:' . md5(serialize($count_search));
+    $cache = cache_get($cache_key);
+
+    if (!$cache) {
+      $count_search['conditions']['result_id'] = 'ding_facetbrowser_count_search';
+      search_data($count_search['searchkey'], 'ting_search', $count_search['conditions']);
```
kasperg: Before you read this then please read my comment about `$search_result->conditions` further down:

If we have the search query and the filters available at this point then we can execute the query object directy without using the Drupal wrapper `search_data`. This provides us with greater control of the query which has a number of important benefits:

1. We can reduce the query to not include results (`$query->withCount(0)`) and only include the year facet so we do not spend time and resources retrieving anything we do not need.
2. We can increase the number of terms per facet to a very high number. The default set by `ting_search_search_execute` is 25 and when we only include actual facet values returned in the result in the slider this becomes important if we want to show the entire range.
3. We bypass the `ting_search_current_results()` static cache and can thus the changes in this PR which has do to with introducing result ids.

modules/ding_facetbrowser/ding_facetbrowser.module
```diff
@@ -137,6 +276,14 @@ function ding_facetbrowser_form($form, &$form_state, $facets = [], $search = FAL
           ),
         );
 
+        switch ($facet->getName()) {
+          case 'facet.date':
+            $form[$facet->getName()]['year_span'] = _ding_facetbrowser_form_year_span_element($search, $conditions);
```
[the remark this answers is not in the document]
BirkAndMe: I've gone with 2, but it still checks for facet.date in the `TingSearchFacet` class, any ideas on how to suggest a type for a facet?

modules/ding_facetbrowser/ding_facetbrowser.module
```diff
@@ -90,23 +90,170 @@ function ding_facetbrowser_ctools_plugin_api($module, $api) {
...
+ */
+function _ding_facetbrowser_form_interval_element(TingSearchFacet $facet, TingSearchResultInterface $search_result) {
+  // To get the correct counter for the span, we need to run the
+  // search without the actual interval filter on, so the complete dataset
+  // is included.
+  $run_search = FALSE;
+  $range = array();
+
+  // The checksum is used as a cache key for the count request.
+  $checksum = [
```
BirkAndMe: This feels a bit hacky, but it should work for the current search requests.

Another solution was to add a `__sleep` function to the `TingSearchRequest` class. The problem is all other classes connected to the search result (filters, these filters statements and others) should also implement the `__sleep` function, and keep this function updated when changed in the future, this adds extra maintenance for future features.

modules/ting/src/Search/BooleanStatementGroup.php
```diff
@@ -67,7 +67,7 @@ public function setLogicOperator($logic_operator) {
    * @return TingSearchFieldFilter[]
    *   One or more instances of TingSearchFieldFilter.
    */
-  public function getStatements() {
+  public function &getStatements() {
```
kasperg: Passing the statements by reference is not a good idea. We do not want our search result to mutate.  If we need to modify the filters on a search request then we should retrieve the filters, modify them and then create a new search request using `withFilters()`.

modules/ting/src/Search/TingSearchRequest.php
```diff
@@ -486,7 +486,7 @@ public function withCollectionType($collectionType) {
    * @return FilterStatementInterface[]
    *   List of BooleanStatementGroup instances used to filter field.
    */
-  public function getFilters() {
+  public function &getFilters() {
```
kasperg: See comment regarding `getStatements()`.

modules/ting_field_search/ting_field_search.facets.inc
```diff
@@ -63,6 +63,18 @@ function ting_field_search_form_ding_facetbrowser_form_alter(&$form, &$form_stat
 
       // No need to do anything if the profile facet wasn't returned.
       if ($facet && count($facet->getTerms())) {
+        switch ($name) {
+          case 'facet.date':
```
kasperg: This could be replaced with:

```php
switch ($facet->getType()) {
  case TingSearchFacet::TYPE_INTERVAL:
```

modules/ting_search/ting_search.module
```diff
@@ -444,24 +462,26 @@ function ting_search_search_execute($keys = NULL, $conditions = NULL) {
...
+ *   The search result id.
  * @param \Ting\Search\TingSearchResultInterface $new_search_result
  *   New search result to make globally accessible.
  *
  * @return \Ting\Search\TingSearchResultInterface|NULL
  *   The result or NULL if it is not available.
  */
-function ting_search_current_results(TingSearchResultInterface $new_search_result = NULL) {
-  $global_search_result = &drupal_static('ting_search_results');
+function ting_search_current_results($result_id = TING_SEARCH_DEFAULT_RESULT_ID, TingSearchResultInterface $new_search_result = NULL) {
```
kasperg: Given the latest changes I do not think result ids should be needed anymore? Everywhere it is referenced it now uses the default result id making this complexity unnecessary.

modules/ding_facetbrowser/ding_facetbrowser.module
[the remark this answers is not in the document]
kasperg: I suggest where [the Facet objects are created in the OpenSearch module](https://github.com/ding2/ding2/blob/7c8dee248ce91ca6bc9881590fb249b8d5d06a79/modules/opensearch/src/OpenSearchTingSearchResult.php#L100-L112).

modules/ding_facetbrowser/ding_facetbrowser.module
kasperg: Why do you think you need to implement `__sleep` for `TingSearchRequest`?

I am guessing it is because you wanted to use `serialize` on the request to generate a cache key. That should be fine. Two different objects with the same properties serialize to the same value by default. Using __sleep() also allows you to only serialize a subset of the object's properties. 

Example: https://3v4l.org/oCfC6.

modules/p2/ding_entity_rating/ding_entity_rating.module
```diff
@@ -38,26 +38,72 @@ function ding_entity_rating_menu() {
     'delivery callback' => 'ajax_deliver',
   );
 
-  $items['ding_entity_rating/get'] = array(
-    'title' => 'Ding entity get rating',
-    'page callback' => 'ding_entity_rating_get_user_rating_callback',
+  $items['ding_entity_rating/user/get'] = array(
```
kasperg: All these changes to the rating module are not relevant to the change at hand. Has something gone wrong with a merge or a rebase?

modules/ting/src/Search/TingSearchFacet.php
```diff
@@ -86,7 +106,8 @@ public function setTerms($terms) {
    * @param \Ting\Search\TingSearchFacetTerm $term
    *   The term.
    */
-  public function addTerm($term) {
+  public function addTerm(TingSearchFacetTerm $term) {
```
kasperg: :+1:

modules/ting/src/Search/TingSearchFacet.php
```diff
@@ -34,16 +40,30 @@ class TingSearchFacet {
...
+   * @return string
+   *   The type.
+   */
+  public function getType() {
+    switch ($this->getName()) {
+      case 'facet.date':
+        return self::TYPE_INTERVAL;
+    }
+
+    return self::TYPE_DEFAULT;
```
kasperg: This ties the facets to OpenSearch naming. That is unfortunate. Instead you should pass the type in the constructor just as you do with the name and the terms.

modules/ding_facetbrowser/ding_facetbrowser.module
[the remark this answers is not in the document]
kasperg: Fixed here: https://github.com/ding2/ding2/pull/1440/files#diff-18a52ee5ddca5ff90826d890d337f9a4R279

modules/ting/src/Search/BooleanStatementGroup.php
kasperg: Resolved.

kasperg:
We are continuing the discussion in #1440.

PR #1127, opened 2018-06-18
2175: Support for information about entity buttons

cableman:
In the "guldknap" projekt we need to know which buttons have already been added via hook_ding_entity_buttons. This is need to not add the new button if e.g. the "See online" button have already been added to the entity.

So this PR adds the previous results form the hook to the next module.

Please not that #1075 changes entity_button_parameters, so this PR is dependent on that PR.


cableman:
Need to test the is('online') the background for this change have shifted many times during development.

PR #1128, opened 2018-06-19
3611: Validate german phone numbers

b14sha:
https://platform.dandigbib.org/issues/3611

### DIFF
--- a/modules/fbs/fbs.module
+++ b/modules/fbs/fbs.module
@@ -571,12 +571,18 @@ function fbs_form_profile2_phone_validate($element, $form_state) {
   $phone = $form_state['input']['profile_provider_fbs']['field_fbs_phone']['und'][0]['value'];
 
   if (!empty($phone)) {
-    // Preg match for danish phonenumbers
-    // Pass:
-    // (+45) 35 35 35 35 ||| +45 35 35 35 35 ||| 35 35 35 35 ||| 35353535
-    // Fail:
-    // (45)35353535 ||| 4535353535.
-    $re = '/^((\(?\+45\)?)?)(\s?\d{2}\s?\d{2}\s?\d{2}\s?\d{2})$/';
+    // Preg match for phonenumbers
+    // Allows for countrycode with, and without parenthesis.
+    // Allows for spaces between numberpairs in number
+    // Allows for / and - in number (used in foreign phonenumbers).
+    // First matching group: ([\(]{0,1}[\+][0-9]{1,3}[ \.\-]?[\)]{0,1})
+    // Matches countrycode (with and without parenthesis)
+    // Second matching group: ([\(]{1}[0-9]{1,6}[\)])
+    // Matches local area codes with upto 5 digits eg. (089)
+    // Third matching group: ([0-9 \.\-\/]{3,20})((x|ext|extension)[ ]?[0-9]{1,4})
+    // Matches main phonenumber - allowes for special characters . - / in number.
+    // Also matches phone extensions.
+    $re = '/^([\(]{0,1}[\+][0-9]{1,3}[ \.\-]?[\)]{0,1})?([\(]{1}[0-9]{1,6}[\)])?([0-9 \.\-\/]{3,20})((x|ext|extension)[ ]?[0-9]{1,4})?$/';
     preg_match_all($re, $phone, $matches, PREG_SET_ORDER, 0);
 
     if (empty($matches)) {


modules/fbs/fbs.module
```diff
@@ -571,12 +571,7 @@ function fbs_form_profile2_phone_validate($element, $form_state) {
   $phone = $form_state['input']['profile_provider_fbs']['field_fbs_phone']['und'][0]['value'];
 
   if (!empty($phone)) {
-    // Preg match for danish phonenumbers
-    // Pass:
-    // (+45) 35 35 35 35 ||| +45 35 35 35 35 ||| 35 35 35 35 ||| 35353535
-    // Fail:
-    // (45)35353535 ||| 4535353535.
-    $re = '/^((\(?\+45\)?)?)(\s?\d{2}\s?\d{2}\s?\d{2}\s?\d{2})$/';
+    $re = '/^([\+][0-9]{1,3}[ \.\-]?)?([\(]{1}[0-9]{1,6}[\)])?([0-9 \.\-\/]{3,20})((x|ext|extension)[ ]?[0-9]{1,4})?$/';
```
kasperg: The update to this regular expression increases complexity and drops comment which might explain what is going on. It even 

Please add a code comment which might help future readers.
If this is a regular expression which you found elsewhere the please add a link to the source.
Also note that it fails to match one previously allowed value: `(+45) 35 35 35 35`. That may leave users in a situation where they try to change something else and this now fails.

modules/fbs/fbs.module
```diff
@@ -571,12 +571,23 @@ function fbs_form_profile2_phone_validate($element, $form_state) {
...
+    // (+45) 35 35 35 35 ||| +45 35 35 35 35 ||| 35 35 35 35 |||
+    // (+45)35353535 ||| +4535353535 ||| 35353535
+    // Pass (german phonenumbers):
+    // (+49) 089/12345678 ||| (+49) 089-12345678 ||| +49 089/12345678 ||| +49 089-12345678
+    // +49089/12345678 ||| +49089-12345678 ||| (+49)(089) 12345678
+    // Allows for countrycode with, and without parenthesis.
+    // Allows for spaces between numberpairs in number
+    // Allows for / and - in number (used in german phonenumbers).
+    // First matching group: ([\(]{0,1}[\+][0-9]{1,3}[ \.\-]?[\)]{0,1})
+    // Matches countrycode (with and without parenthesis)
```
kasperg: But this will match almost any country code - not just Denmark and Germany. Is that intentional?

modules/fbs/fbs.module
```diff
@@ -571,12 +571,23 @@ function fbs_form_profile2_phone_validate($element, $form_state) {
...
+    // Second matching group: ([\(]{1}[0-9]{1,6}[\)])
+    // Matches local area codes with upto 5 digits eg. german (089)
```
kasperg: `{1,6}` seems to be up to 6 digits. Why 5 or 6 if Germany only uses 3?

modules/fbs/fbs.module
```diff
@@ -571,12 +571,23 @@ function fbs_form_profile2_phone_validate($element, $form_state) {
...
+    // Third matching group: ([0-9 \.\-\/]{3,20})((x|ext|extension)[ ]?[0-9]{1,4})
+    // Matches main phonenumber - allowes for special characters . - / in number.
+    $re = '/^([\(]{0,1}[\+][0-9]{1,3}[ \.\-]?[\)]{0,1})?([\(]{1}[0-9]{1,6}[\)])?([0-9 \.\-\/]{3,20})((x|ext|extension)[ ]?[0-9]{1,4})?$/';
```
kasperg: Why is the whole extension part relevant?

b14sha: This part is added to validate all phonenumbers, also with local extensions - but maybe could/should be left out in this case? A number like this would be validated: 415-555-4242 ext99. Read more about extensions here: https://getvoip.com/library/difference-between-a-phone-number-an-extension-and-a-line/

modules/fbs/fbs.module
kasperg: If the scope of the task has been updated then please make sure that the PR is updated accordingly.

It is quite confusing that multiple aspects refer to German phone numbers when the implementation is now intentionally not restricted to these two countries. This includes:
- Code comments
- Commit message
- PR title

The examples in the code comments, the commit message and event the PR title all refer to german phone numbers. It is quite co

PR #1130, opened 2018-06-28
2175: Guldknap (ting_das)

cableman:
#### Link to issue

See https://platform.dandigbib.org/issues/2175

#### Description

Support for placing copies at digital article service at the the danish royal library. The service is un-documented but an agreement to test the service by a group of libraries have been accepted.

The service was create long ago for use by bib.dk only.

It adds an "Order digital copy" button to articles that don't already has an "See online" button and if the user is logged in the button is only displayed for users that is resident in the municipality (sa this is another limitation of the condition of usage).

Also note that an requirement (if possible) was that the module use be usable without any other core changes.

#### Screenshot of the result

No screenshot.

#### Checklist

No comments.

### DIFF
--- a/modules/ting_das/css/ting_das.css
+++ b/modules/ting_das/css/ting_das.css
@@ -15,10 +15,6 @@
 }
 
 @media screen and (min-width: 600px) {
-  .popupbar .digital-order-form > div > p {
-    display: none;
-  }
-
   .popupbar .digital-order-form .form-item.form-item-name,.popupbar .digital-order-form .form-item.form-item-mail {
     width:48.5%;
     float: left;
--- a/modules/ting_das/css/ting_das.css
+++ b/modules/ting_das/css/ting_das.css
@@ -71,7 +71,7 @@
     overflow-y: auto !important;
   }
 }
-@media screen and (max-height: 750px) {
+@media screen and (max-height: 750px) and (max-width: 600px) {
   .popupbar {
     height:100% !important;
     overflow-y: auto !important;
--- a/modules/ting_das/css/ting_das.css
+++ b/modules/ting_das/css/ting_das.css
@@ -50,7 +50,8 @@
   .popupbar .digital-order-form .form-submit {
     width: 31.333%;
     float: left;
-    margin-right: 3%
+    margin-right: 3%;
+    margin-bottom: 14px;
   }
 
   .popupbar #popupbar-ting_das .messages {
--- a/modules/ting_das/css/ting_das.css
+++ b/modules/ting_das/css/ting_das.css
@@ -1,0 +1,78 @@
+.popupbar .popupbar-container {
+  margin-bottom: 50px;
+}
+
+.popupbar .popupbar-container .digital-order-form .popupbar-title {
+  padding-top: 40px;
+  font-size: 18px;
+  line-height: 24px;
+  margin-top: -7%;
+  margin-bottom: 20px;
+}
+
+.popupbar .popupbar-container .digital-order-form p {
+  margin-bottom: 10px;
+}
+
+@media screen and (min-width: 600px) {
+  .popupbar .digital-order-form > div > p {
+    display: none;
+  }
+
+  .popupbar .digital-order-form .form-item.form-item-name,.popupbar .digital-order-form .form-item.form-item-mail {
+    width:48.5%;
+    float: left;
+    margin-right: 3%;
+  }
+
+  .popupbar .digital-order-form .form-item.form-item-mail {
+    margin-right: 0;
+  }
+
+  .popupbar .digital-order-form .form-item-fbs-preferred-branch,.popupbar .digital-order-form .form-item-interest-period {
+    width: 48.5%;
+    float: left;
+    margin-right: 3%;
+  }
+
+  .popupbar .digital-order-form .form-item-interest-period {
+    margin-right: 0;
+  }
+
+  .popupbar .digital-order-form .form-type-markup {
+    float: left;
+  }
+
+  .popupbar .digital-order-form .form-type-markup a {
+    margin: 0;
+  }
+
+  .popupbar .digital-order-form .form-submit {
+    width: 31.333%;
+    float: left;
+    margin-right: 3%
+  }
+
+  .popupbar #popupbar-ting_das .messages {
+    margin-bottom: 55px;
+    margin-top: -40px;
+    display: block;
+  }
+
+  .popupbar .digital-order-form .form-item select.error {
+    width: 100%;
+  }
+}
+
+@media screen and (max-width: 600px) {
+  .popupbar {
+    height:90% !important;
+    overflow-y: auto !important;
+  }
+}
+@media screen and (max-height: 750px) {
+  .popupbar {
+    height:100% !important;
+    overflow-y: auto !important;
+  }
+}
--- a/modules/ting_das/includes/ting_das.admin.inc
+++ b/modules/ting_das/includes/ting_das.admin.inc
@@ -81,11 +81,11 @@ function ting_das_admin_settings_form() {
   $defaults = variable_get('ting_das_text', array(
     'description' => array(
       'format' => 'ding_wysiwyg',
-      'value' => '',
+      'value' => _ting_das_default_usage_text('description'),
     ),
     'gdpr' => array(
       'format' => 'ding_wysiwyg',
-      'value' => '',
+      'value' => _ting_das_default_usage_text('gdpr'),
     ),
     'placement' => TING_DAS_PLACEMENT_WEIGHT_ABOVE,
   ));
--- a/modules/ting_das/includes/ting_das.admin.inc
+++ b/modules/ting_das/includes/ting_das.admin.inc
@@ -79,5 +79,16 @@ function ting_das_admin_settings_form() {
     '#default_value' => isset($defaults['gdpr']) ? $defaults['gdpr']['value'] : '',
   );
 
+  $form['ting_das_text']['placement'] = array(
+    '#type' => 'select',
+    '#title' => t('Placement'),
+    '#description' => t('The placement of the text in relation to the "Place order" button.'),
+    '#options' => array(
+      10 => t('Above'),
+      20 => t('Beneath'),
+    ),
+    '#default_value' => isset($defaults['placement']) ? $defaults['placement'] : 10,
+  );
+
   return system_settings_form($form);
 }
--- a/modules/ting_das/includes/ting_das.admin.inc
+++ b/modules/ting_das/includes/ting_das.admin.inc
@@ -24,15 +24,6 @@ function ting_das_admin_settings_form() {
     'url' => '',
     'username' => '',
     'password' => '',
-    'description' => array(
-      'format' => 'ding_wysiwyg',
-      'value' => '',
-    ),
-    'gdpr' => array(
-      'format' => 'ding_wysiwyg',
-      'value' => '',
-    ),
-    'placement' => TING_DAS_PLACEMENT_WEIGHT_ABOVE,
   ));
 
   $form['ting_das']['url'] = array(
@@ -72,10 +63,21 @@ function ting_das_admin_settings_form() {
   $form['ting_das_types']['ting_das_data_well_types'] = array(
     '#type' => 'checkboxes',
     '#options' => drupal_map_assoc($types),
-    '#default_value' => _ting_das_get_default_well_types(),
+    '#default_value' => _ting_das_get_configured_well_types(),
   );
 
-  $defaults = variable_get('ting_das_text', array());
+  $defaults = variable_get('ting_das_text', array(
+    'description' => array(
+      'format' => 'ding_wysiwyg',
+      'value' => '',
+    ),
+    'gdpr' => array(
+      'format' => 'ding_wysiwyg',
+      'value' => '',
+    ),
+    'placement' => TING_DAS_PLACEMENT_WEIGHT_ABOVE,
+  ));
+
   $form['ting_das_text'] = array(
     '#type' => 'fieldset',
     '#title' => t('Digital article server'),
--- a/modules/ting_das/includes/ting_das.admin.inc
+++ b/modules/ting_das/includes/ting_das.admin.inc
@@ -1,0 +1,83 @@
+<?php
+
+/**
+ * @file
+ * Define the administration interface for the module.
+ */
+
+/**
+ * Admin settings form for Ting DAS integration.
+ */
+function ting_das_admin_settings_form() {
+  $form = array();
+
+  $form['ting_das'] = array(
+    '#type' => 'fieldset',
+    '#title' => t('Digital article server'),
+    '#tree' => TRUE,
+  );
+
+  $defaults = variable_get('ting_das', array());
+  $form['ting_das']['url'] = array(
+    '#type' => 'textfield',
+    '#title' => t('End point'),
+    '#description' => t('The service endpoint for "placeCopyRequest", e.g. https://developer.statsbiblioteket.dk/elba-webservices/services/placeCopyRequest'),
+    '#default_value' => isset($defaults['url']) ? $defaults['url'] : '',
+  );
+
+  // Username / password.
+  $form['ting_das']['username'] = array(
+    '#type' => 'textfield',
+    '#title' => t('Username'),
+    '#description' => t('The username required to use the web-service.'),
+    '#default_value' => isset($defaults['username']) ? $defaults['username'] : '',
+  );
+
+  $form['ting_das']['password'] = array(
+    '#type' => 'textfield',
+    '#title' => t('Password'),
+    '#description' => t('The password required to use the web-service.'),
+    '#default_value' => isset($defaults['password']) ? $defaults['password'] : '',
+  );
+
+  $types = variable_get('ting_well_types', _ting_fetch_well_types());
+  $form['ting_das_types'] = array(
+    '#type' => 'fieldset',
+    '#title' => t('Data well types'),
+    '#description' => t('Ting objects defined as being able order via the DAS service.'),
+    '#collapsible' => TRUE,
+    '#collapsed' => TRUE,
+  );
+
+  $form['ting_das_types']['ting_das_data_well_types'] = array(
+    '#type' => 'checkboxes',
+    '#options' => drupal_map_assoc($types),
+    '#default_value' => _ting_das_get_default_well_types(),
+  );
+
+  $defaults = variable_get('ting_das_text', array());
+  $form['ting_das_text'] = array(
+    '#type' => 'fieldset',
+    '#title' => t('Digital article server'),
+    '#description' => t('Information to the user about the usage for the data sent and handled by the DAS service.'),
+    '#tree' => TRUE,
+  );
+
+  $form['ting_das_text']['description'] = array(
+    '#type' => 'text_format',
+    '#title' => t('Description about the form (prefix) or service'),
+    '#rows' => 8,
+    '#format' => isset($defaults['description']) ? $defaults['description']['format'] : NULL,
+    '#default_value' => isset($defaults['description']) ? $defaults['description']['value'] : '',
+  );
+
+  $form['ting_das_text']['gdpr'] = array(
+    '#type' => 'text_format',
+    '#title' => t('Description of data usage (GDPR)'),
+    '#rows' => 8,
+    '#format' => isset($defaults['gdpr']) ? $defaults['gdpr']['format'] : NULL,
+    '#default_value' => isset($defaults['gdpr']) ? $defaults['gdpr']['value'] : '',
+  );
+
+  return system_settings_form($form);
+}
--- a/modules/ting_das/includes/ting_das.admin.inc
+++ b/modules/ting_das/includes/ting_das.admin.inc
@@ -109,7 +109,7 @@ function ting_das_admin_settings_form() {
       TING_DAS_PLACEMENT_WEIGHT_ABOVE => t('Above'),
       TING_DAS_PLACEMENT_WEIGHT_BELOW => t('Below'),
     ),
-    '#default_value' => isset($defaults['placement']) ? $defaults['placement'] : 10,
+    '#default_value' => isset($defaults['placement']) ? $defaults['placement'] : TING_DAS_PLACEMENT_WEIGHT_ABOVE,
   );
 
   return system_settings_form($form);
--- a/modules/ting_das/includes/ting_das.admin.inc
+++ b/modules/ting_das/includes/ting_das.admin.inc
@@ -5,6 +5,9 @@
  * Define the administration interface for the module.
  */
 
+define('TING_DAS_PLACEMENT_WEIGHT_ABOVE', 10);
+define('TING_DAS_PLACEMENT_WEIGHT_BELOW', 20);
+
 /**
  * Admin settings form for Ting DAS integration.
  */
@@ -17,12 +20,27 @@ function ting_das_admin_settings_form() {
     '#tree' => TRUE,
   );
 
-  $defaults = variable_get('ting_das', array());
+  $defaults = variable_get('ting_das', array(
+    'url' => '',
+    'username' => '',
+    'password' => '',
+    'description' => array(
+      'format' => 'ding_wysiwyg',
+      'value' => '',
+    ),
+    'gdpr' => array(
+      'format' => 'ding_wysiwyg',
+      'value' => '',
+    ),
+    'placement' => TING_DAS_PLACEMENT_WEIGHT_ABOVE,
+  ));
+
   $form['ting_das']['url'] = array(
     '#type' => 'textfield',
     '#title' => t('End point'),
     '#description' => t('The service endpoint for "placeCopyRequest", e.g. https://developer.statsbiblioteket.dk/elba-webservices/services/placeCopyRequest'),
-    '#default_value' => isset($defaults['url']) ? $defaults['url'] : '',
+    '#default_value' => $defaults['url'],
+    '#required' => TRUE,
   );
 
   // Username / password.
@@ -30,21 +48,23 @@ function ting_das_admin_settings_form() {
     '#type' => 'textfield',
     '#title' => t('Username'),
     '#description' => t('The username required to use the web-service.'),
-    '#default_value' => isset($defaults['username']) ? $defaults['username'] : '',
+    '#default_value' => $defaults['username'],
+    '#required' => TRUE,
   );
 
   $form['ting_das']['password'] = array(
     '#type' => 'textfield',
     '#title' => t('Password'),
     '#description' => t('The password required to use the web-service.'),
-    '#default_value' => isset($defaults['password']) ? $defaults['password'] : '',
+    '#default_value' => $defaults['password'],
+    '#required' => TRUE,
   );
 
   $types = variable_get('ting_well_types', _ting_fetch_well_types());
   $form['ting_das_types'] = array(
     '#type' => 'fieldset',
     '#title' => t('Data well types'),
-    '#description' => t('Ting objects defined as being able order via the DAS service.'),
+    '#description' => t('Ting objects defined as being to able order via the DAS service.'),
     '#collapsible' => TRUE,
     '#collapsed' => TRUE,
   );
@@ -67,16 +87,16 @@ function ting_das_admin_settings_form() {
     '#type' => 'text_format',
     '#title' => t('Description about the form (prefix) or service'),
     '#rows' => 8,
-    '#format' => isset($defaults['description']) ? $defaults['description']['format'] : NULL,
-    '#default_value' => isset($defaults['description']) ? $defaults['description']['value'] : '',
+    '#format' => $defaults['description']['format'],
+    '#default_value' => $defaults['description']['value'],
   );
 
   $form['ting_das_text']['gdpr'] = array(
     '#type' => 'text_format',
     '#title' => t('Description of data usage (GDPR)'),
     '#rows' => 8,
-    '#format' => isset($defaults['gdpr']) ? $defaults['gdpr']['format'] : NULL,
-    '#default_value' => isset($defaults['gdpr']) ? $defaults['gdpr']['value'] : '',
+    '#format' => $defaults['gdpr']['format'],
+    '#default_value' => $defaults['gdpr']['value'],
   );
 
   $form['ting_das_text']['placement'] = array(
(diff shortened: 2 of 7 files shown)


cableman:
The Scrutinizer error about type hint string should not be an error: http://php.net/manual/en/functions.arguments.php#functions.arguments.type-declaration

modules/ting_das/includes/ting_das.admin.inc
```diff
@@ -0,0 +1,94 @@
...
+    '#tree' => TRUE,
+  );
+
+  $defaults = variable_get('ting_das', array());
+  $form['ting_das']['url'] = array(
+    '#type' => 'textfield',
+    '#title' => t('End point'),
+    '#description' => t('The service endpoint for "placeCopyRequest", e.g. https://developer.statsbiblioteket.dk/elba-webservices/services/placeCopyRequest'),
+    '#default_value' => isset($defaults['url']) ? $defaults['url'] : '',
+  );
```
kasperg: Idea: Drop the issets for the default value and place default values in the default array if `ting_das` is not set.

Example:

```php
$defaults = variable_get('ting_das', array(
  'url' => '')
);
$form['ting_das']['url'] = array(
    '#type' => 'textfield',
    '#title' => t('End point'),
    '#description' => t('The service endpoint for "placeCopyRequest", e.g. https://developer.statsbiblioteket.dk/elba-webservices/services/placeCopyRequest'),
    '#default_value' => $defaults['url'],
);

File: modules/ting_das/includes/ting_das.admin.inc [hunk 2]
```diff
@@ -0,0 +1,94 @@
...
+    '#title' => t('Password'),
+    '#description' => t('The password required to use the web-service.'),
+    '#default_value' => isset($defaults['password']) ? $defaults['password'] : '',
+  );
+
+  $types = variable_get('ting_well_types', _ting_fetch_well_types());
+  $form['ting_das_types'] = array(
+    '#type' => 'fieldset',
+    '#title' => t('Data well types'),
+    '#description' => t('Ting objects defined as being able order via the DAS service.'),
```
kasperg:
```suggestion
    '#description' => t('Ting objects defined as being able to order via the DAS service.'),
```

modules/ting_das/includes/ting_das.admin.inc
```diff
@@ -0,0 +1,94 @@
...
+    '#default_value' => isset($defaults['gdpr']) ? $defaults['gdpr']['value'] : '',
+  );
+
+  $form['ting_das_text']['placement'] = array(
+    '#type' => 'select',
+    '#title' => t('Placement'),
+    '#description' => t('The placement of the text in relation to the "Place order" button.'),
+    '#options' => array(
+      10 => t('Above'),
+      20 => t('Beneath'),
```
kasperg: Please use proper constants instead of magic values.

modules/ting_das/includes/ting_das.client.inc
```diff
@@ -0,0 +1,430 @@
...
+   *
+   * The date until the user is not longer interested in getting the material.
+   *
+   * @param string $userInterestDays
+   *   The users interest date in number of days.
+   *
+   * @return TingDasCopyRequest
+   *   The current request object.
+   */
+  public function setUserInterestDate(string $userInterestDays) {
```
kasperg: If this accepts a number of days then `int` seems more appropriate than `string`?

modules/ting_das/includes/ting_das.client.inc
```diff
@@ -0,0 +1,430 @@
...
+   *
+   * @param string $format
+   *   The date format to use.
+   *
+   * @return string|null
+   *   The users interest date in the format given from now.
+   */
+  public function getUserInterestDate($format = 'Y-m-d') {
+    $date = strtotime('+' . $this->userInterestDays . ' day');
+    return date($format, $date);
```
kasperg: How can this return `null`?

Personally I would prefer to se a `DateTime` object returned.

modules/ting_das/includes/ting_das.client.inc
```diff
@@ -0,0 +1,430 @@
...
+  }
+
+  /**
+   * Set the agency id for pickup.
+   *
+   * @param string $pickupAgencyId
+   *   The pickup agency id.
+   *
+   * @Note: Not sure why this exists as all articles are sent by mail to the
+   *        end user.
```
kasperg: If we do not use it then I think it should be removed along with the getter. If it is needed when sending requests we can keep it internally.

modules/ting_das/includes/ting_das.client.inc
```diff
@@ -0,0 +1,430 @@
...
+   *
+   * @return string|null
+   *   The article title.
+   */
+  public function getPublicationTitle() {
+    return $this->publicationTitle;
+  }
+
+  /**
+   * Set the date of publication.
```
kasperg: I think this warrants some explanation regarding what the date of the publication means. Especially since we also have a publication year.

We could let clients pass a `DateTime` object and just extract date + year internally if we need to send it as separate values.

modules/ting_das/includes/ting_das.client.inc
```diff
@@ -0,0 +1,430 @@
...
+  /**
+   * Set the page number in the publication.
+   *
+   * @param string $pagesOfComponent
+   *   The page number.
+   *
+   * @return TingDasCopyRequest
+   *   The current request object.
+   */
+  public function setPagesOfComponent(string $pagesOfComponent) {
```
kasperg: Should this be an integer?

modules/ting_das/includes/ting_das.client.inc
```diff
@@ -0,0 +1,430 @@
...
+  public function send() {
+    $options = array(
+      'headers' => array(
+        'Content-Type' => 'application/xml; charset=UTF8',
+      ),
+      'body' => $this->buildMessage(),
+    );
+
+    try {
+      $client = new Client();
```
kasperg: In my opinion the client should be injected into the class - not created internally. This prevents unit testing of the class.

modules/ting_das/includes/ting_das.client.inc
```diff
@@ -0,0 +1,430 @@
...
+      $response = $client->request('POST', $this->url, $options);
+
+      if ($response->getStatusCode() == 204) {
+        watchdog('ting_das', 'Service returned: !code', array('!code' => $response->getStatusCode()), WATCHDOG_ERROR);
+        return TRUE;
```
kasperg: Why do we check for status code 204 explicitly? It might be what the service currently returns if things go well but there is no handling of other success status codes even though they might be just as valid. Adding to this the method also returns `FALSE` in this situation indicating to the caller that the request was not sent.

[Guzzle throws an exception for all non-200 status codes and other errors](http://docs.guzzlephp.org/en/stable/quickstart.html) so such situations are well handled by the remaining code.

Regarding the logging also note:
- You have a nice non-Drupal specific class here. Consider injecting a PSR3 logger interface instead of calling `watchdog` directly and then use an adapter such as [`drupol/drupal7_psr3_watchdog`](https://packagist.org/packages/drupol/drupal7_psr3_watchdog).
- The current code sets log level `WATCHDOG_ERROR` for a successful call. That does not seem right.
- It can be extremely helpful to have an option which when enabled logs outgoing requests and responses. Consider adding that.

modules/ting_das/ting_das.info
```diff
@@ -0,0 +1,8 @@
+name = Ting digital article service
+description = Use DAS from KB to order articles.
```
kasperg: This seems a bit internal (KB, DAS etc.). Consider dropping acronyms and elaborating.

modules/ting_das/includes/ting_das.client.inc
```diff
@@ -0,0 +1,430 @@
...
+    $xml = new SimpleXMLElement('<?xml version="1.0" encoding="utf-8"?><placeCopyRequest xmlns="http://statsbiblioteket.dk/xws/elba-placecopyrequest-schema"/>');
+
+    // Set required fields.
+    $xml->addChild('ws_user', $this->wsUsername);
+    $xml->addChild('ws_password', $this->wsPassword);
+    $xml->addChild('pid', $this->getPid());
+    $xml->addChild('user_loaner_id', $this->getUserLoanerId());
+    $xml->addChild('userName', $this->getUserName());
+    $xml->addChild('userMail', $this->getUserMail());
+    $xml->addChild('user_interest_date', $this->getUserInterestDate('Y-m-d'));
```
kasperg: Snake + camel casing are all over the place 🙄 .

modules/ting_das/includes/ting_das.client.inc
```diff
@@ -0,0 +1,430 @@
...
+    return $this;
+  }
+
+  /**
+   * The data well id.
+   *
+   * @return string|null
+   *   The data well pid if set else NULL.
+   */
+  public function getPid() {
```
kasperg: All getters in this class are only called internally where you might as well use the properties. If that is the case then you might as well drop the accessors as they have little value.

modules/ting_das/ting_das.install
```diff
@@ -0,0 +1,13 @@
...
+/**
+ * @file
+ * Handle the modules basic configuration.
+ */
+
+/**
+ * Implements hook_uninstall().
+ */
+function ting_das_uninstall() {
+  variable_del('ting_das');
```
kasperg: :+1: but remember `ting_das_text`.

modules/ting_das/ting_das.module
```diff
@@ -0,0 +1,378 @@
+<?php
+
+/**
+ * @file
+ * Handle the integration of digital article service (DAS) into the CMS.
```
kasperg: Do we have a reference to the documentation for this service? A link or so would be nice.

modules/ting_das/includes/ting_das.admin.inc
kasperg: Also: Should this - and other values be `#required`. The module will not work with a partial set configuration.

modules/ting_das/ting_das.module
```diff
@@ -0,0 +1,378 @@
...
+}
+
+/**
+ * Implements hook_libraries_info().
+ *
+ * For defining external libraries.
+ */
+function ting_das_libraries_info() {
+  return array(
+    'guzzle' => array(
```
kasperg: On its own this is not sufficient to use Guzzle. As I see it the preferred approach going forward is to depend on `ding_libs` (see c3e8bd0f46c0167fcc94ef5dd5ce260b36052497).

modules/ting_das/ting_das.module
```diff
@@ -0,0 +1,378 @@
...
+    'page arguments' => array('ting_das_admin_settings_form'),
+    'access arguments' => array('administer site configuration'),
+    'file' => 'includes/ting_das.admin.inc',
+  );
+
+  $items['das/%ting_object/order'] = array(
+    'page callback' => 'ting_das_ajax',
+    'page arguments' => array(1),
+    'delivery callback' => 'ajax_deliver',
+    'access arguments' => array('access content'),
```
kasperg: Is this really the permission we want to use here?

modules/ting_das/ting_das.module
```diff
@@ -0,0 +1,378 @@
...
+ */
+function ting_das_menu() {
+  $items = array();
+
+  $items['admin/config/ting/das'] = array(
+    'title' => 'Digital article service',
+    'description' => 'Configure access to DAS.',
+    'page callback' => 'drupal_get_form',
+    'page arguments' => array('ting_das_admin_settings_form'),
+    'access arguments' => array('administer site configuration'),
```
kasperg: Perhaps a little bit lazy here compared to implement our own permissions?

modules/ting_das/ting_das.module
```diff
@@ -0,0 +1,378 @@
...
+
+    $display_button = TRUE;
+    if (user_is_logged_in()) {
+      try {
+        $creds = ding_user_get_creds();
+        $display_button = isset($creds['resident']) && $creds['resident'];
+      }
+      catch (DingProviderAuthException $exception) {
+        // This is okay as the use simply was not logged in as an provider
+        // user. So do noting in this case.
```
kasperg: :+1:

modules/ting_das/ting_das.module
```diff
@@ -0,0 +1,378 @@
...
+ */
+function ting_das_ding_entity_buttons($type, $entity, $view_mode, $widget = 'default') {
+  $type = drupal_strtolower($entity->getType());
+  $types = array_filter(_ting_das_get_default_well_types());
+
+  // Check that the entity is not already available as an online resources and is one of the configured types.
+  if (isset($types[$type]) && !$entity->is('online') && !$entity->is('infomedia')) {
+    drupal_add_library('system', 'drupal.ajax');
+
+    $display_button = TRUE;
```
kasperg: Do we want to display the button by default and then hide it for some users? That seems a bit odd.

modules/ting_das/ting_das.module
```diff
@@ -0,0 +1,378 @@
...
+          '#path' => 'das/' . $entity->id . '/order',
+          '#options' => array(
+            'attributes' => array(
+              'class' => array(
+                'action-button',
+                'button-see-online',
+                'use-ajax',
+              ),
+              'id' => 'das-order-' . $entity->id,
+              'onclick' => 'wt.sendinfo({ linkId: "open_order_digital_copy" });',
```
kasperg: This creates a hard (and undeclared) dependency on `ding_webtreds`. That does not seem right?

modules/ting_das/ting_das.module
```diff
@@ -0,0 +1,378 @@
...
+            ),
+            'html' => FALSE,
+          ),
+          '#attached' => array(
+            'css' => array(
+              drupal_get_path('module', 'ting_das') . '/css/ting_das.css',
```
kasperg: As far as I can tell the content in `ting_das` does not style the button at all. Perhaps it should be returned as a part of the AJAX response?

modules/ting_das/ting_das.module
```diff
@@ -0,0 +1,378 @@
...
+  $commands = array();
+
+  // Check if the logged in user is a library user.
+  global $user;
+  if (!user_is_logged_in()) {
+    $commands[] = ajax_command_ding_user_authenticate('');
+  }
+  elseif (!ding_user_is_provider_user($user)) {
+    // Error not library user.
+    $commands[] = ajax_command_ding_popup('ting_das', t('Error'), '<p>' . t('Only library user can make DAS requests.') . '</p>', array('refresh' => FALSE));
```
kasperg: I doubt end users will know what a "DAS request" is.

modules/ting_das/ting_das.module
```diff
@@ -0,0 +1,378 @@
...
+  }
+  elseif (!(is_object($entity) && $entity instanceof TingEntity)) {
```
kasperg: How does this conditional make sense? It is triggered if the passed variable is not an object *and* is an instance og `TingEntity`.

modules/ting_das/ting_das.module
```diff
@@ -0,0 +1,378 @@
...
+      '#default_value' => $profile ? $profile->{'field_' . $provider['module'] . '_mail'}->value() : '',
+      '#element_validate' => array(
+        'ting_das_mail_validate',
+      ),
+      '#required' => TRUE,
+    );
+
+    $function = $provider['module'] . '_user_branches';
+    $preferred_branch = $profile ? $profile->{'field_' . $provider['module'] . '_preferred_branch'}->value() : '';
+    $form += ding_reservation_default_options_branch('fbs_preferred_branch', $preferred_branch, $function($user));
```
kasperg: It seems as if this creates a dependency on `fbs`? That may be OK but it is not declared.

modules/ting_das/ting_das.module
```diff
@@ -0,0 +1,378 @@
...
+
+    $form['submit'] = array(
+      '#type' => 'submit',
+      '#submit' => array('ting_das_order_form_submit'),
+      '#ajax' => array(
+        'callback' => 'ting_das_order_form_ajax_callback',
+      ),
+      '#value' => t('Order digital copy'),
+      '#attributes' => array(
+        'onclick' => 'wt.sendinfo({ linkId: "order_digital_copy" });',
```
kasperg: Again: Note the dependency on `ding_webtreds`.

modules/ting_das/ting_das.module
```diff
@@ -0,0 +1,378 @@
...
+    );
+  }
+  else {
+    $form['title'] = array(
+      '#markup' => '<div class="popupbar-title"><strong>' . t('Not resident in the municipality') . '</strong></div>',
+    );
+
+    $form['description'] = array(
+      '#prefix' => '<div class="form-item form-type-markup description">',
+      '#markup' => t('Only resident in the municipality is allowed to order digital copies.'),
```
kasperg:
```suggestion
      '#markup' => t('Only residents in the municipality are allowed to order digital copies.'),
```

modules/ting_das/ting_das.module
```diff
@@ -0,0 +1,378 @@
...
+    }
+  }
+}
+
+/**
+ * Submit handler for the order form.
+ */
+function ting_das_order_form_submit($form, &$form_state) {
+
+}
```
kasperg: This is empty. If that is intentional then please state that.

modules/ting_das/ting_das.module
```diff
@@ -0,0 +1,378 @@
...
+  static $client;
+
+  if (!isset($client)) {
+    module_load_include('inc', 'ting_das', 'includes/ting_das.clint');
+    $config = variable_get('ting_das', FALSE);
+    if (is_array($config)) {
+      $client = new TingDasCopyRequest($config['url'], $config['username'], $config['password']);
+    }
+    else {
+      watchdog('ting_das', 'The ting_das module is missing its configuration', NULL, WATCHDOG_ERROR);
```
kasperg: This has the risk of returning `NULL` and not a `TingDasCopyRequest` object.

modules/ting_das/ting_das.module
```diff
@@ -0,0 +1,378 @@
...
+
+/**
+ * Get the default material types that can be ordered in DAS.
+ *
+ * @return string[]
+ *   The types
+ */
+function _ting_das_get_default_well_types() {
+  return variable_get('ting_das_data_well_types', array(
+    'tidsskriftsartikel',
```
kasperg: This does not return defaults as indicated by the name of the function and the documentation.

kasperg:
Also note failing Scrutinizer and CircleCI tests.

modules/ting_das/includes/ting_das.admin.inc
cableman: Required added and isset removed and defaults moved to... well defaults array

modules/ting_das/includes/ting_das.admin.inc
cableman: Fixed the string with suggested change.

modules/ting_das/includes/ting_das.admin.inc
cableman: Change to constants.

modules/ting_das/includes/ting_das.client.inc
cableman: It should have been != 204. But the service only have to stats 204 = accepted and 400 which it only throws when the XML message is not proper.

modules/ting_das/includes/ting_das.client.inc
cableman: That's the DAS service not something that related to this PR.

modules/ting_das/ting_das.module
cableman: yes, I known... ding_libs did not exits when I made this module 6-7 months ago. Will change it to use ding_libs.

modules/ting_das/ting_das.module
cableman: Why should we want to implement our own permission, when this is exactly what I want?

modules/ting_das/ting_das.module
cableman: It's odd... but that is what is wanted. Always show the button when user not logged in and hide it if the user logges in and is not resident in the municipality.

modules/ting_das/ting_das.module
cableman: Added as dependency in the info file.

modules/ting_das/ting_das.module
cableman: It styles all content inside the "popup-bar" that drops down when the button is pressed. As I understands it, it's not possible to load it in an drupal ajax request as it will not add css to the header dynamically.

The "ajax_command_css($selector, $argument)" is only for adding inline css and I don't think that we should move the style from file into the module code to x linjes of code.

modules/ting_das/ting_das.module
cableman: Changes, but it's only to users that is logged in as non-library users.

modules/ting_das/ting_das.module
cableman: I see... I must confess that I toke the logic from https://github.com/ding2/ding2/blob/master/modules/ding_reservation/ding_reservation.module#L178 which then has the error as well.

:-D

cableman: But is not !(is object && instanceof ting entity) and that's ?

modules/ting_das/ting_das.module
cableman: Added as dependency in info file now. It's required as KB wants the PID in the request sent to them. So as I see it the dependency is there.

modules/ting_das/ting_das.module
cableman: Is not added to the info file.

modules/ting_das/ting_das.module
cableman: Changed.

modules/ting_das/ting_das.module
cableman: Removed.

modules/ting_das/ting_das.module
cableman: Added that to documentation and added code to handle it.

modules/ting_das/ting_das.module
cableman: Function renamed.

modules/ting_das/ting_das.info
cableman: Fixed.

modules/ting_das/includes/ting_das.client.inc
cableman: As the service is not documented I did not known at the time why. It turns out that do to copyright issues an article may be delivered to an location (pickup agency) as a paper copy.

So comment has been removed.

modules/ting_das/includes/ting_das.client.inc
cableman: Changed to FALSE as "date" may return that if not able to parse the date.

cableman: For this client I don't think the DateTime is required for this very simple handling.

modules/ting_das/includes/ting_das.client.inc
cableman: As the service is un-documented all the getters are here for completeness nothing else. The scope and usage for the service in the feature is unknown.

I you insist I'll remove them?

modules/ting_das/includes/ting_das.client.inc
cableman: Changed.

modules/ting_das/includes/ting_das.client.inc
cableman: Changed.

modules/ting_das/css/ting_das.css
```diff
@@ -0,0 +1,79 @@
+.popupbar .popupbar-container {
```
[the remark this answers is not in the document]
cableman: @yepzdk will you comment on this.

modules/ting_das/css/ting_das.css
```diff
@@ -0,0 +1,79 @@
...
+  margin-bottom: 20px;
+}
+
+.popupbar .popupbar-container .digital-order-form p {
+  margin-bottom: 10px;
+}
+
+@media screen and (min-width: 600px) {
+  .popupbar .digital-order-form > div > p {
+    display: none;
```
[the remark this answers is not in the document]
cableman: @yepzdk will you comment on this.

modules/ting_das/ting_das.module
cableman: Added custom access callback function.

modules/ting_das/includes/ting_das.client.inc
cableman: Comment and and the only logger is the exception logger. I'am not sure if this request the effort to load yet another library into the site.

modules/ting_das/includes/ting_das.client.inc
cableman: Done

modules/ting_das/includes/ting_das.client.inc
cableman: The service is un-documented so your guess is as good as mine.

It's here as it's possible to set it.

modules/ting_das/css/ting_das.css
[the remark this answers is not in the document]
yepzdk: You are right. We would prefer to use BEM/SMACSS. But in this situation we are reusing the popupbar component and want it to behave different in some aspects, we needed to override some of the already existing styling and therefore be more specific to ensure the overrides.

modules/ting_das/css/ting_das.css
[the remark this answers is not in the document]
yepzdk: We should absolutely have left a code comment with this. I'll track back and document it.

modules/ting_das/ting_das.module
kasperg: If we do not care about permissions we might as well lock it down to roles instead if that is exactly what we want. Perhaps that is less important.

modules/ting_das/ting_das.info
```diff
@@ -0,0 +1,10 @@
+name = Ting digital article service
+description = Use Digital Articel Service from the Royal Danish Library to order copies of articles.
```
kasperg: Nitpicking (but it is in the module description after all): Articel = Article.

modules/ting_das/includes/ting_das.client.inc
```diff
@@ -0,0 +1,431 @@
...
+
+      // We don't look at the response but assumes that everything is okay. This
+      // is do to the fact that the service at current time only returns 204
+      // regardless of what was sent as long as the XML message is correct. If
+      // the message is wrong a non 200 code will be returned, which will
+      // trigger and guzzle exception (handled below).
+      return TRUE;
+    }
+    catch (GuzzleException $exception) {
+      watchdog_exception('ting_das', $exception);
```
kasperg: This code has been updated but for some reason GitHub will not let me continue the discussion of this. Anyhow:

Not having Drupal specific code in our libraries is a general concern. For logging it would be useful across all our clients. Adding a library for handling this will have to start somewhere. It might as well be here.

If we do not want to add the library then we could at least just not do error handling internally and just let `send` throw an exception. Then the calling Drupal code could catch and call `watchdog_exception()` accordingly.

kasperg:
Regarding the Scrutinizer errors they are in two parts:

`ting_das_ajax($entity)`
The type hint should be `TingEntity` if that is what the function requires. We do our own type checking further down in the function so it seems as if we are actually less strict about typing. If that is the case then we should change the type to something less strict.

Since the type of our argument depends on the value of [`ting_object_load()`](https://github.com/ding2/ding2/blob/e586ef8f094742828d5e7fd07e3028fbc4cee623/modules/ting/ting.module#L502) it seems as if we should actually define the type in the DocBlock as `TingEntity|boolean` and then keep our own type check.

`Unknown type hint "string" found for $param`
[This seems to be an issue with the Drupal Coding standards](https://github.com/squizlabs/PHP_CodeSniffer/issues/805#issuecomment-279388383) when using primitive type hints in PHP 7. I am not sure what the right way to go about this is. I cannot spot this error message in [the latest version of the Coding Standards for Drupal 7](https://cgit.drupalcode.org/coder/tree/coder_sniffer/Drupal/Sniffs/Commenting/FunctionCommentSniff.php?h=7.x-2.x). Perhaps we need to tune our Scrutinizer setup to use a more specific version.

modules/ting_das/css/ting_das.css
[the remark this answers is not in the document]
yepzdk: A refactor of the existing component is out of scope for this. And for the same reason we are working in plain css.

It could probably be done, but then we would need to use the !important statement even more.
So I preferred to keep .popupbar in front to be explicit enough for our overides to work.

modules/ting_das/includes/ting_das.admin.inc
```diff
@@ -0,0 +1,116 @@
...
+
+  $form['ting_das_text']['placement'] = array(
+    '#type' => 'select',
+    '#title' => t('Placement'),
+    '#description' => t('The placement of the text in relation to the "Place order" button.'),
+    '#options' => array(
+      TING_DAS_PLACEMENT_WEIGHT_ABOVE => t('Above'),
+      TING_DAS_PLACEMENT_WEIGHT_BELOW => t('Below'),
+    ),
+    '#default_value' => isset($defaults['placement']) ? $defaults['placement'] : 10,
```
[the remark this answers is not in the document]
cableman: Fixed

modules/ting_das/ting_das.module
```diff
@@ -0,0 +1,388 @@
...
+ * Get DAS client.
+ *
+ * @return \TingDasCopyRequest|NULL
+ *   Client to communicate with the service or NULL if not configured.
+ */
+function _ting_das_get_client() {
+  static $client;
+
+  if (!isset($client)) {
+    module_load_include('inc', 'ting_das', 'includes/ting_das.clint');
```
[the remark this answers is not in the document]
cableman: Fixed

modules/ting_das/ting_das.info
cableman: Fixed

modules/ting_das/includes/ting_das.client.inc
cableman: fixed

GitCop:

 * Commit: 9146d66614fbb152e82cc2032fa1401ecd6ca1ea

---

GitCop:

 * Commit: 9146d66614fbb152e82cc2032fa1401ecd6ca1ea

---

GitCop:

 * Commit: 9146d66614fbb152e82cc2032fa1401ecd6ca1ea

---

GitCop:

 * Commit: 9146d66614fbb152e82cc2032fa1401ecd6ca1ea

---

GitCop:

 * Commit: 9146d66614fbb152e82cc2032fa1401ecd6ca1ea

---

GitCop:

 * Commit: 9146d66614fbb152e82cc2032fa1401ecd6ca1ea

---

GitCop:

 * Commit: 9146d66614fbb152e82cc2032fa1401ecd6ca1ea

---

GitCop:

 * Commit: 9146d66614fbb152e82cc2032fa1401ecd6ca1ea

---

PR #1133, opened 2018-07-03
3388: Changed fulltext XML check from format to objectFormat

cableman:
The detection of the format for response from the datawell has changed to be objectFormat from format, which did break the fulltext module.


kasperg:
I resolved the merge conflict here manually.

PR #1135, opened 2018-07-10
3684: Show add to lists for anonymous users

BirkAndMe:
https://platform.dandigbib.org/issues/3684

### DIFF
--- a/modules/p2/ding_list/theme/theme.inc
+++ b/modules/p2/ding_list/theme/theme.inc
@@ -269,8 +269,6 @@ function template_preprocess_ding_list_add_button(&$variables) {
     ),
   );
 
-  $variables['single_link'] = FALSE;
-
   if ($variables['options'] === FALSE) {
     $options = ding_list_get_add_to_list_data($variables['type'], $variables['value'], $variables['value_title']);
   }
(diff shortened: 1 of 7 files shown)


modules/p2/ding_list/theme/theme.inc
```diff
@@ -269,21 +267,33 @@ function template_preprocess_ding_list_add_button(&$variables) {
...
-        'query' => array('at' => $variables['type'], 'av' => $variables['value']),
-        'html' => TRUE,
-      ),
-    );
+      );
+    }
+    elseif (!user_is_logged_in()) {
+      $new_list = array(
+        'custom' => array(
+          'title' => '<div class="list-title">' . t('Login before you can use lists') . '</div><div class="add-to-list-button button">' . t('Login') . '</div>',
```
kasperg: Is there a reason why the existing approach with `ajax_command_ding_user_authenticate()` is not used here? It should provide a smoother user experience where a login automatically followed by the original intent of the user.

See [`ding_reservation_reserve_ajax()`](https://github.com/ding2/ding2/blob/50f3a749207ddd120772237f1fe3119afd0cee45/modules/ding_reservation/ding_reservation.module#L149-L173) for an example of how.

modules/p2/ding_list/theme/theme.inc
```diff
@@ -269,21 +267,33 @@ function template_preprocess_ding_list_add_button(&$variables) {
...
+          ),
+          'external' => TRUE,
+          'html' => TRUE,
 
-    if (!user_is_logged_in()) {
-      $new_list['custom']['attributes']['class'][] = 'use-ajax';
-      $new_list['custom']['href'] = 'ding-list/login-create';
+        ),
+      );
+      $force_multiple = TRUE;
```
kasperg: Why do we want to force multiple here in this case? Adding a comment explaining something non-obvious is nice.

modules/p2/ding_list/theme/theme.inc
BirkAndMe: At this point we don't know the user, so we don't know which list they want the object added to.
Not using the ajax call saves us a back and forth between the server, when we client side already knows what to do.

kasperg: I see the issue if this element is supported to show an entry for each list the item can be added to then we cannot predict at this point which of these the user intends to click next.

The preexisting solution would be to add a single "Add to list" button to elements and then (after login if needed) the user would choose the list in a modal or the like. I guess that ship has sailed.

If we are to do things differently I would prefer a more widely usable solution e.g. just adding a single link "Add to new list" with a specific class. If a link/button with such a class is clicked and the user is not logged in this will cause the login modal to open and if login succeeds this redirects to the original target.

Otherwise I fear that this pattern of handling the unauthenticated user will spread throughout the codebase.

modules/p2/ding_list/theme/theme.inc
```diff
@@ -269,21 +267,38 @@ function template_preprocess_ding_list_add_button(&$variables) {
...
+          'href' => '#',
+          'attributes' => array(
+            'class' => array('open-login'),
+          ),
+          'external' => TRUE,
+          'html' => TRUE,
 
-    if (!user_is_logged_in()) {
-      $new_list['custom']['attributes']['class'][] = 'use-ajax';
-      $new_list['custom']['href'] = 'ding-list/login-create';
```
kasperg: Remember to remove this entry and related code from `ding_list_menu()`. It does not seem to be used anymore.

tests/behat/features/03-searching-object.feature
```diff
@@ -74,13 +74,6 @@ Feature: test object displays after search
     When I display random object from file
     Then it is possible to add to a list
 
-
-  @api @seek026 @seekNologin @regression @cci
-  Scenario: S026 Search and display material while not logged in should not show all buttons
-    Given filename 'creator.dat' is used
-    When I display random object from file
-    Then it is not possible to add to a list
```
kasperg: Consider keeping this scenario and replacing with  `Then it should be possible to add to a list`

kasperg:
Good question regarding referencing functions. As I see it the only one in question is `it is not possible to add to a list` as the related code. I think we should remove it as well.

As I understand the intent of this change we do not have situations where we do not want to display this possibility.

I am open to other arguments as well though.

kasperg:
Also please ensure that the PR merges.

kasperg:
Thanks for the updates @BirkAndMe!

PR #1136, opened 2018-07-11
2158: Add core patch to disallow gatewayf in robots

holt83:
https://platform.dandigbib.org/issues/2158

Had a lot of issues with the Makefile, since it needs tabs to separate lines. My editor is set to use spaces when working with ding2 and Drupal in general.
And sublime text was not helping. Had to use this plugin: https://packagecontrol.io/packages/TabNukerNuker 

Anyway, I think I got those tabs inserted after some hard work hehe :)

I was wondering: Are there any particular reason why we use a MakeFile for circleci? I can't seem to find any documentation about that.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,8 +25,8 @@ circle-setup:
 	cp -R ./* $(DRUPAL_SITE_PATH)/profiles/ding2/
 	# Install the site using the ding2 profile
 	cd $(DRUPAL_SITE_PATH) && drush site-install ding2 --db-url=mysql://ubuntu@127.0.0.1/circle_test -y ding2_module_selection_form.providers_selection=connie
-  # Apply Drupal core patches
-  git apply -v profiles/ding2/patches/drupal_core.robots.txt.ding2.patch
+	# Apply Drupal core patches
+	cd $(DRUPAL_SITE_PATH) && git apply -v $(DRUPAL_SITE_PATH)/profiles/ding2/patches/drupal_core.robots.txt.ding2.patch
 	# Notices and warnings are seen as an error from simpletests point of view
 	# and is added to the xml-output as such. Idealy we would not have any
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -25,6 +25,8 @@ circle-setup:
 	cp -R ./* $(DRUPAL_SITE_PATH)/profiles/ding2/
 	# Install the site using the ding2 profile
 	cd $(DRUPAL_SITE_PATH) && drush site-install ding2 --db-url=mysql://ubuntu@127.0.0.1/circle_test -y ding2_module_selection_form.providers_selection=connie
+	# Apply Drupal core patches
+	cd $(DRUPAL_SITE_PATH) && git apply -v $(DRUPAL_SITE_PATH)/profiles/ding2/patches/drupal_core.robots.txt.ding2.patch
```
kasperg: Why not include it in `drupal.make` instead [like we do for other Core patches](https://github.com/ding2/ding2/blob/master/drupal.make#L10)?

README.md
```diff
@@ -107,6 +114,13 @@ manually in a production environment.
   ~$ drush make --working-copy --contrib-destination=profiles/ding2/ drupal.make htdocs
 ```
 
+If you want to apply our custom Drupal core robots.txt patch, you still need to
+do that manually, as it's not possible to add it to the standalone drupal.make
+file
+```sh
+  patch -p1 < profiles/ding2/patches/drupal_core.robots.txt.ding2.patch
+```
```
kasperg: I understand the problem referring to a local patch if the user has only retrieved the single make file. I would prefer that we updated our approach for the alternative installation method to clone the entire repository and then run `drush make` on the cloned `drupal.make` file.

holt83:
PR is updated

PR #1137, opened 2018-07-12
3699 sticky events

holt83:
https://platform.dandigbib.org/issues/3699


modules/ding_app_content_rss/ding_app_content_rss.rules_defaults.inc
```diff
@@ -1,7 +1,7 @@
 <?php
 /**
  * @file
- * Rules configuration.
+ * ding_app_content_rss.rules_defaults.inc
```
JrgenGNielsen: Why change this? Also: throws a Scrutinizer error

holt83: I must admit I just did a `drush fu ding_app_content_rss` after updating the views in the UI.

I changed it back now to fix scrutinizer, but it will probably appear again if someone else does an export, since this is appearently how Features wants it.

JrgenGNielsen:
reviewed and approved

PR #1140, opened 2018-07-31
3706: Dont check false collections

BirkAndMe:
https://platform.dandigbib.org/issues/3706

### DIFF
--- a/modules/p2/ding_message/ding_message.module
+++ b/modules/p2/ding_message/ding_message.module
@@ -249,41 +249,42 @@ function ding_message_check_element($uid, DingListElement $element, $message = N
       if ($result->getNumTotalCollections() > 0) {
         // Get the first element in the search result set.
         $collections = $result->getTingEntityCollections();
-        $first_collection = current($collections);
-        if (!empty($message)) {
-          // If the first search result element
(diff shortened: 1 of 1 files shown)


modules/p2/ding_message/ding_message.module
```diff
@@ -249,42 +249,43 @@ function ding_message_check_element($uid, DingListElement $element, $message = N
...
-            $count = array_search($last_element_id, $collection_ids);
-            // Add the count to the new elements count.
-            $new_elements_count += $count;
-            if ($status != DING_MESSAGE_ACTIVE) {
-              // If the messages has been seen,
-              // set the last and first element ids.
-              // Set all values for a new message to be created.
-              $last_element_id = $first_collection->getId();
-              $first_element_id = $collection_ids[$count - 1];
+        if ($first_collection = current($collections)) {
```
kasperg: As I see it this only ensures that the first entry in `$collections` is not `false`. However the remaining entries in might have an unexpected value as well. This would cause errors for `array_map` on L258. This has an anonymous function with a typed argument.

I suggest adding `$collections = array_filter($collections)` before the `if` statement to remove all entries evaluating to `FALSE`.

PR #1141, opened 2018-08-01
3703: Fix fatal php error when viewing /ekurser.

holt83:
And also ensure the search query is using facet.subject.

https://platform.dandigbib.org/issues/3703

Regarding simplification of facet.subject terms extraction. As mentioned in commit message:

As documented in \Ting\Search\TingSearchResultInterface, the getFacets() method returns a list of facets keyed by facet name. Hence there can only be one entry for the facet with name facet.subject and there's no need to do any filtering. 
This simplification also fixes the fatal PHP error, since it remove the need to use the splat-operator (which where failing before because of associative array)


kasperg:
@rasben Take a look here when working on #1163.

PR #1142, opened 2018-08-03
1491: Validate pincode on provider profile form

holt83:
https://platform.dandigbib.org/issues/1491


JrgenGNielsen:
reviewed and approved

PR #1143, opened 2018-08-06
3696: Check drupal_page_is_cacheable() for anonymous users

holt83:
https://platform.dandigbib.org/issues/3696

Some libraries was experiencing spam on the contact-form, which was already honeypot enabled.

I discovered that honeypot was trying to mark page as non-cacheable, but because of the logic in ding_varnish module this wasn't respected.

I have testet with Varnish that the contact-form isn't cached anymore with the changes in this PR.


JrgenGNielsen:
reviewed and approved

PR #1147, opened 2018-08-13
2223: New open search cache model

cableman:
This is a new PR (for #1075) without the availability optimisations.

See https://platform.dandigbib.org/issues/2223


modules/opensearch/opensearch.client.inc
```diff
@@ -732,11 +425,130 @@ function opensearch_execute($request) {
...
     timer_stop('ting');
   }
 }
 
+/**
+ * Try and predict future search requests and warm up cache.
+ *
+ * Base on the collections given and the entity model for ting objects we expect
+ * the next search requests to be getCollectionRequest and getObjectRequest
+ * without relations. The information to populate the cache with response to
```
kasperg: ...  but `opensearch_execute` will always enable relations given the change on L352?

modules/opensearch/opensearch.client.inc
```diff
@@ -732,11 +425,130 @@ function opensearch_execute($request) {
...
+ * Insert search response based on request.
+ *
+ * @param TingClientRequest $request
+ *   The search request that gives the response.
+ * @param \TingClientObject|\TingClientObjectCollection|array $response
+ *   The response to the request.
+ * @param array $opensearch_static_fast
+ *   The fast static cache used by opensearch_execute().
+ */
+function _opensearch_cache_insert(TingClientRequest $request, $response, array &$opensearch_static_fast) {
```
kasperg: If we have an `_opensearch_cache_insert` why is there no `_opensearch_cache_get`?

modules/opensearch/opensearch.client.inc
```diff
@@ -716,14 +394,29 @@ function opensearch_execute($request) {
...
       }
     }
 
+    // Update the cache with the raw response object from the data well for
+    // faster processing. Also set the static cache.
+    _opensearch_cache_insert($request, $response, $opensearch_static_fast);
+    $opensearch_static_fast[$cid] = $response;
+    if ($ttl = variable_get('opensearch_cache_lifetime', OPENSEARCH_DEFAULT_CACHE_LIFETIME)) {
+      cache_set($cid, $response, 'cache_opensearch', REQUEST_TIME + $ttl);
+    }
```
kasperg: L407-L410 seems to do the same as L406?

modules/ding_availability/ding_availability.field.inc
```diff
@@ -225,8 +225,11 @@ function ding_availability_field_formatter_view($entity_type, $entity, $field, $
             $url = entity_uri('ting_collection', $entities[0]);
           }
 
-          // Mark as available if it is an online resource.
-          if (in_array(drupal_strtolower($type), $online_types) || $entities[0]->getOnline_url(FALSE)) {
+          // Mark as available if it is an online resource. As some types are
+          // not always predefined online type we need to check if the entity
+          // has an online URL. Please not tht that this in most cases will
+          // trigger a search request with get all relations.
+          if (in_array(drupal_strtolower($type), $online_types) || $entities[0]->getOnline_url()) {
```
kasperg: Given the change to `opensearch_execute` additional search requests should be prevented.

modules/ding_carousel/js/ding_carousel.js
```diff
@@ -190,9 +190,14 @@
...
-            var id = match.substring(match.indexOf(':') + 1);
+            var local_id = match.substring(match.indexOf(':') + 1);
+            var id = match.match(/(\d+-\w+:\d+)$/)[0];
             match = match.replace('reservation', 'availability').replace(':', '');
-            Drupal.settings.ding_availability[match] = [ id ];
+            Drupal.settings.ding_availability.push({
+              'html_id': match,
+              'local_id': local_id,
+              'id': id
+            });
```
kasperg: This change seems to belong to #1149.

modules/ting/ting.module
```diff
@@ -468,20 +452,26 @@ function ting_ding_entity_buttons($type, $entity) {
...
+    if (strpos($view_mode, 'teaser') === FALSE) {
+      $collection = ting_collection_load($entity->id);
+      if (count($collection->getTypes()) > 1) {
+        $buttons[] = [
+          '#theme' => 'link',
+          '#text' => t('Other formats'),
+          '#path' => '#',
+          '#options' => [
+            'attributes' => [
+              'class' => ['action-button', 'other-formats'],
```
kasperg: To be honest such a button could as easily be generated clientside. It has no relevance serverside.

modules/opensearch/opensearch.client.inc
```diff
@@ -688,19 +330,55 @@ function opensearch_get_request_factory() {
...
 
+  // Due to the way the entity module is used to handle ting_objects a search
+  // with get all relations is going to be triggered for almost all requests. So
+  // to pre-heat the cache the optimal solution will be to get all relations
+  // within the first search request and predict the next request and put that
+  // result into the cache based on the first result. This will result in an
+  // slower first request but overall a lot fewer request to the data well in
+  // the search pages.
+  if ($request instanceof TingClientCollectionRequest || $request instanceof TingClientObjectRequest || $request instanceof TingClientSearchRequest) {
+    $request->setAllRelations(1);
```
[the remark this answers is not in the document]
cableman: Fixed

modules/ding_carousel/js/ding_carousel.js
cableman: Moved to #1149

modules/opensearch/opensearch.client.inc
cableman: Updated the comment

modules/opensearch/opensearch.client.inc
cableman: Yes, the insert is used from with in different helper functions (trying to pre-heat/predict the cache) where get is only preformed in one place.

modules/ding_availability/ding_availability.field.inc
cableman: They should but in some cases with list and cold cache this might still trigger an extra search request or get object.

modules/opensearch/opensearch.client.inc
cableman: Right... fixed

modules/ting/ting.module
cableman: It all comes back to the way the entity_button hook is miss-used in the theme.

E.g. https://github.com/ding2/ding2/blob/master/themes/ddbasic/template.php#L825

PR #1149, opened 2018-08-13
2223: availability optimization

cableman:
This is an optimization of availability from the #1075 pull request.

See https://platform.dandigbib.org/issues/2223


modules/fbs/includes/fbs.availability.inc
```diff
@@ -94,11 +140,7 @@ function fbs_availability_holdings($provider_ids) {
         $result_holding['placement'][] = $material_description[0]->materialGroupName;
       }
 
-      module_load_include('client.inc', 'opensearch');
-      $entity_id = ding_provider_build_entity_id($provider_ids[0]);
-      $suffix_type = variable_get('fbs_holdings_suffix_type', 'shelf_mark');
```
kasperg: Moving this *out* of the loop started on L75 seems to be a key change as `ding_provider_build_entity_id()` will invoke the search provider and [`opensearch` will issue a search](https://github.com/ding2/ding2/blob/078ff150143dd461b32ad842007716bf65328937/modules/opensearch/includes/opensearch.search.inc#L624).

This change seems unrelated to the rest of the changes here. Could you extract that into a separate PR?

modules/fbs/fbs.info.rej
```diff
@@ -0,0 +1,16 @@
...
+  dependencies[] = profile2
+  dependencies[] = text
+  dependencies[] = virtual_field
+--- 14,19 ----
+  dependencies[] = libraries
+  dependencies[] = list
+  dependencies[] = options
+  dependencies[] = profile2
+  dependencies[] = text
+  dependencies[] = virtual_field
```
kasperg: Bemærk at denne fil ikke bør committes.

modules/fbs/fbs.info.orig
```diff
@@ -0,0 +1,61 @@
...
+files[] = includes/classes/FBSCacheDrupal.php
+files[] = includes/classes/FBSCacheInterface.php
+files[] = includes/classes/FBSCacheMemory.php
+files[] = includes/classes/FBSDrupalHttpClient.php
+files[] = includes/classes/FBSFakeHttpClient.php
+files[] = includes/classes/FBSGuzzleHttpClient.php
+files[] = includes/classes/FBSLogDrupal.php
+files[] = includes/classes/FBSLogInterface.php
+files[] = includes/classes/FBSLogStdout.php
+files[] = includes/classes/FBSMisconfiguredException.php
```
kasperg: Bemærk at denne fil ikke bør committes.

modules/fbs/fbs.info.rej
cableman: Ups

modules/fbs/fbs.info.orig
cableman: Ups

modules/fbs/includes/fbs.availability.inc
cableman: No, it's not moved outside the loop???

It checks if the entity id exists in the current loop and loads ding_provider_build_entity_id(reset($item->recordId))

guddo:
I guess `modules/ding_availability/ding_availability.field.inc.orig` and `modules/fbs/includes/fbs.availability.inc.orig` should also not be committed..

DDBdrift:
https://platform.dandigbib.org/issues/2223 | Resolved (tag version)

PR #1152, opened 2018-08-16
3745: Fix misspelling of $account

ubcudvikler:
Fixes an error where global variable $user is
not assigned to $account if no arguments where
provided.


ubcudvikler:
Resolves: https://platform.dandigbib.org/issues/3745

PR #1153, opened 2018-08-16
1663: Implement date repeat entity on ding_event

holt83:
https://platform.dandigbib.org/issues/1663


holt83:
@JrgenGNielsen I have rebased.

holt83:
I got the "Errored" on scrutinizer. Seems like it never completed. Had the same issue yesterday.

holt83:
I get the following error in the behat-tests:

"When I use facets to reduce the search results to the highest possible: Failed: Didn't find any facets on the page. (Exception)"

Will have a look at it tomorrow.

holt83:
Ok, now scrutinizer and behat tests should be good!

PR #1157, opened 2018-08-20
3704 by Inlead: Prevent event nodes from failing.

guddo:
https://platform.dandigbib.org/issues/3704


modules/ding_place2book/ding_place2book.module
```diff
@@ -353,13 +353,26 @@ function ding_place2book_get_ticket_link($node, $place2book_id = NULL) {
...
+    $lock_id = implode('|', [
+      'entity:' => 'node',
+      'nid:' => $node->nid,
+      'type:' => $node->type,
+    ]);
+
+    // Handle concurrency.
+    // Field save might deliver integrity constraint exceptions when invoked
+    // by numerous simultaneous requests.
+    if (lock_acquire($lock_id, 1)) {
```
kasperg: The need for this change is caused by the change in #1104 where `ding_place2book_get_ticket_link()` is called from a hook `node_view()` implemenetation.

@guddo @bozeak: Can  provide a more detailed explanation of why we need to retrieve the ticket link from place2book on each node view? I would think it to be static once the event is created.

*If* we need to retrieve the link dynamically then we should refactor the field into using the `virtual_field` module like we do for similar fields instead of trying to store something in the database where it is not needed.

guddo: As written in the issue, this was done in PR#1096 and PR#1104, due to the lack of older nodes not getting the correct fields changed. It was then moved to node_view() to make sure that older nodes also got this change. However, in hindsight, this is not a feasible implementation as frontend users should not be able to alter anything frankly.

We do need to retrieve the status of the link dynamically, however it should be refactored somehow.

kasperg: > It was then moved to node_view() to make sure that older nodes also got this change. However, in hindsight, this is not a feasible implementation as frontend users should not be able to alter anything frankly.

I agree. If older nodes were not getting the correct fields changed they should have been updated in a `hook_update()`or the like.

Based on your explanation, are you also saying that Place2book is using an existing field `field_ding_event_ticket_link` which is a field that editors can use to manually specify a link for an event. If this is the case then this is obviously not something we can move to `virtual_field`.

If reacting to node changes on the site (`hook_node_insert/update()`) is not sufficient could we then update the link in the background using a cron job or the like?

Do you have other suggestions regarding how this should be handled in a better way?

modules/ding_place2book/ding_place2book.install
```diff
@@ -107,3 +107,27 @@ function ding_place2book_update_7001() {
     ->fields(array('ticket_type' => 'Adgang'))
     ->execute();
 }
+
+/**
+ * Update ding_event nodes with place2book data.
+ */
+function ding_place2book_update_7002() {
+  $queue = DrupalQueue::get('ding_place2book_update_events_queue');
+  $queue->createQueue();
```
kasperg: The verdict still seems to be out [whether it is necessary to call `createQueue` or not](https://api.drupal.org/api/drupal/modules%21system%21system.queue.inc/function/DrupalQueueInterface%3A%3AcreateQueue/7.x).

If you think it is necessary then it should also be added in `hook_install`.

kasperg: The verdict still seems to be out [whether it is necessary to call `createQueue` or not](https://api.drupal.org/api/drupal/modules%21system%21system.queue.inc/function/DrupalQueueInterface%3A%3AcreateQueue/7.x).

If you think it is necessary then it should also be added in `hook_install`.

modules/ding_place2book/ding_place2book.install
```diff
@@ -107,3 +107,27 @@ function ding_place2book_update_7001() {
...
+
+  $result = db_query("select n.nid from {node} n where
+    n.type = 'ding_event' and
+    n.nid not in
+    (select d.nid from {ding_place2book} d)")->fetchCol();
+
+  foreach ($result as $nid) {
+    $queue->createItem($nid);
```
kasperg: Why create a queue for something that is a one-time change?

The queue stays around even after the update hook has run but is not needed afterwards.

I would expect a implementation contained within the update hook/`.install` file unless you see any other architectural perspective in this.

AnatolyMuntean:
The purpose of all this is make one-time update of the events that were missed during previous rolled out change-sets. Yet it's still a one time-change, hence the hook_update. The queue is filled once, in the update hook, and processed during consequent cron runs. Even if it has to stay it would stay empty.

kasperg:
@AnatolyMuntean So why use a queue to do this at all?

AnatolyMuntean:
Since I'm not aware of how many nodes would require update on a certain instance.

kasperg:
But hook_update already supports multiple passes?

> The $sandbox parameter should be used when a multipass update is needed, in circumstances where running the whole update at once could cause PHP to timeout. Each pass is run in a way that avoids PHP timeouts, provided each pass remains under the timeout limit. To signify that an update requires at least one more pass, set $sandbox['#finished'] to a number less than 1 (you need to do this each pass). The value of $sandbox['#finished'] will be unset between passes but all other data in $sandbox will be preserved. The system will stop iterating this update when $sandbox['#finished'] is left unset or set to a number higher than 1. It is recommended that $sandbox['#finished'] is initially set to 0, and then updated each pass to a number between 0 and 1 that represents the overall % completed for this update, finishing with 1.
https://api.drupal.org/api/drupal/modules%21system%21system.api.php/function/hook_update_N/7.x

AnatolyMuntean:
Yes, sure it could be used like that.

GitCop:

 * Commit: bf2c58083aeed66183dbbafc8db4119c301dfc4e

---

GitCop:

 * Commit: bf2c58083aeed66183dbbafc8db4119c301dfc4e

---

PR #1160, opened 2018-08-24
3078: Correct title when renewing multiple loans.

tripox:


kasperg:
This reintroduces changes which were previosly reverted in #994.

PR #1165, opened 2018-08-29
3305 by Inlead: Add node search to ting_field_search.

guddo:
https://platform.dandigbib.org/issues/3305

### DIFF
--- a/modules/ting_field_search/ting_field_search.module
+++ b/modules/ting_field_search/ting_field_search.module
@@ -598,11 +598,6 @@ function ting_field_search_form_search_block_form_alter(&$form, &$form_state) {
     return;
   }
 
-  // Don't alter the form if this is the node search page.
-  if (strpos(current_path(), 'search/node') === 0) {
-    return;
-  }
-
   $exposed_profiles = ting_field_search_filter_profiles();
 
   // Don't render select if there's no exposed profiles.
@@ -611,6 +606,11 @@ function ting_field_search_form_search_block_form_alter(&$form, &$form_state) {
   }
 
   $profile = ting_field_search_get_default_profile();
+
+  if (strpos(current_path(), 'search/node') !== FALSE) {
+    $profile = 'node';
+  }
+
   $form['ting_field_search'] = array(
     '#type' => 'select',
     '#empty_option' => t('Search in all'),
@@ -624,6 +624,10 @@ function ting_field_search_form_search_block_form_alter(&$form, &$form_state) {
     '#type' => 'hidden',
     '#default_value' => $profile ? $profile : '',
   );
+
+  // Adding content search option into ting_field_search dropdown.
+  $node_search = array('node' => t('Search Content'));
+  $form['ting_field_search']['#options'] = $node_search + $form['ting_field_search']['#options'];
 
   // Perform som alterations before ting_search_submit in a validate handler.
   $form['#validate'][] = 'ting_field_search_search_block_validate';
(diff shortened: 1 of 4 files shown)


modules/ting_search/ting_search.module
```diff
@@ -226,9 +226,27 @@ function ting_search_form_search_block_form_alter(&$form, &$form_state, $form_id
...
+  }
+
   return $form;
 }
 
+/**
+ * Form submit callback for ting_field_search site content search option.
+ */
+function ting_search_content_search_submit(&$form, &$form_state) {
+  if ($form_state['input']['ting_field_search'] == 'node') {
```
holt83: Any reason why not using form_state['values']? It should be the preferred way: https://drupal.stackexchange.com/questions/112312/difference-between-form-statevalues-and-form-stateinput 

Also, form_state['values] are used elsewhere.

modules/ting_search/ting_search.module
```diff
@@ -226,9 +226,27 @@ function ting_search_form_search_block_form_alter(&$form, &$form_state, $form_id
...
+    unset($form_state['redirect']);
+    $form_state['action'] = "search/node";
+    $form_state['redirect'][0] = 'search/node/' . $form_state['values']['search_block_form'];
```
holt83: Why not just override it like this:

`$form_state['redirect'] = 'search/node/' . $form_state['values']['search_block_form'];`

You could also delete the `unset($form_state['redirect']);` then.

modules/ting_search/ting_search.module
```diff
@@ -226,9 +226,27 @@ function ting_search_form_search_block_form_alter(&$form, &$form_state, $form_id
...
+    $form['#submit'][] = 'ting_search_content_search_submit';
+  }
+
   return $form;
 }
 
+/**
+ * Form submit callback for ting_field_search site content search option.
+ */
+function ting_search_content_search_submit(&$form, &$form_state) {
```
holt83: This could also be handled in ting_field_search in `ting_field_search_search_block_submit()`, if you move the adding of the option to `ting_field_search_form_search_block_form_alter()` also. (See comment above).

modules/ting_search/ting_search.module
```diff
@@ -226,9 +226,27 @@ function ting_search_form_search_block_form_alter(&$form, &$form_state, $form_id
   // information.
   $form['#action'] = url('search-blank');
 
+  // Adding content search option into ting_field_search dropdown.
+  if (!empty($form['ting_field_search'])) {
+    $node_search = array('node' => t('Search Content'));
+    $form['ting_field_search']['#options'] = $node_search + $form['ting_field_search']['#options'];
+    $form['#submit'][] = 'ting_search_content_search_submit';
+  }
```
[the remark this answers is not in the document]
bozeak: Moved to ting_field_search.

modules/ting_search/ting_search.module
bozeak: +

modules/ting_search/ting_search.module
bozeak: +

modules/ting_field_search/ting_field_search.module
```diff
@@ -611,6 +606,11 @@ function ting_field_search_form_search_block_form_alter(&$form, &$form_state) {
   }
 
   $profile = ting_field_search_get_default_profile();
+
+  if (strpos(current_path(), 'search/node') !== FALSE) {
```
holt83: I think more correct would be `strpos(current_path(), 'search/node') === 0` so you're actually verifying that current_path starts with 'search/node'?

bozeak: \+

JrgenGNielsen:
reviewed and approved

PR #1096, opened 2018-05-14
3538 by Inlead: Populating ticket link and event status from place2book.

inlead:
https://platform.dandigbib.org/issues/3538

PR #1166, opened 2018-08-30
3755 by Inlead: Re-implement changes from #1096 (issue 3538).

guddo:
https://platform.dandigbib.org/issues/3755

### DIFF
--- a/modules/ding_place2book/ding_place2book.module
+++ b/modules/ding_place2book/ding_place2book.module
@@ -1011,8 +1011,6 @@ function place2book_get_api_key($library_id = NULL) {
  * Rendering function for place2book tickets info
  */
 function ding_place2book_render_place2book_ticketsinfo($node) {
-  $type = '';
-  $url = '';
   $place2book_id = 0;
   $maintain_copy = 0;
   $passive = 0;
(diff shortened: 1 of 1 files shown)


modules/ding_place2book/ding_place2book.module
```diff
@@ -1067,52 +1065,68 @@ function ding_place2book_render_place2book_ticketsinfo($node) {
       $p2b_result = drupal_http_request($service_settings['service_url'] . '/available_tickets', $options);
 
       //we must have a sales-status header
-      if ($p2b_result->headers['sales-status']) {
+      $status_type = _ding_place2book_status_processing($p2b_result);
+    }
+  }
+  return theme('place2book_ticketsinfo', array(
```
cableman: Would it be possible to return an render array here? Would allow easy changes in the theme layer, if any libraries should want to?

guddo: Yes we can move this into a renderable array, but in this case we will have to review the modality of displaying this button on the frontend and thorough testing. Should we proceed with this?

cableman: Okay, not a the pressen time as all tests for this release is currently halted and we need to get this fix into the code base.

cableman:
Reviewed and awaiting release

PR #1170, opened 2018-09-03
Default information text when creating new library users

cableman:
See https://platform.dandigbib.org/issues/3590

* New default text to gatewayf create new user information page (better default text from DDB).
* Reset to default text button.
* Styled the new page + highlight "words" that should be changed.


modules/ding_gatewayf/modules/ding_gatewayf_registration/templates/ding-gatewayf-registration-information.tpl.php
```diff
@@ -18,13 +18,11 @@
     <div class="pane-content">
       <div class="primary-content ding-gatewayf-registration-information">
         <h2 class="pane-title"><?php print render($title); ?></h2>
-        <div class="ding-gatewayf-registration-information--left">
+        <div class="long-text-wrapper">
```
kasperg: In the spirit of better class names should this be `ding-gatewayf--registration-information` and the parent class changed to `ding-gatewayf` (instead of `ding-gatewayf-registration-information`)  just as the class for the link is `ding-gatewayf--link-wrapper`?

modules/ding_gatewayf/modules/ding_gatewayf_registration/ding_gatewayf_registration.module
```diff
@@ -328,6 +334,19 @@ function ding_gatewayf_registration_acceptance_form() {
...
+ * Sumbut handler that resets the information text.
+ */
+function ding_gatewayf_registration_information_reset($form, $form_state) {
+  $config = variable_get('ding_gatewayf_registration', FALSE);
+  if ($config) {
+    unset($config['information']);
+  }
+  variable_set('ding_gatewayf_registration', $config);
+
+  drupal_set_message(t('Information text have been reset to DDB default text'));
```
kasperg: This will also reset the text on the live site visible by end users until the editor has submitted the new text. Could we load the default text into the editor as a default value when clicking the button?

modules/ding_gatewayf/modules/ding_gatewayf_registration/default_information_text.html
```diff
@@ -0,0 +1,58 @@
+<h2><span class="change-me">Danevangs Kommunes Bibliotekers</span> behandling af personoplysninger om dig</h2>
```
kasperg: `<span class="change-me">` - could we replace this with something like .ding-gateway--temporary-text`?

Could we replace this with something better/more relevant like the site name from the drupal configuration?

cableman: Class names will be changed.

modules/ding_gatewayf/modules/ding_gatewayf_registration/templates/ding-gatewayf-registration-information.tpl.php
cableman: Class names will be changed.

modules/ding_gatewayf/modules/ding_gatewayf_registration/ding_gatewayf_registration.module
cableman: I think the idea i very good. But i really don't known how to do this when its wysiwyg element with out writing a bounce of JS code?

PR #1175, opened 2018-09-06
3672: "Add search to list" shortcut in header search.

rasben:
The button for adding a search to a list
must be displayed instead in the header search
as a star icon.
This commits:
- Expands icomoon icon font with an outline star
- Adds new detector classes to the add-list element
- Using JS, we hide the original button and display
  a shortcut star in the header search.
- Conditional styling, depending if the shortcut is displayed


rasben:
To me, the things that scrutinizer complains about makes no sense:
- "else must start on a new line"  / "Closing brace must be on a line by itself" - They're totally on its own line already
- "The final ?> should be omitted from all code files" - Thats not following the Drupal guidelines as far as i know - and it doesnt complain about that in other .tpl files?

rasben:
This is now ready to be reviewed by a non-Reload core developer.

cableman:
Please rebase the PR to match the new master branch :-)

Also please use the "template" that is displayed when making PR's (https://github.com/ding2/ding2/blob/master/.github/PULL_REQUEST_TEMPLATE.md) as this helps in the review process and ensures that other are informed about the changes and get an easy overview of a given PR.

modules/ding_webtrends/ding_webtrends.module
```diff
@@ -151,17 +151,34 @@ function ding_webtrends_get_search_info() {
...
+
+/**
+ * Helper function for preprocessing ding list add button.
  *
  * The ding list add buttons is AJAX-enabled and we want to track those with
  * dcsMultiTrack. Unfortunately, the standard classes is not enough to make the
  * selector specific enough, so we add our own here.
  */
-function ding_webtrends_preprocess_ding_list_add_button(&$variables) {
+function _ding_webtrends_preprocess_ding_list_add_button($variables) {
```
cableman: Why not parse the $variables as ref. the whole way around? It kind of jump into my eyes that the $variables where total overriden in the functions calling this function.

Drupal normally change the $vars variable not override it. I known the end result is the same.

cableman: Maybe also remove the word "preprocess" from the helper function as that makes it look like a hook function. It might lead to confusion later on.

rasben: In this case, we want to mutate/extend the $variables, which is why i think the $var = $newvar is a better syntax for readability

Just using some_function($var) signals in my brain that literally anything could be going on - maybe we are mutating a third part based on $var..

Does that make sense? :)

modules/ding_webtrends/ding_webtrends.module
```diff
@@ -151,17 +151,34 @@ function ding_webtrends_get_search_info() {
 
 /**
  * Implements hook_preprocess_ding_list_add_button().
+ */
+function ding_webtrends_preprocess_ding_list_add_button(&$variables) {
+  $variables = _ding_webtrends_preprocess_ding_list_add_button($variables);
+
```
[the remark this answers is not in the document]
rasben: Fixed in 440b09f 👍

modules/ding_webtrends/ding_webtrends.module
rasben: Also, "preprocess" has been replaced by "prepare" in Fixed in 440b09f 👍

rasben:
I had no idea we had a PR template, so yeah I'll fill that out for you ASAP :)
I just have to get my local environment up and running again so I can take screenshots.
I also need to double check that everything still works after the merge rebase - I'll make sure to tag you when its ready again

PR #1177, opened 2018-09-10
3818: Fix missing bundle

BirkAndMe:
https://platform.dandigbib.org/issues/3818


cableman:
Approved and awaiting release

PR #1180, opened 2018-09-10
2690: Added Teaser no overlay

b14sha:
https://platform.dandigbib.org/issues/2690


DDBdrift:
https://platform.dandigbib.org/issues/2690 | Unable to reproduce (Closed)

PR #1182, opened 2018-09-11
3582: Expire inactive provider sessions on cron run

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/3583

#### Description

Ensure that rows in the session table for providers users is removed if they are older than the value set in `ding_user_provider_session_expire` variable.

These rows contains sensitive information about library users so it's important that the session is cleaned up periodically. Especially now where the autologout module has been removed the probability for stagnant rows session has increased.

##### IMPORTANT
The module will clear session on cron runs as default. If a more frequent cleaning is needed, a lightweight cron endpoint is provided which only clears sessions. It might not be appropriate to use the normal cron for this if it's too frequent since it performs a lot of other operations.  

An alternative approach would be to implement a drush-command and have a cron job call that instead.

#### Screenshot of the result

#### Checklist

`ding_user_provider_session_expire`  is set to 15 min as default, since this was the old autologout value for provider users. Maybe a different value is more appropiate now?

### DIFF
--- a/modules/ding_user/ding_user.admin.inc
+++ b/modules/ding_user/ding_user.admin.inc
@@ -1,0 +1,43 @@
+<?php
+/**
+ * @file
+ * Ding user settings.
+ */
+
+/**
+ * Ding user admin.
+ */
+function ding_user_admin_setting_form($form, &$form_state) {
+  $limits = drupal_map_assoc(array(
+    0,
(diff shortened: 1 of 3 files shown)


JrgenGNielsen:
reviewed and approved

PR #1187, opened 2018-09-13
3485: Align signup btn left

b14sha:
https://platform.dandigbib.org/issues/3485


DDBdrift:
https://platform.dandigbib.org/issues/3485 | Closed

PR #1189, opened 2018-09-13
3350: Changed labels search-result

b14sha:
https://platform.dandigbib.org/issues/3350

### DIFF
--- a/themes/ddbasic/template.node.php
+++ b/themes/ddbasic/template.node.php
@@ -50,6 +50,8 @@ function ddbasic_process_node(&$variables, $hook) {
   // For search result view mode add type and title.
   if ($variables['view_mode'] == 'search_result') {
     $node_type = node_type_get_name($variables['type']);
+
+    // Create markup.
     $variables['content']['type'] = array(
       '#markup' => '<div class="view-mode-search-result-content-type">' . $node_type . '</div>',
       '#weight' => -9999,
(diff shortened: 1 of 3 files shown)


themes/ddbasic/template.node.php
```diff
@@ -50,6 +50,8 @@ function ddbasic_process_node(&$variables, $hook) {
   // For search result view mode add type and title.
   if ($variables['view_mode'] == 'search_result') {
     $node_type = node_type_get_name($variables['type']);
+
+    // Create markup.
```
kasperg: However harmless this change on its' own does not seem relevant to this PR.

translations/da.po
```diff
@@ -17064,13 +17064,13 @@ msgid "Date/time on which the article will be automatically unpublished"
...
 msgid "Event"
-msgstr "Begivenhed"
+msgstr "Arrangement"
 
 msgid "Category"
 msgstr "Kategori"
 
 msgid "Groups"
-msgstr "Grupper"
+msgstr "Tema"
```
kasperg: Please remember to add [an update hook to load the updated translations](https://github.com/ding2/ding2/blob/master/ding2.install#L598).

kasperg: ~~Also: Are you sure that "Groups" (plural form) should be translated to "Tema" (singular form)?~~

I see that this is fine [as `ding_groups` actually calls the content type Groups](https://github.com/ding2/ding2/blob/e9b89ed2b14caa58d168a83a805fc3e03da5db6d/modules/ding_groups/ding_groups.features.inc#L73).

JrgenGNielsen:
reviewed and approved

PR #1193, opened 2018-09-19
3827: Error handling in predict cache

cableman:
#### Link to issue

https://platform.dandigbib.org/issues/3827

#### Description

Predict cache function do not handle empty collections do to data-well errors 

#### Checklist

### DIFF
--- a/modules/opensearch/opensearch.client.inc
+++ b/modules/opensearch/opensearch.client.inc
@@ -410,7 +410,7 @@ function opensearch_execute(TingClientRequest $request) {
     // Alternative rewrite the whole ting entity module, which currently is
     // overkill.
     if ($response instanceof TingClientSearchResult) {
-      _opensearch_predict_cache($response->collections, $opensearch_static_fast);
+      _opensearch_predict_cache($response->collections, $opensearch_static_fast, $request);
     }
 
     return $response;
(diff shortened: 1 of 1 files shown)


modules/opensearch/opensearch.client.inc
```diff
@@ -441,66 +441,73 @@ function opensearch_execute(TingClientRequest $request) {
...
  * @param array $opensearch_static_fast
  *   The fast static cache used by opensearch_execute().
+ * @param TingClientRequest $request
+ *   The request used to get the collections.
  *
  * @throws \TingClientException
  *   If the search client request fails.
  */
-function _opensearch_predict_cache(array $collections, array &$opensearch_static_fast) {
+function _opensearch_predict_cache(array $collections, array &$opensearch_static_fast, TingClientRequest $request) {
```
kasperg: Instead of this change I would much prefer that `_opensearch_predict_cache()` took a [`TingSearchResultInterface`](https://github.com/ding2/ding2/blob/c668d5e52c165b62efc55a4efde7796a5e3c4943/modules/ting/src/Search/TingSearchResultInterface.php#L14) object as the first argument. You can retrieve both resulting collections and the original request from the response.

cableman: From opensearch_execute() which uses this function I dont have the OpenSearchTingSearchResult which is the only object that implements that interface?

OpenSearchTingSearchResult is first created in opensearch_search_search(), which calls opensearch_do_search() which then calls opensearch_execute().

kasperg: You are absolutely right. Sorry about the confusion.

PR #1195, opened 2018-09-19
3804: Rename getNumTotalCollections to getNumCollections

xendk:
#### Link to issue

https://platform.dandigbib.org/issues/3804

#### Description

Rename method as described in the issue.

#### Checklist


xendk:
Circle is red, but it isn't related to this change.

xendk:
Rebuilding fixed circle.

PR #1201, opened 2018-09-24
2376: Fixed getElements() keyed array

BirkAndMe:

### DIFF
--- a/modules/p2/ding_list/src/DingList.php
+++ b/modules/p2/ding_list/src/DingList.php
@@ -602,7 +602,7 @@ class DingList {
         $this->attachElement($element, TRUE);
       }
     }
-    $result = array_slice($this->elements, $offset, $count);
+    $result = array_slice($this->elements, $offset, $count, TRUE);
 
     return $result;
   }


modules/p2/ding_list/src/DingList.php
```diff
@@ -602,7 +602,7 @@ public function getElements($offset = 0, $count = NULL) {
         $this->attachElement($element, TRUE);
       }
     }
-    $result = array_slice($this->elements, $offset, $count);
+    $result = array_slice($this->elements, $offset, $count, TRUE);
```
kasperg: This change indicates that there are callers which expect something regarding the keys of the returned array of list elements. I can tell when reading the code from `attachElement` that ids are used as keys.

Please implement one of two changes:

1. Update the documentation to reflect the contract of the function e.g.

```php
    * @return DingListElement[]
    *   The elements. Element ids are used as keys.
```

2. That callers are updated to extract keys themselves from the elements that they retrieve.

```php
$list_element_ids = array_map(function(\DingListElement $element) {
  return $element->getId();
}, $list->getElements());
```

1 is probably easier than 2.

Please remember provide information about *why* a change is needed in commits and in PRs. That makes them much easier for others to grasp - both then reviewing and in the future.

PR #1203, opened 2018-09-24
3473 update ckeditor

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/3473

#### Description

Security update of ckeditor library: https://www.drupal.org/sa-core-2018-003 

I had to update the wysiwyg module also. Refer to this editor support matrix: https://www.drupal.org/docs/7/modules/wysiwyg/supported-editors-matrix 

I also did re-export of our Wysiwyg profile config after running the updates from the updated wysiwyg-module.

The version we used before was from may 2015 (ref: 7981731f4f3db2f932419499d2ec13a073e9b88f), There's a lot of changes from 2.2 to 2.5. Looks like editor preferences is separated from profile settings and editor version is now exported in features.

The available options appear to be largely the same, apart from a view additions to the paste plugin (I have informed about this in the case) and that now you can set editor language (which ofc just will be the site's default danish)

#### Checklist

### DIFF
--- a/ding2.make
+++ b/ding2.make
@@ -454,12 +454,8 @@ projects[workflow][patch][] = "http://www.drupal.org/files/issues/features_impor
 ; Prevent fatal errors on cron when using Scheduler, https://www.drupal.org/node/2499193.
 projects[workflow][patch][] = "https://www.drupal.org/files/issues/workflow-php_fatal_error_call-2499193-7-2.5.patch"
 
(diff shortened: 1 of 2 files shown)


holt83:
I don't know what's going on with build at Circle. When it tries to build with `drush make` it gets `drush: Command not found` resulting in errors in the rest of build process.

holt83:
I just started a rebuild. Seems like it's going better this time.

JrgenGNielsen:
reviewed and approved

PR #1210, opened 2018-10-04
3043: Fixed on hold date validation

b14sha:
https://platform.dandigbib.org/issues/3043

### DIFF
--- a/modules/fbs/fbs.module
+++ b/modules/fbs/fbs.module
@@ -586,6 +586,16 @@ function fbs_form_profile2_form_alter(&$form, &$form_state) {
       '#weight' => $form['profile_provider_fbs']['field_fbs_phone_notification']['#weight'] + 0.1,
     );
 
+    // If on_hold values are outdated (eg. to date less than today), reset on_hold default values.
+    $to_date = $form['profile_provider_fbs']['field_fbs_on_hold']['und'][0]['to']['#default_value'];
+    $to_date_stamp = strtotime($to_date);
+    $now = date('d-m-Y');
+    $now_stamp = strtotime($now);
+    if ($to_date_stamp < $now_stamp) {
+      $form['profile_provider_fbs']['field_fbs_on_hold']['und'][0]['from']['#default_value'] = '';
+      $form['profile_provider_fbs']['field_fbs_on_hold']['und'][0]['to']['#default_value'] = '';
+    }
+
     // Add validation to email.
     $form['profile_provider_fbs']['field_fbs_mail']['#element_validate'] = array('fbs_form_profile2_mail_validate');
     // Add validation to phone.
@@ -646,10 +656,6 @@ function fbs_form_profile2_onhold_validate($element, $form_state) {
   $to_date_stamp = strtotime($to_date);
   $now = date('d-m-Y');
   $now_stamp = strtotime($now);
-
-  if ($from_date_stamp < $now_stamp) {
-    form_set_error('profile_provider_fbs][field_fbs_on_hold', t('Start date must be higher than or equal to today') . ' - ' . $now . '.');
-  }
 
   if ($to_date_stamp < $now_stamp && $to_date_stamp < $from_date_stamp) {
     form_set_error('profile_provider_fbs][field_fbs_on_hold', t('To date must both be higher than or equal to today, and be higher than the start date.'));
--- a/modules/fbs/fbs.module
+++ b/modules/fbs/fbs.module
@@ -603,7 +603,7 @@ function fbs_form_profile2_form_alter(&$form, &$form_state) {
     // Add validation to on_hold
     $form['profile_provider_fbs']['field_on_hold']['#element_validate'] = array('fbs_form_profile2_onhold_validate');
     // Add submit handler, to redirect to same URL.
-    $form['#submit'][] = 'fbs_form_profile2_rebuild_submit';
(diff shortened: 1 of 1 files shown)


modules/fbs/fbs.module
```diff
@@ -647,10 +657,6 @@ function fbs_form_profile2_onhold_validate($element, $form_state) {
   $now = date('d-m-Y');
   $now_stamp = strtotime($now);
 
-  if ($from_date_stamp < $now_stamp) {
-    form_set_error('profile_provider_fbs][field_fbs_on_hold', t('Start date must be higher than or equal to today') . ' - ' . $now . '.');
-  }
-
```
kasperg: I understand the situation regarding this validation failing when the user makes unrelated changes to her profile. However I am not a fan of removing this validation. It will allow the user to set a value in the past which we do not want to.

Could we instead only perform this check if  it has been changed by the user - e.g. by checking that the field has a different value than the default value?

By using this approach we also avoid removing the users ability to see previous state.

kasperg:
Please remember to include *why* a change is needed. Preferably in the commit comment. "Fixed on hold date validation" does not really say much and it is really helpful for others who would otherwise have to dig through an issue with 38 comments in danish.

modules/fbs/fbs.module
```diff
@@ -649,31 +649,47 @@ function fbs_form_profile2_phone_validate($element, $form_state) {
...
 
 /**
  * Submit handler for fbs_form_profile2
  */
-function fbs_form_profile2_rebuild_submit($form, &$form_state) {
-  // Rebuilds form and stay on same page.
-  $form_state['rebuild'] = TRUE;
+function fbs_form_profile2_redirect_submit($form, &$form_state) {
+  // Redirects form to stay on same page on submit.
+  $form_state['redirect'] = $_GET['q'];
```
kasperg: Do we need this submit handler at all if we just want to redirect to the same page? That seems to be the default behavior of [`drupal_redirect_form()`](https://api.drupal.org/api/drupal/includes%21form.inc/function/drupal_redirect_form/7.x)?

modules/fbs/fbs.module
```diff
@@ -639,35 +649,47 @@ function fbs_form_profile2_phone_validate($element, $form_state) {
...
+    $from_date_stamp != $default_from_date_stamp && 
+    $from_date_stamp <= $now_stamp) {
+    form_set_error('profile_provider_fbs][field_fbs_on_hold', 
+      t('Your reservations is already on hold from start date: %default-date. If you wish to change the start date from: %default-date, the start date must be higher than or equal to today: %now', 
+      array(
+        '%default-date' => format_date($default_from_date_stamp, 'custom', 'd-m-Y'), 
+        '%now' => $now),
+      )
+    );
+  }
```
kasperg: I am not sure I understand the use case for this validation error.

It triggers if:
- The user already has reservations on hold
- The user sets from date to another date in the past

I would think that the current error message on 674 is sufficient.

Perhaps you think it is necessary because the changes on L589 reset the default values for the hold period so the user might not be aware of any current onhold values?

In that case I suggest (again) that we do not reset past on hold values and only validate the from date if it has changed from the current default.

That means:
- Removing lines 589-597
- Removing lines 662-672
- Changing the validation on line 673 to:

```php
if ((empty($default_from_date) || $default_from_date_stamp != $from_date_stamp) &&
  $from_date_stamp <= $now_stamp) {
```

modules/fbs/fbs.module
```diff
@@ -639,35 +649,47 @@ function fbs_form_profile2_phone_validate($element, $form_state) {
...
+  if (empty($default_from_date) && $from_date_stamp <= $now_stamp) {
+    form_set_error('profile_provider_fbs][field_fbs_on_hold', t('Start date must be higher than or equal to today: %now', array('%now' => $now)));
```
kasperg: :+1: for improving strings with variable usage.

modules/fbs/fbs.module
b14sha: This form do not redirect to the same page, but redirects to /user/[uid]

modules/fbs/fbs.module
b14sha: I think it makes good sense to clear the on_hold values, if the end_date is surpassed, since then the on_hold values are no longer effective. I have implemented the solution in respect to the client wishes - ultimately it must be the client decision if we want to clear the values or not.

If a user tries to alter the on_hold values (when they are set), the user is prompted to choose a start date larger than today. If the user input was made due to an (user) error, the user will have no way of resetting the selection if we do not provide the default_date entered. Then the user will either have to guess the old default date, or update the start date. Even though an update dont actually make any difference, it might confuse the user if we do not supply the opportunity to set the form back to its orginal state. Therefore we provide the default date in the validation string, to help the user set the field back to the original date, and hence validate the form.

kasperg: > If a user tries to alter the on_hold values (when they are set), the user is prompted to choose a start date larger than today. If the user input was made due to an (user) error, the user will have no way of resetting the selection if we do not provide the default_date entered. Then the user will either have to guess the old default date, or update the start date. Even though an update dont actually make any difference, it might confuse the user if we do not supply the opportunity to set the form back to its orginal state. Therefore we provide the default date in the validation string, to help the user set the field back to the original date, and hence validate the form.

Thanks for elaborating. I see your point. In that regard the improved feedback message makes sense.

PR #1214, opened 2018-10-09
Revert "813: Validate phone number"

kasperg:
Reverts ding2/ding2#580

Phone number validation is now handled by change in #1128.


kasperg:
Failing test does not seem to relevant for the change at hand. I will merge this.

PR #1215, opened 2018-10-10
3859: Fatal error when editing libraries

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/3859

#### Description

We have fixed this error before, but somehow the changes have made it out of the codebase. This PR re-introduces the changes from: #941.

Link to the original case: https://platform.dandigbib.org/issues/2610 

To sum up: The fatal error is caused by the changes in https://github.com/ding2/ding2/pull/244. When the URL-alias has the same language as the node, the error is triggered.
The solution is to not remove the alias form element and use this approach instead. 

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1228, opened 2018-10-18
2609: User consent text changed from node to panel pane

KristianB14:
#### Link to issue

https://platform.dandigbib.org/issues/2609

#### Description

An update function has been created. This deletes the existing node containing consent text and the variable containing the link to the node. A new panel pane and panel page with the text has been created instead. The code creating the user consent node has been removed, and the link to the consent page replaced with the link to the new panel page.

#### Screenshot of the result

No changes in the user interface.

#### Checklist

No further comments


JrgenGNielsen:
reviewed and approved

PR #1009, opened 2018-02-15
3296: Added on waiting list state

cableman:
Added new state for place2book tickets - on waiting list.

See https://platform.dandigbib.org/issues/3296

P.S. Also did a little coding style clean-up, but just the worst of it.

PR #1237, opened 2018-10-25
3296 by Inlead: Refactored waiting list state.

inlead:
#### Link to issue

https://platform.dandigbib.org/issues/3296

#### Description

Initial implementation (https://github.com/ding2/ding2/pull/1009) was made by checking whether sales status was open and amount of tickets is 0:
```
if ($p2b_result->headers['sales-status'] == 'open' && (int) $p2b_result->headers['available-tickets'] == 0) {
  $type = 'order-waiting-list';
  $url = $p2b_result->headers['location'];
}
```
However this is not what defines whether an event is with a waiting list or not. This is defined in the properties of the event in the external system, and the event gets a header called "`waiting-list`".

This value can be true even if the event is paused, stopped or open for sales. Therefore the check in this change-set is duplicated. Code is convoluted and refactoring is not a part of this task.

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1240, opened 2018-10-30
284 by Inlead: Added event target as filtering option.

inlead:
#### Link to issue

https://platform.dandigbib.org/issues/284

#### Description

Event Target has been added to the Filtering options of the Event liste page (/arrangementer).

#### Screenshot of the result

screenshot 2018-10-30 12 35 28

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1244, opened 2018-11-01
3908: Allow searching without keywords when using profile

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/3908 

#### Description

There was an option to allow for "empty searches" on profiles, but it stopped working because the '()' was, rightfully, removed from the query when there was no keywords entered in the UI.

I fixed this and decided to remove the option, such that it's always possible to search without keywords when using a profile. I think it makes really good sense to allow searching without keywords, when the user already have selected a profile. And I couldn't come up with any good reasons to why you would not want that functionality on a search profile.

#### Checklist

### DIFF
--- a/modules/ting_field_search/plugins/ting_field_search_ctools_export_ui.inc
+++ b/modules/ting_field_search/plugins/ting_field_search_ctools_export_ui.inc
@@ -125,12 +125,6 @@ function ting_field_search_profile_export_form(&$form, &$form_state) {
     '#default_value' => $profile->config['search_request']['new_materials'],
     '#element_validate' => array('element_validate_integer_positive'),
   );
-  $form['config']['search_request']['allow_empty'] = array(
-    '#type' => 'checkbox',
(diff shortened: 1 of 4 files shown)


holt83:
Regarding the behat test fail: I get the following error:

`"getMessages" method is not available on the SearchPage (BadMethodCallException)`

It seems to call a non-existing method here: https://github.com/ding2/ding2/blob/master/tests/behat/features/bootstrap/StepsContext.php#L73

What to do?

holt83:
Investigating further it seems to be because it didn't find any search results here: https://github.com/ding2/ding2/blob/master/tests/behat/features/bootstrap/Page/SearchPage.php#L591 for the search:

`term.language=dansk AND term.type=bog  AND term.title=?æ*`

I have verified that the classes are correct and in my local install I get a search result.

holt83:
Nevermind above comments.

I did a rerun of the workflow on circle and now it has passed :)

PR #1245, opened 2018-11-01
3893: Add and enable expanded to node menu settings

BirkAndMe:
#### Link to issue

https://platform.dandigbib.org/issues/3893

#### Description

Add and enable the expanded checkbox to the menu settings part of a node form.

#### Screenshot of the result

screenshot_2018-11-01 edit side samtykke og opbevaring af information tidligere lan pbj ding virt b14cms dk

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1248, opened 2018-11-06
3639 by Inlead: Remove authorship checkbox.

AnatolyMuntean:
#### Link to issue ####
https://platform.dandigbib.org/issues/3639

#### Description ####
Simply remove the authorship checkbox and push all content with FALSE authorship value.
This change will display the authorship in backend (when syndicating) and not in frontend (after syndication). Normally authorship would appear as a suffix to the body value, which is not the case now.
For existing nodes that were pushed with authorship flag enabled, this also involves service changes to make sure no authoring information is appended to the content.

#### Screenshot of the result ####
screenshot from 2018-11-06 12-44-35

### DIFF
--- a/modules/bpi/bpi.module
+++ b/modules/bpi/bpi.module
@@ -854,7 +854,6 @@ function bpi_form_workflow_transition_form_submit($form, &$form_state) {
     'bpi_push_category',
     'bpi_push_audience',
     'bpi_push_images',
-    'bpi_push_ccl',
     'bpi_push_editable',
     'bpi_push_refs',
   );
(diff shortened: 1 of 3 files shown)


cableman:
Code approved, would it be possible for you to re-run the behat tests at CI.

guddo:
@cableman We do not have permissions to trigger builds on CirlceCI.

kasperg:
@guddo Could you please double check this?

When I access the rebuild menu I get a message that I need write permissions to trigger rebuilds. I think that may be because the build occurs in easySuite context as indicated by the path to the build on CircleCI: https://circleci.com/gh/easySuite/ding2/1503. In that context you should have write permissions.

ding2 1503 - circleci 2018-11-28 07-50-36

PR #1253, opened 2018-11-07
3703: Fix ekurser date sort

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/3703

#### Description

Opensearch is using date field for sorting, but ding_ekurser was using acquisitionDate. Before the introduction of SAL this was no problem, but now Ting search has to verify that the provider supports the current sort. Since opensearch is not returning acquisitionDate it's ignored bothwhen performing search and setting default value.

date = The date the material was published.
acquisitionDate = The date the material was added to the library.

Even though the Opensearch webservice is supporting acquisitionDate (and also localAcquisitionDate), the simplest and best solution seems to be just make ding_ekurser use date field for sort. I think it's the most intuitive sort and it prevents us from having to introduce acquisitionDate in the main search functionality by adding it to `opensearch_search_sort_options()`.

Since ding_ekurser is dependendant on opensearch to work correctly I also added a dependency.

#### Checklist

### DIFF
--- a/modules/ding_ekurser/ding_ekurser.info
+++ b/modules/ding_ekurser/ding_ekurser.info
@@ -5,6 +5,7 @@ project = ding_ekurser
 dependencies[] = ctools
 dependencies[] = oembed
 dependencies[] = page_manager
+dependencies[] = opensearch
 features[ctools][] = oembed:oembed_provider:1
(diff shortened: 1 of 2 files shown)


holt83:
I included a commen in the commit message and Gitcop is complaining.

PR #1255, opened 2018-11-08
3816 by Inlead: Filter by taxonomy in /content.

bozeak:
#### Link to issue

https://platform.dandigbib.org/issues/3816

#### Description

As it was requested, was added possibility to filter nodes on /content page by taxonomy terms. The taxonomy term filtration depends on selected taxonomy vocabulary filter from the same form.
New filter is realized as a textfield which will be in disabled state until taxonomy vocabulary filter is not selected. If an taxonomy vocabulary is selected, terms filter changes it's state to active and autocomplete functionality is added.

Regarding opportunity to assign taxonomies through bulk update, this is already realized in VBO actions and can be done with "Change value" action.

#### Screenshot of the result

content-view_voc-term

#### Checklist

### DIFF
--- a/modules/ding_base/ding_base.module
+++ b/modules/ding_base/ding_base.module
@@ -25,6 +25,13 @@ function ding_base_menu() {
     'file path' => drupal_get_path('module', 'system'),
   );
 
+  $items['ding_base/taxonomy/autocomplete'] = array(
+    'title' => 'Autocomplete taxonomy',
+    'page callback' => 'ding_base_taxonomy_autocomplete',
+    'access arguments' => array('access content'),
+    'type' => MENU_CALLBACK,
+  );
+
   return $items;
 }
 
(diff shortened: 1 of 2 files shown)


modules/ding_base/ding_base.module
```diff
@@ -750,3 +757,139 @@ function ding_base_validate_email($email) {
   $domain = strtoupper(substr($email, strrpos($email, '.') + 1));
   return empty($tld_list) || in_array($domain, $tld_list);
 }
+
+/**
+ * Implements hook_views_pre_view().
+ */
+function ding_base_views_pre_view(&$view, &$display_id, &$args) {
```
holt83: Are you using this hook to add filter to the admin view and are you sure this is the correct hook to use?

In the documentation it's mentioned that this hook is only run for previews: https://api.drupal.org/api/views/views.api.php/function/hook_views_pre_view/7.x-3.x

FYI: The BPI module contains an example of adding filters to the admin node view via `hook_views_default_views_alter()`

modules/ding_base/ding_base.module
```diff
@@ -750,3 +757,139 @@ function ding_base_validate_email($email) {
...
+        'class' => [
+          'ctools-auto-submit-exclude',
+        ],
+      ];
+    }
+  }
+}
+
+/**
+ * Implements hook_preprocess_HOOK().
```
cableman: Should this be hook_preprocess_views_exposed_form()?

cableman: It would also be nice with a comment about why the process is needed? Is it to change the input fields order?

modules/ding_base/ding_base.module
```diff
@@ -750,3 +757,139 @@ function ding_base_validate_email($email) {
...
+    $variables['widgets'] = $widgets;
+  }
+}
+
+/**
+ * Admin content "Terms" filter autocomplete callback.
+ */
+function ding_base_taxonomy_autocomplete($vid, $tag) {
+  if (!empty($vid)) {
+    $vocabulary = taxonomy_vocabulary_load($vid);
```
cableman: Why load the vocabulary here? It's not used in the function!

modules/ding_base/js/ding_base.prevent_auto_submit.js
```diff
@@ -0,0 +1,39 @@
...
+
+  Drupal.behaviors.prevent_auto_submit = {
+    attach: function(context) {
+      // Processing autocomplete functionality for "Terms" filter field.
+      $("input[name='ding_base_term_name']", context).on('change keyup', function (e) {
+        if (e.type === 'change' || (e.type === 'keyup' && e.which === 13)) {
+          var form = $(this).closest('form');
+          var selected;
+
+          if ($(form).find('#autocomplete ul li.selected').length !== 0) {
```
cableman: Maybe move "$(form).find('#autocomplete ul li.selected')" into a variable!

modules/ding_base/ding_base.module
cableman: If it's not the case a @holt83 writes please add an comment to the code about why this is used and not the other hook.

bozeak: + Good point. Thanks, fixed this.

modules/ding_base/ding_base.module
bozeak: No, unfortunately there is no such hook. Added description into DocBlock.

modules/ding_base/ding_base.module
bozeak: + My bad. Removed.

modules/ding_base/js/ding_base.prevent_auto_submit.js
bozeak: +

PR #1259, opened 2018-11-09
3914: Enable campaign plus as default

cableman:
#### Link to issue

https://platform.dandigbib.org/issues/3914

#### Description

Enables campaign plus as default.

#### Screenshot of the result

No screenshot

#### Checklist

No comments


JrgenGNielsen:
revieved and approved

PR #1260, opened 2018-11-09
3830: Remove list from public lists

BirkAndMe:
#### Link to issue

https://platform.dandigbib.org/issues/3830

#### Description

Allow admins to hide user lists from public lists

#### Screenshot of the result

screenshot_2018-11-09 huskeliste pbj ding virt b14cms dk

#### Checklist

### DIFF
--- a/modules/p2/ding_list/ding_list.module
+++ b/modules/p2/ding_list/ding_list.module
@@ -126,6 +126,14 @@ function ding_list_menu() {
     'type' => MENU_CALLBACK,
   );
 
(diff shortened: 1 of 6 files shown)


modules/p2/ding_list/ding_list.module
```diff
@@ -126,6 +126,14 @@ function ding_list_menu() {
     'type' => MENU_CALLBACK,
   );
 
+  $items['list/%/toggle-not-public'] = array(
+    'title' => 'Show carousel results',
```
cableman: Is this title correct?

PR #1261, opened 2018-11-09
3828: workmatch memory

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/3828

Related PR: https://github.com/ding2/ting-client/pull/25

#### Description

We used a special 'manifestation' search request to translate fausts to pids, but this was interfering with caching of collections, such that the manifestation would be cached as single object collections. This would result in material types getting "hidden" from users in the UI, because of these incomplete collections.

Switching to getObject request instead of search will fix this (in situations where the translation is the culprit) and is also a better solution in my opinion. 

There might still be problems with incomplete collection in other places, but this is out of scope of this PR.

Also fixes issue with get_objects, when over 169 objects are requested. See this comment for more info about this: https://platform.dandigbib.org/issues/3828#note-16

#### Checklist

This PR is dependant on https://github.com/ding2/ting-client/pull/25 to work correctly.

### DIFF
--- a/modules/opensearch/includes/opensearch.search.inc
+++ b/modules/opensearch/includes/opensearch.search.inc
@@ -623,31 +623,13 @@ function opensearch_search_sort_options() {
  */
 function opensearch_search_object_ids($ids) {
   if (is_array($ids) && !empty($ids)) {
-    $translated_ids = array();
+    $translated_ids = [];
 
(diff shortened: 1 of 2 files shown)


holt83:
It appears that circle didn't run for this PR?

JrgenGNielsen:
reviewed and approved

PR #1262, opened 2018-11-13
3889 by Inlead: Add ding_nodelist to ding2 core.

inlead:
#### Link to issue

https://platform.dandigbib.org/issues/3889

#### Description

Moved ding_nodelist from [easySuite/ding_nodelist.git](https://github.com/easySuite/ding_nodelist) to ding2 core.

#### Checklist

Previous issue can be seen at https://platform.dandigbib.org/issues/1597.


NicolaeGrigoriu:
Currently scrutinizer issues are regarding possition of the php code in template. Based on drupal coding standards, this is the way it needs to be and not as scrutinizer asks.

JrgenGNielsen:
reviewed and approved

PR #1266, opened 2018-11-15
3922: Aggregates news-list view results and removes group criteria

b14sha:
#### Link to issue

https://platform.dandigbib.org/issues/3922

#### Description

The news-list now aggregates results to remove duplicates when a news node is related to more than one library group. Also the filter criteria that a node needs a library-group relationship has been removed, to allow news to be shown that is not connected to any specific library.

#### Comments
It seems it fails CI behat_test, but since the only changes is made in a ding_news feature, and the fail is in 01-Searching.feature:57, it seems to be unrelated to the contents of this commit.

### DIFF
--- a/modules/ding_news/ding_news.strongarm.inc
+++ b/modules/ding_news/ding_news.strongarm.inc
@@ -179,13 +179,6 @@ function ding_news_strongarm() {
   $strongarm = new stdClass();
   $strongarm->disabled = FALSE; /* Edit this to true to make a default strongarm disabled initially */
   $strongarm->api_version = 1;
-  $strongarm->name = 'flag_mobile_app_default_ding_news';
-  $strongarm->value = 1;
-  $export['flag_mobile_app_default_ding_news'] = $strongarm;
-
-  $strongarm = new stdClass();
-  $strongarm->disabled = FALSE; /* Edit this to true to make a default strongarm disabled initially */
-  $strongarm->api_version = 1;
   $strongarm->name = 'language_content_type_ding_news';
   $strongarm->value = '0';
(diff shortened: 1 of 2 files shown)


modules/ding_news/ding_news.strongarm.inc
```diff
@@ -176,13 +176,6 @@ function ding_news_strongarm() {
   $strongarm->value = 1;
   $export['flag_push_to_mongo_default_ding_news'] = $strongarm;
 
-  $strongarm = new stdClass();
-  $strongarm->disabled = FALSE; /* Edit this to true to make a default strongarm disabled initially */
-  $strongarm->api_version = 1;
-  $strongarm->name = 'flag_mobile_app_default_ding_news';
-  $strongarm->value = 1;
-  $export['flag_mobile_app_default_ding_news'] = $strongarm;
```
kasperg: Why has this variable been removed? It does not seem to be related to the change at hand. As far as I can tell it relates to the use of the `mobile_app` flag for Ding News.

I digged into this and it seems as if the situation is caused by #1095 adding the variable to the strongarm definition but not updating the info file accordingly. Thus Features thinks the value should be removed from the export.

I have created a ticket for adding this: https://platform.dandigbib.org/issues/3966. For this PR the change to this file should simply be removed.

b14sha: My local installation failed completely on rebase. So had to do a fresh install, and therefore it seemed easier to do the changes again from a fresh installation. So - please disregard this PR, and check the changes here: https://github.com/ding2/ding2/pull/1276

PR #1267, opened 2018-11-15
3923: Fixed Ding news render empty materials ref

b14sha:
#### Link to issue

https://platform.dandigbib.org/issues/3923

#### Description

ting_reference default formatter attached js even when the field was empty, that causes the field to be rendered. Now the js is only attached if the field has content.


JrgenGNielsen:
reviewed and approved

PR #1271, opened 2018-11-20
3905: Cache multiple objects per collection

kasperg:
#### Link to issue

https://platform.dandigbib.org/issues/3905

#### Description

This PR is based on the work from #1269. Only the last three commits should be considered.

OpenSearch supports retrieving many objects in a single request by
passing multiple ids.

This results in a unique request for all these ids and thus a unique
cache entry despite the fact that information for some or all of these
objects may already be cached based on individual requests.

Consequently we update our cache handling to split all object requests 
into multiple single id object requests, check each request for a
cached version and if found merge these cached responses into the
combined response.

To facilitate this cache handling is moved to single function combining
retrieval and insertion as well as static and persistent caching.

We also cache all objects in a collection when predicting cache entries.
There does not seem to be a reason to only cache the first one.

With this change displaying search results from an empty cache is reduced 
to a single request to OpenSearch.

#### Checklist

### DIFF
--- a/modules/opensearch/opensearch.client.inc
+++ b/modules/opensearch/opensearch.client.inc
@@ -355,28 +355,48 @@ function opensearch_execute(TingClientRequest $request) {
     $request->setRelationData('full');
   }
 
-  // Build cache id for this request.
-  $cid = _opensearch_cache_id($request);
-
   // Check static cache first as this is the fasts lookup. We use the fast
   // static pattern described on drupal.org in the drupal_static
   // documentation.
   static $opensearch_static_fast = array();
-  if (array_key_exists($cid, $opensearch_static_fast)) {
-    return $opensearch_static_fast[$cid];
+
+  if ($response = _opensearch_cache_get($request, $opensearch_static_fast)) {
+    return $response;
   }
 
-  // Check the database/memcached cache for the request also check that the
-  // cache data has not expired before using it.
-  $cache = cache_get($cid, 'cache_opensearch');
-  if ($cache && ($cache->expire > REQUEST_TIME)) {
-    $data = $cache->data;
+  // If we are retrieving multiple objects then there might not be a single
+  // request for all these objects combined in the cache but they might be
+  // cached individually.
+  $cached_objects = [];
+  if ($request instanceof TingClientObjectRequest) {
+    $ids = array_combine($request->getObjectIds(), $request->getObjectIds());
 
-    // Update the static cache with the database cache content to speed up the
-    // next request for the same data.
-    $opensearch_static_fast[$cid] = $data;
+    // Try to retrieve an object for each id. Entries which does not have cached
+    // data will get the value FALSE for easy subsequent filtering.
+    $object_ids = array_map(function ($id) use ($request, $opensearch_static_fast) {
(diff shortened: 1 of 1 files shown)


modules/opensearch/opensearch.client.inc
```diff
@@ -532,6 +558,40 @@ function _opensearch_cache_insert(TingClientRequest $request, $response, array &
   }
 }
 
+/**
+ * Retrieve a earch response based on a request.
```
cableman: nitpicking: missing "s" in search in the function comment.

modules/opensearch/opensearch.client.inc
```diff
@@ -355,28 +355,48 @@ function opensearch_execute(TingClientRequest $request) {
...
-  // Check the database/memcached cache for the request also check that the
-  // cache data has not expired before using it.
-  $cache = cache_get($cid, 'cache_opensearch');
-  if ($cache && ($cache->expire > REQUEST_TIME)) {
-    $data = $cache->data;
-
-    // Update the static cache with the database cache content to speed up the
-    // next request for the same data.
-    $opensearch_static_fast[$cid] = $data;
+  if ($response = _opensearch_cache_get($request, $opensearch_static_fast)) {
```
cableman: Maybe a comment that if an exact match for the $request is found fast-track and return the result.

modules/opensearch/opensearch.client.inc
```diff
@@ -532,6 +558,40 @@ function _opensearch_cache_insert(TingClientRequest $request, $response, array &
...
+    $data = $cache->data;
+
+    // Update the static cache with the database cache content to speed up the
+    // next request for the same data.
+    $opensearch_static_fast[$cid] = $data;
+
+    return $data;
+  }
+
+  return NULL;
```
cableman: Note: in the old cache we hade a special value for "empty cache" as "null" would be cached. I don't known if this an issue in the new cache...

cableman: Has I remember it was in relation to "0-hits" searches.

kasperg: I do not think the suggested change makes the situation any better or worse.

modules/opensearch/opensearch.client.inc
kasperg: This DocBlock has been removed in a subsequent commit so I do not think it is worth fixing.

modules/opensearch/opensearch.client.inc
```diff
@@ -532,64 +526,55 @@ function _opensearch_predict_cache(array $collections, array &$opensearch_static
...
+  static $static_cache;
   $cid = _opensearch_cache_id($request);
-  if (array_key_exists($cid, $opensearch_static_fast)) {
-    return $opensearch_static_fast[$cid];
+
+  // Count the number of arguments to determine if $value has been provided.
+  // This allows us to distinguish between situations where $value is not set
+  // and we want to retrieve the cache value and where $value is set to NULL
+  // explicitly.
+  if (func_num_args() == 2) {
```
cableman: 👍 nice with the comment.

PR #1273, opened 2018-11-23
3573: Trying to fix webtrekk race-condition

cableman:
#### Link to issue

https://platform.dandigbib.org/issues/3573

#### Description

There is a race-condition between when campaigns are loaded and the tracking code from web-trekk. So this is the best take on trying to fix it for campaigns.

A very simple queue is added and when window.wt becomes available the events in the queue is pushed to webtrekk. 

#### Screenshot of the result

No screenshot.

#### Checklist

This PR have not be tested as web-trekk is not able to load on local development machines. So I did not have any way to test it.

### DIFF
--- a/modules/ding_campaign_plus/js/ding_campaign_plus_trekk.js
+++ b/modules/ding_campaign_plus/js/ding_campaign_plus_trekk.js
@@ -5,6 +5,8 @@
 (function ($) {
   'use strict';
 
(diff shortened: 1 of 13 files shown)


cableman:
The errors in scrutinize is JS error with used before defined. But you can't cancel in the interval...

PR #1278, opened 2018-11-29
153 by Inlead: Disable ability to syndicate own content.

AnatolyMuntean:
#### Link to issue

https://platform.dandigbib.org/issues/153

#### Description

Simply hide the `Syndicate` link in the result set for the content that has the same agency as the current bpi setup.

#### Screenshot of the result

image

#### Checklist

- [ ] My code passes continuous integration process. If not then I have added a comment explaining why this change should be exempt from the code standards and process. _Unknown reason._


JrgenGNielsen:
reviewed and approved

PR #1280, opened 2018-11-29
3905: Ensure static OpenSearch cache is always initialized

kasperg:
#### Link to issue

https://platform.dandigbib.org/issues/3905

#### Description

If it is not then we risk passing NULL to array_key_exists(). This
will issue a warning.

#### Checklist

### DIFF
--- a/modules/opensearch/opensearch.client.inc
+++ b/modules/opensearch/opensearch.client.inc
@@ -560,7 +560,7 @@ function _opensearch_predict_cache(array $collections, TingClientRequest $reques
  *   The cached response. NULL if no cached entry exists.
  */
 function _opensearch_cache(TingClientRequest $request, $response = NULL) {
-  static $static_cache;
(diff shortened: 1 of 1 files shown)


modules/opensearch/opensearch.client.inc
```diff
@@ -561,6 +561,9 @@ function _opensearch_predict_cache(array $collections, TingClientRequest $reques
  */
 function _opensearch_cache(TingClientRequest $request, $response = NULL) {
   static $static_cache;
```
cableman: Is it not possible to simply do:

`static $static_cache = array();`

http://php.net/manual/en/language.variables.scope.php#language.variables.scope.static see under "Example #7 Declaring static variables"

cableman: Or "Example #5 Example use of static variables"

PR #1283, opened 2018-12-03
3901: Webtrekk MWA traking

holt83:
#### Link to issue

#### Description

In this PR the tracking capabilities of ding_webtrekk is extended to include several new Most Wanted Actions (MWAs) like reservation, rating, carousel click etc. The implementation follows the guidelines for parameter names and values from following spreadsheet, which also lists all the tracking that is added in this PR (those marked with "Skal implementeres"):

[Webtrekk 20171128 - MWA list](https://docs.google.com/spreadsheets/d/1OdUWGQc_hiY8F6BPViKRrBykSQfHLnhv4-YMdUzswSI/edit#gid=0)

With Webtrekk we have three methods of tracking:

**Page parameters**
These are used for information about the page that are known at load time and are only related to that page. For example what is the type of the material being shown or what sort was used on the current search result.
These are added as properties on the global _ti object. Example:
```
window._ti = window._ti || {};
window._ti['OSS'] = 'Krimi';
window._ti['OSSr'] = '5099';
window._ti['p_s_Page'] = '1';
window._ti['p_s_Sort'] = 'date_descending';
window._ti['p_OSSprofile'] = 'boeger'
```

**URL parameters**
These are used when we want track relations between pages. For example, did the material view originate from a carousel and if so which one? On the ting_search_carousel the "Read more"-button and Reserve-button will have following parameter appended to their URL:
/ting/object/870970-basis%3A29997829?u_navigatedby=Unge
Where "Unge" is the title of the carousel. 

**Events**
For user interactions that doesn't result in a page load Webtrekk's event functionality is used. Examples is the carousel prev/next buttons and the reserve button which uses AJAX to make the reservation. It's implemented by calling the push() method on the wts Webtrekk-object and passing JSON-formatted event data to the function. Example:
```
wts.push(['send', 'click', {
  linkId:'Karousel, click på næste knappen',
  customClickParameter:{ 59:'Biblioteket foreslår materialer til dig'}
}]);
```

I have tried to handle as much on the server as possible, since there's much more reliable and convienient access to data. Events can be completely setup and attached on the server using HTML data-attribute, and a single catch-all click-handler will take care of pushing the events in the client.

Some things like autocomplete events are dependant on dynamic data created when the user is interacting, and therefore needs to be handled in client JS. Other stuff proved very problematic to handle on the server and is also handled in the client. For example, the ding_carousel theme often doesn't even know the name of it's carousel and it would require a rework or a hacky approach, to get the needed value for tracking parameter.

#### Checklist

The PR also includes some fixes for some critical stuff:

1. The Webtrekk loading script was incorrectly placed in the footer of the page. I have moved it to the 
top of the head together with the page parameters as mentioned in the implementation guidelines: [Webtrekk-Implementation-Guidelines](https://support.webtrekk.com/hc/en-us/article_attachments/115007805409/Webtrekk-Implementation-Guidelines-v6-EN.PDF). Also mentioned  in this PR: https://github.com/ding2/ding2/pull/1274
2. The internal node search was never meant to be tracking with separate parameters, resulting in these parameters never being actually send to webtrekk. The same parameters that is used for ting search should be used (OSS and OSSr).
3. The code that was adding the page-parameter-script filterede out values zero-values and as a result important tracking of zero-hit searches has been lost and never tracked in Webtrekk. I have updated the code to still ensure no empty values, but make exceptions for the values 0 and '0'.

### DIFF
--- a/modules/ding_webtrekk/ding_webtrekk.js
+++ b/modules/ding_webtrekk/ding_webtrekk.js
@@ -19,7 +19,6 @@
     return link + seperator + urlParameter;
   };
 
-
   Drupal.behaviors.ding_webtrekk = {
     attach: function(context) {
       // Attach Webtrekk events. This is the preffered to track events in ding
@@ -46,6 +45,18 @@
             });
           });
         })
+      });
+
+      // Track autocomplete selections.
+      $('.form-item-search-block-form .form-autocomplete', context)
+        .once('ding-webtrekk')
+        .on('autocompleteSelect', function(e, selected) {
+          if (($(selected).text())) {
+            wt.sendinfo({
+              type: 'e_autocomplete_search',
+              text: $(selected).text()
+            });
+          }
       });
 
       // Special handling for ding_carousel. The primary reason for this is that
--- a/modules/ding_webtrekk/ding_webtrekk.js
+++ b/modules/ding_webtrekk/ding_webtrekk.js
@@ -39,6 +39,7 @@
       $('.ding-webtrekk-rating-event', context).once('ding-webtrekk', function() {
         var contentId = $(this).data('ding-entity-rating-id');
         $('.js-rating-symbol', this).each(function(index) {
+          var rating = index + 1;
           $(this).click(function() {
             wts.push(['send', 'click', {
               linkId:'Materiale rating',
--- a/modules/ding_webtrekk/ding_webtrekk.js
+++ b/modules/ding_webtrekk/ding_webtrekk.js
@@ -93,14 +93,14 @@
 
         $('.slick-arrow', this).once('ding-webtrekk').click(function() {
           var linkId = 'Karousel, click på forrige knappen';
-          var clickParameter = 60;
+          var customClickParameter = { 60: carouselTitle };
           if ($(this).hasClass('slick-next')) {
             linkId = 'Karousel, click på næste knappen';
-            clickParameter = 59;
+            var customClickParameter = { 59: carouselTitle };
           }
           wts.push(['send', 'click', {
             linkId: linkId,
-            customClickParameter: { clickParameter: carouselTitle }
+            customClickParameter: customClickParameter
           }]);
         });
       });
--- a/modules/ding_webtrekk/ding_webtrekk.js
+++ b/modules/ding_webtrekk/ding_webtrekk.js
@@ -63,6 +63,24 @@
           }
       });
 
+      // Track ding loan renew selected events. The renew all button is attached
+      // completely on the server, but for renew selected, we need information
+      // about selected loans in the UI, before we send event.
+      $('.ding-webtrekk-event-renew-selected', context)
+        .once('ding-webtrekk')
+        .click(function(e) {
+          var selectedMaterials = [];
+          $('.material-item input[type=checkbox]:checked').each(function() {
+            // We have collected the material ids (pids) in our event data
+            // attribute for each select box.
+            selectedMaterials.push($(this).data('ding-webtrekk-event'));
+          });
+          wts.push(['send', 'click', {
+            linkId: 'Forny valgte materialer',
+            customClickParameter: { 55: selectedMaterials.join(';') }
+          }]);
+      });
+
       // Special handling for ding_carousel. The primary reason for this is that
       // we need the title of the carousel to pass as value in event and URL-
       // parameter. For tabbed carousels the title of the individuals carousels
--- a/modules/ding_webtrekk/ding_webtrekk.js
+++ b/modules/ding_webtrekk/ding_webtrekk.js
@@ -22,7 +22,10 @@
 
   Drupal.behaviors.ding_webtrekk = {
     attach: function(context) {
-      // Attach Webtrekk events.
+      // Attach Webtrekk events. This is the preffered to track events in ding
+      // webtrekk. Modules can alter elements in the backend and add the class
+      // 'ding-webtrekk-event' to enable event traking for that element. Event
+      // data is expected to be passed in the data attribute on the element.
       $('.ding-webtrekk-event', context).once('ding-webtrekk').click(function() {
         wt.sendinfo($(this).data('ding-webtrekk-event'));
       });
--- a/modules/ding_webtrekk/ding_webtrekk.js
+++ b/modules/ding_webtrekk/ding_webtrekk.js
@@ -69,6 +69,14 @@
       $('.ding-webtrekk-event-renew-selected', context)
         .once('ding-webtrekk')
         .click(function(e) {
+          // The renew-all button is implemented by checking all renewable loans
+          // and then trigger a click on the renew-selected button. We can use
+          // this jQuere marker to ensure we don't send two events in this case.
+          // See: https://stackoverflow.com/a/10704256
+          if (e.isTrigger) {
+            return;
+          }
+
           var selectedMaterials = [];
           $('.material-item input[type=checkbox]:checked').each(function() {
             // We have collected the material ids (pids) in our event data
--- a/modules/ding_webtrekk/ding_webtrekk.js
+++ b/modules/ding_webtrekk/ding_webtrekk.js
@@ -13,6 +13,14 @@
 
   "use strict";
 
+  // Webtrekk tracking parameters.
+  var DING_WEBTREKK_PARAMETER_AUTOCOMPLETE = 54;
+  var DING_WEBTREKK_PARAMETER_RENEW_SELECTED = 55;
+  var DING_WEBTREKK_PARAMETER_RENEW_RATING_ID = 57;
+  var DING_WEBTREKK_PARAMETER_RENEW_RATING = 58;
+  var DING_WEBTREKK_PARAMETER_CAROUSEL_NEXT = 59;
+  var DING_WEBTREKK_PARAMETER_CAROUSEL_PREV = 60;
+
   var appendQueryParameter = function(url, key, value) {
     var seperator = (url.indexOf('?') !== -1) ? '&' : '?';
     var urlParameter = key + '=' + encodeURIComponent(value);
@@ -41,8 +49,8 @@
             wts.push(['send', 'click', {
               linkId: 'Materiale rating',
               customClickParameter: {
-                58: rating,
-                57: contentId
+                DING_WEBTREKK_PARAMETER_RENEW_RATING: rating,
+                DING_WEBTREKK_PARAMETER_RENEW_RATING_ID: contentId
               }
             }]);
           });
@@ -54,9 +62,12 @@
         .once('js-ding-webtrekk')
         .on('autocompleteSelect', function(e, selected) {
           if (($(selected).text())) {
+            console.log($(selected).text());
             wts.push(['send', 'click', {
               linkId: 'Autocomplete søgning clicks',
-              customClickParameter: { 54: $(selected).text() }
+              customClickParameter: {
+                DING_WEBTREKK_PARAMETER_AUTOCOMPLETE: $(selected).text()
+              }
             }]);
           }
       });
@@ -84,7 +95,9 @@
           });
           wts.push(['send', 'click', {
             linkId: 'Forny valgte materialer',
-            customClickParameter: { 55: selectedMaterials.join(';') }
+            customClickParameter: {
+              DING_WEBTREKK_PARAMETER_RENEW_SELECTED: selectedMaterials.join(';')
+            }
           }]);
       });
 
@@ -119,10 +132,14 @@
 
         $('.slick-arrow', this).once('js-ding-webtrekk').click(function() {
           var linkId = 'Karousel, click på forrige knappen';
-          var customClickParameter = { 60: carouselTitle };
+          var customClickParameter = {
+            DING_WEBTREKK_PARAMETER_CAROUSEL_PREV: carouselTitle
+          };
           if ($(this).hasClass('slick-next')) {
             linkId = 'Karousel, click på næste knappen';
-            customClickParameter = { 59: carouselTitle };
+            customClickParameter = {
+              DING_WEBTREKK_PARAMETER_CAROUSEL_NEXT: carouselTitle
+            };
           }
           wts.push(['send', 'click', {
             linkId: linkId,
--- a/modules/ding_webtrekk/ding_webtrekk.js
+++ b/modules/ding_webtrekk/ding_webtrekk.js
@@ -5,13 +5,8 @@
  * Some user interactions that we want to track doesn't result in a page load.
  * These interactions are tracked here with Webtrekk events in the client.
  *
- * URL-parametes are used when we want to track relations. For example, was this
- * ting object view triggered by a link in a carousel? We prefer to attach these
- * parameters on the server, but in some cases this gets very complicated. For
- * example, all the object links in a carousel are not loaded all at once on the
- * page load, but is instead lazy loaded dynamically when the users uses the
- * carousel. In such cases, it's much easier to add the URL-parameters in
- * javascript on the client.
+ * Also handles has special handling for some elements, which are problematic
+ * to handle in the backend.
  */
 
 (function($) {
@@ -32,9 +27,9 @@
         wt.sendinfo($(this).data('ding-webtrekk-event'));
       });
 
-      // Unfortunately we need special handling for ding entity rating event. It
-      // would require a complete rework of the ding_entity_rating_display theme
-      // hook to avoid this.
+      // Secial handling for ding entity rating event. It would require a
+      // complete rework of the ding_entity_rating_display theme hook to avoid
+      // this.
       $('.ding-webtrekk-rating-event', context).once('ding-webtrekk', function() {
         var contentId = $(this).data('ding-entity-rating-id');
         $('.js-rating-symbol', this).each(function(index) {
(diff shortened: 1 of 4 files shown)


holt83:
It should be ready for renew now. 

Scrutinizer complains about wts is not defined, but it's a global variable added by the Webtrekk script. Could maybe check if it exists in top of attach function and return early if it's not defined? Don't know if that would fix scrutinizer also.

holt83:
Ok, the scrutinizer errors magically dissapeared. I just love when that happens! :D

holt83:
I don't know what to do about that array indentation problem Scrutinizer keeps complaining about.

modules/ding_webtrekk/ding_webtrekk.js
```diff
@@ -0,0 +1,131 @@
...
+        else if ($(this).parent().siblings('.pane-title').length > 0) {
+          carouselTitle = $(this).parent().siblings('.pane-title').html();
+        }
+
+        // Add tracking URL-pararmeters to items in the carousel.
+        $('.ding-carousel-item .ting-object a[href^="/ting/object/"]', this).once('ding-webtrekk', function() {
+          $(this).attr('href', appendQueryParameter($(this).attr('href'), key, carouselTitle));
+        });
+
+        $('.slick-arrow', this).once('ding-webtrekk').click(function() {
```
kasperg: Is there a reason why you do not use [the events provided by Slick](http://kenwheeler.github.io/slick/) for this?

modules/ding_webtrekk/ding_webtrekk.module
```diff
@@ -51,24 +46,92 @@ function ding_webtrekk_page_alter(&$page) {
         'option: {}
       };' . $tag_integration_logic,
       'type' => 'inline',
-      'scope' => 'footer',
+      'group' => JS_LIBRARY,
+      'weight' => -21,
```
kasperg: Please add a comment about why this weight is set as it is. It is often in reference to a weight set by another element.

modules/ding_webtrekk/ding_webtrekk.module
```diff
@@ -51,24 +46,92 @@ function ding_webtrekk_page_alter(&$page) {
...
+      // logged in. So by also requiring a logged in user before setting parameter
+      // we reduce the chances of false positives. For example, it will prevent
+      // random access to the URL from anonymous users, from being tracked.
+      if (strpos(current_path(), DING_GATEWAYF_REGISTRATION_SUCCESS_URL) === 0) {
+        $params['p_usercreate'] = 'NemId';
+      }
+    }
+    // Fallback to cpr+pin if none the modules we know about was used to log in.
+    else {
+      $params['p_isloggedin'] = 'cpr+pinkode';
```
kasperg: Note that this will also track Drupal users as logged in with CPR+Pin. I do not know if that is intentional?

modules/ding_webtrekk/ding_webtrekk.module
```diff
@@ -51,24 +46,92 @@ function ding_webtrekk_page_alter(&$page) {
...
         'OSS' => $search_result->getSearchRequest()->getFullTextQuery(),
         'OSSr' => $search_result->getNumTotalObjects(),
       ];
-      _ding_webtrekk_set_data_layer_values($params);
+
+      // We want to track facets that users have added themselves from the UI.
+      // Looking for the special URL parameter that is added in these cases
+      // seems to be more reliable than looking on the filters added to the
+      // search request object.
+      if (!empty($_GET['facets']) && is_array($_GET['facets'])) {
```
kasperg: I appreciate the comment but I am skeptical about the idea. Do you have an example of a situation where going through the filters would provide unreliable data?

If accessing the search object for facets is unreliable would it not be the same face for the full text query?

modules/ding_webtrekk/ding_webtrekk.module
```diff
@@ -51,24 +46,92 @@ function ding_webtrekk_page_alter(&$page) {
...
+        $current_sort = $search_result->getSearchRequest()->getSorts()[0];
+        $params['p_s_Sort'] = ting_search_sort_key_from_sort($current_sort);
+      }
+      else {
+        $default_sort = variable_get('ting_search_default_sort', '');
+        // If empty string the default "sort" is ranking. If search provider is
+        // opensearch we know how to get the actual ranking used. If not we will
+        // use a generic 'rank' value for the parameter.
+        if ($default_sort === '') {
+          $params['p_s_Sort'] = 'rank';
```
kasperg: Why not just set the default value for `ting_search_default_sort` to `rank`:

```php
$default_sort = variable_get('ting_search_default_sort', 'rank');
```

modules/ding_webtrekk/ding_webtrekk.module
```diff
@@ -90,35 +153,170 @@ function ding_webtrekk_views_post_execute(&$view) {
...
+
+  // The page parametes for the ting object view should only be added on the
+  // full page view for the material.
+  if ($view_mode == 'full' && current_path() == $entity_path) {
+    $params = [
+      'p_mat_type' => $entity->getType(),
+      'p_mat_indexno' => $entity->getClassification(),
+      'p_mat_lang' => $entity->getLanguage(),
+      'p_mat_source' => $entity->getAc_source(),
+      'p_mat_category' => implode(';', $entity->reply->getGenre()),
```
kasperg: Is there a reason to access `reply` here? I would expect you to be able to call the method directly on the entity as you do for the other parameters.

modules/ding_webtrekk/ding_webtrekk.module
```diff
@@ -90,35 +153,170 @@ function ding_webtrekk_views_post_execute(&$view) {
...
+    foreach (element_children($entity->content['ding_entity_buttons']) as $delta) {
+      foreach ($entity->content['ding_entity_buttons'][$delta] as $key => &$button) {
+        // Reserve button.
+        if (isset($button['#path']) && $button['#path'] == $entity_path . '/reserve') {
+          $button['#options']['attributes']['class'][] = 'ding-webtrekk-event';
+          $event_data = json_encode([
+            'linkId' => 'Reserver',
+            'customClickParameter' => [50 => $entity->getId()],
+          ]);
+          $button['#options']['attributes']['data-ding-webtrekk-event'] = $event_data;
```
kasperg: Perhaps it is worth implementing a shared method for adding webtreck data to an element? It seems to occur multiple places throughout the code.

Example:

```php
function _ding_webtrekk_add_data(&$element, $link_id, $data) {
  $event_data = [
            'linkId' => $link_id,
            'customClickParameter' => $data,
  ];
  $element['#attributes']['class'][] = 'ding-webtrekk-event';
  $element'#attributes']['data-ding-webtrekk-event'] = json_encode($event_data);
}
```

modules/ding_webtrekk/ding_webtrekk.module
```diff
@@ -90,35 +153,170 @@ function ding_webtrekk_views_post_execute(&$view) {
...
+      'p_mat_lang' => $entity->getLanguage(),
+      'p_mat_source' => $entity->getAc_source(),
+      'p_mat_category' => implode(';', $entity->reply->getGenre()),
+    ];
+    ding_webtrekk_add_page_parameters($params);
+  }
+
+  // Look for at reserve button on this ding entity and setup a Webtrekk event
+  // if present. These could be added in every view_mode and page.
+  if (isset($entity->content['ding_entity_buttons'])) {
```
kasperg: Hmmm... a case for a `ding_entity_buttons_alter` 🤔 .

modules/ding_webtrekk/ding_webtrekk.module
```diff
@@ -90,35 +153,170 @@ function ding_webtrekk_views_post_execute(&$view) {
...
+    return;
+  }
+
+  // Ding loan stores the loans in the this form element and we take advantage
+  // of that to get the material ids for the renewable loans.
+  $loans = $form['items']['#value'];
+  $material_ids = [];
+  foreach ($loans as $loan) {
+    if ($loan->renewable) {
+      $material_ids[$loan->id] = $loan->getEntity()->getId();
```
kasperg: Do we really want/need to track every element renewed on the website? That seems a bit execessive 🤔 .

modules/ding_webtrekk/ding_webtrekk.module
```diff
@@ -90,35 +153,170 @@ function ding_webtrekk_views_post_execute(&$view) {
...
+/**
+ * Returns all the Webtrekk page parameteres collected during the request.
+ *
+ * @see ding_webtrekk_preprocess_html()
+ */
+function ding_webtrekk_get_page_parameters() {
+  $parameters = ding_webtrekk_add_page_parameters();
+  // Give other modules a last chance to alter the webtrekk page parameters
+  // before they are added to the current page load.
+  drupal_alter('ding_webtrekk_page_parameters', $parameters);
```
kasperg: :+1:

modules/ding_webtrekk/ding_webtrekk.module
```diff
@@ -131,5 +329,13 @@ function _ding_webtrekk_set_data_layer_values(array $value_map) {
   $script = implode(";\n", $script_lines);
   drupal_add_js($script, [
     'type' => 'inline',
+    'group' => JS_LIBRARY,
+    'weight' => -22,
```
kasperg: A comment about the weight here would be good.

modules/ding_webtrekk/ding_webtrekk.js
```diff
@@ -0,0 +1,131 @@
...
+ */
+
+(function($) {
+
+  "use strict";
+
+  var appendQueryParameter = function(link, key, value) {
+    var seperator = (link.indexOf('?') !== -1) ? '&' : '?';
+    var urlParameter = key + '=' + encodeURIComponent(value);
+    return link + seperator + urlParameter;
```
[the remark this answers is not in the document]
holt83: Agree, will chance this.

modules/ding_webtrekk/ding_webtrekk.js
```diff
@@ -0,0 +1,131 @@
...
+      // theme hook to avoid this. It doesn't provide the ability to add custom
+      // attributes/classes in preprocess and is not using using render arrays
+      // for rating elements, making them problematic to modify.
+      $('.ding-webtrekk-rating-event', context).once('ding-webtrekk', function() {
+        var contentId = $(this).data('ding-entity-rating-id');
+        $('.js-rating-symbol', this).each(function(index) {
+          var rating = (index + 1) + '';
+          $(this).click(function() {
+            wts.push(['send', 'click', {
+              linkId:'Materiale rating',
```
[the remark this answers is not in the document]
holt83: Well spotted :) Will fix this.

modules/ding_webtrekk/ding_webtrekk.js
```diff
@@ -0,0 +1,131 @@
...
+              customClickParameter: {
+                58: rating,
+                57: contentId
```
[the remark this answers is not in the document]
holt83: These are just some arbitrary numbers, that the business has decided to use for the different parameters. They are listed in the [spreadsheet](https://docs.google.com/spreadsheets/d/1OdUWGQc_hiY8F6BPViKRrBykSQfHLnhv4-YMdUzswSI/edit#gid=0) under the column WTK.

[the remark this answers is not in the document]
holt83: Perhaps they should be documented in code or maybe a link to the spreadsheet?

modules/ding_webtrekk/ding_webtrekk.js
```diff
@@ -0,0 +1,131 @@
...
+              }
+            }]);
+          });
+        })
+      });
+
+      // Track autocomplete selections.
+      $('.form-item-search-block-form .form-autocomplete', context)
```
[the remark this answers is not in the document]
holt83: Will do 👍

modules/ding_webtrekk/ding_webtrekk.js
```diff
@@ -0,0 +1,131 @@
...
+      //
+      // The primary reason for this is that we need the title of the carousel
+      // to pass as value in event and URL-parameter. For tabbed carousels the
+      // title of the individuals carousels in the tabs are set in the
+      // data-title attribute of the carousels. For single carousels the title
+      // is not avialable here, but instead it's controlled by the configuration
+      // on the panel pane. There are no panel hooks where we can get the title
+      // and easily change the content tree to add event and URL-parameters to
+      // elements (content is already rendered when hook is run).
+      $('.ding-carousel').each(function() {
```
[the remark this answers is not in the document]
holt83: I actually left out context and once here on purpose, since the items in the carousels can be lazy-loaded when the users reach the end of the carousel. I wanted to make sure the selectors applied when new items is added.

I must admit I havn't been able to test if this is necessary, since there was some problems with the lazy-loadings. (probably because of: https://platform.dandigbib.org/issues/3903).

I didn't see it as a big issue though, since all the operations inside the selector are once'd and using context.

I see now that maybe a comment about it would be appropriate?

modules/ding_webtrekk/ding_webtrekk.js
```diff
@@ -0,0 +1,131 @@
...
+        var carouselTitle = false;
+        var key = 'u_navigatedby';
+
+        // We need a title to send as value in the 'u_navigatedby' parameter, so
+        // if we can't find a carousel title we will just have to fallback to
+        // this generic one.
+        carouselTitle = 'unknown_carousel';
```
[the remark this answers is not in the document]
holt83: Yep, I missed that one. Will fix.

modules/ding_webtrekk/ding_webtrekk.js
holt83: I wasn't aware of the events to be honest. Which event do you suggest we use? After looking at the available events it seems to be more complicated to determine if it was prev or next that was clicked.

modules/ding_webtrekk/ding_webtrekk.module
holt83: Will do

modules/ding_webtrekk/ding_webtrekk.module
holt83: Good point. I will ask the business about this.

modules/ding_webtrekk/ding_webtrekk.module
holt83: You're right. It was one of the first things I made in this PR and actually and I see it doesn't make sense now. Will change it so that it also uses the values from the object.

modules/ding_webtrekk/ding_webtrekk.module
holt83: Good point. Will fix this.

modules/ding_webtrekk/ding_webtrekk.module
holt83: I would have thought so too, but the `TingEntity `class doesn't have the `getGenre()` method, so we have to through to the reply that implements `TingObjectInterface` to get the correct DKABM-field that we want to use for this parameter.

modules/ding_webtrekk/ding_webtrekk.module
holt83: Good idea. I will implement this.

modules/ding_webtrekk/ding_webtrekk.module
holt83: Maybe, I'll check it out..

modules/ding_webtrekk/ding_webtrekk.module
holt83: I agree, but this was the requirement.

modules/ding_webtrekk/ding_webtrekk.module
holt83: Agree, will add comment.

modules/ding_webtrekk/ding_webtrekk.js
```diff
@@ -0,0 +1,131 @@
...
+        wts.push(['send', 'click', eventData]);
+      });
+
+      // Secial handling for ding entity rating event.
+      //
+      // It would require a complete rework of the ding_entity_rating_display
+      // theme hook to avoid this. It doesn't provide the ability to add custom
+      // attributes/classes in preprocess and is not using using render arrays
+      // for rating elements, making them problematic to modify.
+      $('.ding-webtrekk-rating-event', context).once('ding-webtrekk', function() {
```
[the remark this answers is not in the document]
holt83: I'll add js-prefix to the classes!

holt83:
@kasperg I updated the PR with the suggested changes.

modules/ding_webtrekk/ding_webtrekk.js
holt83: I looked a bit further into this. There doesn't seem to be any events for the arrow-button and therefore the easiest approach seems to use custom click handlers as i done now in the PR.

holt83: There is a swipe event, which we could use to send event, when mobile users swipes the carousel or it's dragged. I believe the business hasn't thought about this when choosing what to track and I have asked them about it.

modules/ding_webtrekk/ding_webtrekk.module
holt83: I asked and there was no problem with this. They are working on a way to exclude administrative users anyway.

modules/ding_webtrekk/ding_webtrekk.js
holt83: I asked the business and there was no interest in tracking swipes/drags on carousel.

modules/ding_webtrekk/ding_webtrekk.module
```diff
@@ -282,6 +277,18 @@ function ding_webtrekk_form_ding_loan_loans_form_alter(&$form, &$form_state) {
   }
 }
 
+/**
+ * Use this function to attach webtrekk events to an elements attribute array.
+ */
+function ding_webtrekk_attach_event(&$attributes, $link_id, $data) {
```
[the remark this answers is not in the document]
holt83: I did consider passing entire elements as you suggested, but some elements have 'attributes' under '#options' and some elements has attributes under '#attributes'.

modules/ding_webtrekk/ding_webtrekk.js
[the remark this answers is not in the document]
holt83: I'll add some constants

modules/ding_webtrekk/ding_webtrekk.js
```diff
@@ -69,6 +69,14 @@
       $('.ding-webtrekk-event-renew-selected', context)
         .once('ding-webtrekk')
         .click(function(e) {
+          // The renew-all button is implemented by checking all renewable loans
+          // and then trigger a click on the renew-selected button. We can use
+          // this jQuere marker to ensure we don't send two events in this case.
+          // See: https://stackoverflow.com/a/10704256
+          if (e.isTrigger) {
+            return;
```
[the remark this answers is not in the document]
holt83: Agree. I also found some additional information about it:

https://github.com/jquery/api.jquery.com/issues/319
http://newcodeandroll.blogspot.com/2012/01/how-to-detect-if-event-is-triggered.html

The last one above also follows your suggestion. I'll update the code to use this approach

modules/ding_webtrekk/ding_webtrekk.module
holt83: Just wanted to put it as high as possible in the head and [jQuery seems to have the lowest weight of system libraries](https://api.drupal.org/api/drupal/modules%21system%21system.module/function/system_library/7.x)

I can see the comment is misleading. I'll change it.

holt83:
PR is updated!

modules/ding_webtrekk/ding_webtrekk.module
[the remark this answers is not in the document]
kasperg: :+1:

PR #1285, opened 2018-12-04
3603: Patch localization client to support newer jQuery versions

kasperg:
#### Link to issue

#### Description

The latest version of the localization client module is ~5 years old
and has problems with newer versions of jQuery because $.browser has
been removed.

We patch the module to fix the problem. The chosen patch has also been
committed by the module maintainer.

#### Screenshot of the result

Faulty site:

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1286, opened 2018-12-04
3603: Downgrade jQuery to 1.7 for the administration interface/theme

kasperg:
#### Link to issue

#### Description

The latest version of jQuery causes problems for the Module Filter
module: https://www.drupal.org/project/module_filter/issues/2986511

There is no immediate fix for this and the problem which originally
caused the upgrade is only relevant for patron actions performed in the 
DDBasic theme.

Consequently we downgrade the version of jQuery used to 1.7 (in practice
1.7.2) which is the same version currently used.

We need to ensure that the correct version is set on current and new
sites so we implement both hook_update and update hook_install to
make this change.

#### Screenshot of the result

Current version for new sites is actually 1.4 because the previous version only included a hook_update:
jquery update 2018-12-04 11-43-30

When updating the site to 3.1.1 for the admin theme the module filter breaks:
moduler 2018-12-04 11-44-09

Downgrading to 1.7.2 fixes the module filer:
moduler 2018-12-04 11-44-50

#### Checklist

### DIFF
--- a/modules/ding_base/ding_base.install
+++ b/modules/ding_base/ding_base.install
@@ -74,6 +74,8 @@ function ding_base_install() {
   variable_set('date_format_long', 'l, j. F, Y H:i');
   variable_set('date_format_medium', 'D, d-m-Y H:i');
   variable_set('date_format_short', 'd-m-Y H:i');
+
+  _jquery_update_set_theme_version('seven', '1.7');
 }
 
 /**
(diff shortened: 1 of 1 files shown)


modules/ding_base/ding_base.install
```diff
@@ -192,3 +194,10 @@ function ding_base_update_7006() {
   // Use the default version by setting an falsy value.
   _jquery_update_set_theme_version('seven', '');
 }
+
+/**
+ * Use jQuery 1.7 for admin theme.
+ */
+function ding_base_update_7007() {
+  _jquery_update_set_theme_version('seven', '1.7');
```
JrgenGNielsen: _jquery_update_set_theme_version() is defined in jquery_update.install
it won't be available for ding_base.install, will it?

JrgenGNielsen:
reviewed and approved

PR #1287, opened 2018-12-04
3708: Fixed result count

rimi-itk:
#### Link to issue

https://platform.dandigbib.org/issues/3708

#### Description

Show actual result count. 

#### Checklist

- [ ] My complies with [our coding guidelines](../docs/code_guidelines.md).
- [ ] My code passes our static analysis suite. If not then I have added a comment explaining why this change should be exempt from the code standards and process.
- [ ] My code passes our continuous integration process. If not then I have added a comment explaining why this change should be exempt from the code standards and process.

PR #1288, opened 2018-12-04
3832: Added result count after search result

rimi-itk:
#### Link to issue

https://platform.dandigbib.org/issues/3832

#### Description

Adds display of result count after actual search result (cf. https://github.com/ding2/ding2/pull/1287)

#### Checklist

Note: Scrutinizer complains about issues with linebreaks in template files, but these issues are not related to the code in this pull request.

### DIFF
--- a/modules/ting_search/templates/ting-search-results-count.tpl.php
+++ b/modules/ting_search/templates/ting-search-results-count.tpl.php
@@ -1,0 +1,11 @@
+<?php
+/**
+ * @file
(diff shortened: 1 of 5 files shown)


rimi-itk:
@kasperg: I've completed the checklist and added a note on the issues reported by Scrutinizer.

I'm not sure whether Scrutinizer checks template files incorrectly or if our use of e.g. `<?php endif ?>` in template files are incorrect.

PR #1293, opened 2018-12-10
3936: Fix missing titles on loan and resevation list

holt83:
#### Link to issue

This PR is dependant on: https://github.com/ding2/ting-client/pull/27

https://platform.dandigbib.org/issues/3936

#### Description

In https://github.com/ding2/ding2/pull/1147 we switched from using a rec.id search to using a getObject request, when fetching multiple objects from opensearch. This has led to a change in the way the system is handling unknown/missing/inaccessible objects.

When using a rec.id search to fetch objects, opensearch will just return an empty search result set if nothing was found. But getObject behaves differently and will return a "fake" search result record:

```
<searchResult>
  <collection>
    <resultPosition>1</resultPosition>
    <numberOfObjects>1</numberOfObjects>
      <object>
        <dkabm:record>
          <dc:title>
            Error: unknown/missing/inaccessible record: 456456456
          </dc:title>
        </dkabm:record>
        <identifier>456456456</identifier>
        <primaryObjectIdentifier/>
        <formatsAvailable>
          <format>dkabm</format>
        </formatsAvailable>
      </object>
  </collection>
</searchResult>
```

As a consequence the system will try to display these objects. For example on the loan/reservation list for non-searchable ILL-loans it will display a material with the title "Error: unknown/missing/inaccessible record: ...".

It's also possible to show fake material views by just choosing some random namespace/faust. Example:

https://vanilla-fbs.ddbcms.dk/ting/object/870970-basis:123123123123

Instead of passing these fake objects to Drupal, it's better to handle in the opensearch client, so that these objects are discarded early. That is what is done in https://github.com/ding2/ting-client/pull/27.

This PR mainly just removes some unnecessary checks and updates some code/comment to handle array from TingClientObjectRequest.

#### Checklist

### DIFF
--- a/modules/ding_loan/plugins/content_types/loans.inc
+++ b/modules/ding_loan/plugins/content_types/loans.inc
@@ -176,7 +176,7 @@ function ding_loan_loans_form($form, &$form_state, $account, $loans = array()) {
       // about the associated material.
       /** @var \TingEntity $entity */
       $entity = $loan->entity;
-      if (!is_object($entity) || NULL === $entity->getTingObject()->data) {
+      if (!is_object($entity)) {
         // Patron have this item in loans list, but the search provider can't
         // find this record for current agencyId, so we load it separately.
         $ting_object = ting_search_object_load($loan->ding_entity_id);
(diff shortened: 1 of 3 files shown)


JrgenGNielsen:
reviewed and approved

guddo:
Is this solved in https://github.com/ding2/ding2/pull/1384 and should be closed?

holt83:
Yes, this PR is superseded by #1384. Will close now.

PR #1294, opened 2018-12-11
3638 by Inlead: Require image share permission prior to BPI pushing.

AnatolyMuntean:
#### Link to issue

https://platform.dandigbib.org/issues/3638

#### Description

On top of the original 3638 requirement, is to explicitly mark the field required and only validate it's value prior to BPI push.

#### Screenshot of the result

image

#### Checklist

_Scrutinizer issue is unrelated._
- [x] My code passes continuous integration process. If not then I have added a comment explaining why this change should be exempt from the code standards and process.

### DIFF
--- a/modules/bpi/bpi.push.inc
+++ b/modules/bpi/bpi.push.inc
@@ -63,7 +63,7 @@ function bpi_http_push_action_form($nid) {
     ),
     '#description' => t('You should have permission to publish the images before selecting this option.'),
     '#default_value' => '_none',
-    '#element_validate' => array('bpi_push_with_images_validate'),
+    '#required' => TRUE,
   );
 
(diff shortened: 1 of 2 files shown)


modules/bpi/bpi.push.inc
```diff
@@ -63,7 +63,7 @@ function bpi_http_push_action_form($nid) {
     ),
     '#description' => t('You should have permission to publish the images before selecting this option.'),
     '#default_value' => '_none',
-    '#element_validate' => array('bpi_push_with_images_validate'),
+    '#required' => TRUE,
```
kasperg: It seems as if this should be handled by setting ` '#required' => TRUE` and then defining `#empty_option` for the element? Then we could drop the `_none` values and the custom validation?.

PR #1295, opened 2018-12-12
3603: Update definition of js.cookie library

kasperg:
#### Link to issue

#### Description

As a part of #1243 we now download js.cookie 2.20 for jQuery 3 compatibility.

We have to update the library definition accordingly for the library to
get loaded properly.

#### Checklist

The Scrutinizer error is irrelevant to the change at hand and should be disregarded.


JrgenGNielsen:
reviewed and approved

PR #1296, opened 2018-12-12
3985: Remove .slick-slide placeholder

BirkAndMe:
#### Link to issue

https://platform.dandigbib.org/issues/3985

#### Description

Instead of removing the .placeholder in the slick slide, it now makes sure to remove the entire .sick-slide, if the placeholder is nested inside (it also removes the .slick-slide if the placeholder is not nested).

#### Screenshot of the result

screenshot_2018-12-12 offentlige lister pbj ding virt b14cms dk

#### Checklist


BirkAndMe:
The scrutinizer bug seems to be related to an older issue.

PR #1297, opened 2018-12-12
3991 by Inlead: Do not let P2b overtake buy-button.

inlead:
#### Link to issue

https://platform.dandigbib.org/issues/3991

#### Description

When the P2b module is active, it assumes that all ticket-links are given by the P2b module, disallowing the possibility of manual inputting a link to tickets.

#### Checklist

### DIFF
--- a/modules/ding_place2book/ding_place2book.module
+++ b/modules/ding_place2book/ding_place2book.module
@@ -311,17 +311,15 @@ function ding_place2book_process_node(&$variables) {
   if ($variables['type'] == 'ding_event' && $variables['view_mode'] == 'full') {
     $p2b_settings = ding_place2book_settings($variables['nid']);
     $p2b_id = $p2b_settings['place2book_id'];
-    if (empty($p2b_id) || $p2b_id == DING_PLACE2BOOK_NO_ID) {
-      unset($variables['book_button']);
-      return;
-    }
 
-    $url = !empty($variables['field_ding_event_ticket_link'][0]['url']) ? $variables['field_ding_event_ticket_link'][0]['url'] : '';
-    $variables['book_button'] = theme('place2book_ticketsinfo', [
-      'place2book_id' => $p2b_settings['place2book_id'],
-      'url' => $url,
-      'type' => $p2b_settings['ticket_status'],
-    ]);
+    if (!empty($p2b_id) && $p2b_id != DING_PLACE2BOOK_NO_ID) {
+      $url = !empty($variables['field_ding_event_ticket_link'][0]['url']) ? $variables['field_ding_event_ticket_link'][0]['url'] : '';
+      $variables['book_button'] = theme('place2book_ticketsinfo', [
+        'place2book_id' => $p2b_settings['place2book_id'],
+        'url' => $url,
+        'type' => $p2b_settings['ticket_status'],
+      ]);
+    }
   }
 }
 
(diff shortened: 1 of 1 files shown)


modules/ding_place2book/ding_place2book.module
```diff
@@ -501,7 +500,7 @@ function ding_place2book_event_op($op, $node, $place2book_id = NULL) {
 
       // Catch insert error by checking the event id.
       if (empty($req_result->headers['event-id'])) {
-        $p2b_error = $req_result->headers['status'] . ' - ' . $req_result->status_message;
```
kasperg: These changes regarding error messages does not seem to be related to the change at hand?

Also: I could not find any documentation of these `error-message` and `error-class` headers anywhere. If we want to use them is there and url we could refer to or the like? A inline comment explaining the situation could also be helpful.

modules/ding_place2book/ding_place2book.module
```diff
@@ -553,7 +552,7 @@ function ding_place2book_event_op($op, $node, $place2book_id = NULL) {
       $req_result = drupal_http_request($service_settings['service_url'] . '/delete_event', $options);
 
       if ($req_result->status_message != 'Accepted') {
-        $p2b_error = $req_result->headers['status'];
+        $p2b_error = $req_result->headers['error-class'];
```
kasperg: See comment above. Also I find it odd that we refer to `error-class` here but `error-message` above.

The context here seems to indicate that we would want an error message.

modules/ding_place2book/ding_place2book.module
```diff
@@ -311,17 +311,16 @@ function ding_place2book_process_node(&$variables) {
...
-      return;
+    $passive = $p2b_settings['passive'];
+
+    if (!empty($p2b_id) && $p2b_id != DING_PLACE2BOOK_NO_ID && !$passive) {
+      $url = !empty($variables['field_ding_event_ticket_link'][0]['url']) ? $variables['field_ding_event_ticket_link'][0]['url'] : '';
+      $variables['book_button'] = theme('place2book_ticketsinfo', [
+        'place2book_id' => $p2b_settings['place2book_id'],
+        'url' => $url,
+        'type' => $p2b_settings['ticket_status'],
+      ]);
```
kasperg: :+1:

modules/ding_place2book/ding_place2book.module
AnatolyMuntean: This was a finding during run-time/debugging. The response headers simply lack the `status` key on those lines and delivered notice errors.
From the implemented logic I assume it tries to get an error string explaining what happened service-wise, so the best value resided in the `error-class` key.

kasperg: The two situations seem very similar. Is there a reason why we would use `error-message` combined with `$req_result->status_message` on L503 and `error-class` (without `$req_result->status_message`) here?

Without any further information I would expect we would use the same key.

AnatolyMuntean: Good question, since I didn't compare whether the service delivers consistent responses for both actions - `/create_event` and `/delete_event`. In other words, whether these - `status_message`, `error-message` and `error-class` - are present and under what circumstance.

modules/ding_place2book/ding_place2book.module
AnatolyMuntean: Likewise there was no documentation for the `status`. Either way this gotta be consulted with place2book api. The scope was to fix the notices.

PR #1298, opened 2018-12-13
3603: Prevent access to changing jQuery version

kasperg:
#### Link to issue

#### Description

The version of jQuery that we use can have significant impact on the
functionality of the site and break other JavaScript elements.
Consequently normal users should not be able to change this.

To address this we patch jQuery Update to add a new permission which is
used to determine access to functionality provided by jQuery Update.

The new permission is intentionally not assigned to any roles. This
prevents access to everyone except user 1.

#### Checklist

#### Comments

I do not know why Scrutinizer fails for this PR. It is not related to the change at hand.


JrgenGNielsen:
reviewed and approved

PR #1300, opened 2018-12-14
3910: Remove padding and change font-style

KristianB14:
#### Link to issue

https://platform.dandigbib.org/issues/3910#change-39903

#### Description

Padding-left removed from ul in ding-carousel. Font-style changed for ting-object with no overlay

#### Screenshot of the result

image

#### Checklist

No further comments or questions.


cableman:
Please remember to comment on why the Scrutinizer checker fails?

PR #1308, opened 2018-12-18
3908: Fix DB error on searches without keywords

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/3908

#### Description

When there's no search string the array to match against becomes empty, since FALSE and empty values are filtered out. db_select condition throws a [syntax error exception](https://platform.dandigbib.org/issues/3908#note-17) when using an empty array with 'IN'. The query doesn't make much sense anyway, since it will always return 0 rows. 
More info: https://www.drupal.org/project/drupal/issues/1976666 

So instead we just skip the select-query when the search string is empty and return empty array.

Alternatively, we should not filter out empty strings in array_filter, so that it would be possible to make campaign triggers that matches against empty strings, but I didn't think this was the right way to go. However, my knowledge about this module is limited, so I may be in the wrong here?

#### Checklist


JrgenGNielsen:
reviewed and approved
the php_lint error is not related to this PR

PR #1312, opened 2018-12-21
4039: fix search overlay spinner

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/4039

#### Description

The search overlay spinner suddenly started appearing when clicking the reserve-button. Investigating further I found these click-handlers from ting.js to be the culprit:

https://github.com/ding2/ding2/blob/7.x-4.6.0-rc5/modules/ting/js/ting.js#L18 
https://github.com/ding2/ding2/blob/7.x-4.6.0-rc5/modules/ting/js/ting.js#L25

These were not a problem before, since they were implemented with the jQuery `.live()` method, which for some reason wasn't fired (probably because of some code stopping event propagation). However, with the jQuery update, these were changed to `.on()` making them work and resulting in the search overlay incorrectly popping up when clicking on ding entity buttons (for example).

We only want these to pop up when users are clicking on actual search links, so I have made the selector more specific and removed some unused stuff (comments about this in commits).

I also discovered that the cancel button on the search overlay had stopped working for the same reason: the live-event was changed to an on-event. But in this case we need event delegation, since the overlay HTML is added dynamically. 

The `window.stop()` method, which we use to cancel the request when cancelling the overlay, apparently doesn't work in Chrome, but only on IE and Firefox (of the browsers I tested). Since a lot of users is on Chrome I tried to fix this and ended up using the `window.location.reload()` instead. It has the disadvantage that it will start a new page load of the current page instead of just stopping immediately. But at least it will work in all browsers, and these pages will also often be cached, so I don't think it's a big issue.

#### Checklist

### DIFF
--- a/modules/ting/js/ting.js
+++ b/modules/ting/js/ting.js
@@ -6,14 +6,6 @@
   "use strict";
 
   $(document).ready(function() {
-    $('a.js-search-overlay').on('click', function() {
-      var link = $(this);
-      if (link.attr('href').charAt(0) !== '#') {
-        // Only show overlay for non-local links.
-        Drupal.TingSearchOverlay();
-      }
-    });
-
     // Ensure overlay on collection view links.
     $('.ting-collection-wrapper a[href*="/ting/"]').on('click', function() {
       if ($(this).not('[target="_blank"]').length) {
--- a/modules/ting/js/ting.js
+++ b/modules/ting/js/ting.js
@@ -6,15 +6,8 @@
   "use strict";
 
   $(document).ready(function() {
-    // Ensure overlay on collection view links.
-    $('.ting-collection-wrapper a[href*="/ting/"]').on('click', function() {
-      if ($(this).not('[target="_blank"]').length) {
-        Drupal.TingSearchOverlay();
-      }
-    });
-
-    // Ensure overlay on object view links.
-    $('.ting-object-wrapper a[href*="/ting/"]').on('click', function() {
+    // Ensure search overlay on search links in a ting object view.
+    $('.ting-object a[href^="/search/ting/"]').on('click', function() {
       if ($(this).not('[target="_blank"]').length) {
         Drupal.TingSearchOverlay();
       }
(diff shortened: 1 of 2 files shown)


modules/ting/js/ting.js
```diff
@@ -6,23 +6,8 @@
...
-    $('.ting-collection-wrapper a[href*="/ting/"]').on('click', function() {
-      if ($(this).not('[target="_blank"]').length) {
-        Drupal.TingSearchOverlay();
-      }
-    });
-
-    // Ensure overlay on object view links.
-    $('.ting-object-wrapper a[href*="/ting/"]').on('click', function() {
+    // Ensure search overlay on search links in a ting object view.
+    $('.ting-object a[href^="/search/ting/"]').on('click', function() {
```
kasperg: Before the migration to jQuery 3.x [this was also using `live()`](https://github.com/ding2/ding2/blob/7.x-4.5.0/modules/ting/js/ting.js#L18). I am not sure if it is needed. though. I cannot come up with any situations where we load content with links inside the `.ting-object` selector so perhaps this is fine.

kasperg:
@holt83: Please look into the GitCop warning. While you are at it please also reply to my comment 😄 .

GitCop:

 * Commit: ccd9ceea0c6dc38876fce910a9654f85f21e6ae0

---

GitCop:

 * Commit: ccd9ceea0c6dc38876fce910a9654f85f21e6ae0

---

GitCop:

 * Commit: 4068b5a08f400768249c77e20e1cc479677d7846

---

GitCop:

 * Commit: 4068b5a08f400768249c77e20e1cc479677d7846

---

modules/ting/js/ting.js
holt83: I can't come up with anything related to search links either. So I think it's fine. It could easily be changed from direct to delegated though.

holt83:
@kasperg I updated the PR and fixed the GitCop warning (by replacing long URL with commit sha).

PR #1313, opened 2019-01-02
3983 search page reorg

xendk:
#### Link to issue

https://platform.dandigbib.org/issues/3983

#### Description

Reorders panes on search page. Also removes a deleted pane

#### Screenshot of the result

image
...
image


JrgenGNielsen:
reviewed and approved

PR #1316, opened 2019-01-04
3739: Availability optimization

holt83:
This PR is dependant on https://github.com/ding2/ting-client/pull/26

#### Link to issue

https://platform.dandigbib.org/issues/3739

#### Description

Adds support for fetching multiple marc-objects from opensearch at once and use this when generating shelfmark for holdings in FBS. The translation of faust to pid is also moved out of the loop, to also translate them all at once.

#### Checklist

The process could in theory be optimized even further by avoiding the translation of faust completely. I do not think this is worth pursuing, since in all cases the holdings are fetched in a subsequent AJAX-request and doesn't prolong the initial request. Also, the translation is also done in a single getObject request now, which is very fast.

### DIFF
--- a/modules/fbs/includes/fbs.availability.inc
+++ b/modules/fbs/includes/fbs.availability.inc
@@ -62,7 +62,11 @@ function fbs_availability_holdings($provider_ids) {
   $entities = [];
   $suffix_type = variable_get('fbs_holdings_suffix_type', 'shelf_mark');
   if ($suffix_type == 'simple') {
-    $entities = ding_entity_load_multiple($datawell_pids);
+    // Here the entities are keyed by local entity id, so build an array keyed
+    // by datawell pids for easier access.
+    foreach (ding_entity_load_multiple($datawell_pids) as $entity) {
+      $entities[$entity->ding_entity_id] = $entity;
+    }
   }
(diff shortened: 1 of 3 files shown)


cableman:
@holt83 please try re-running the CI tests

cableman:
The code looks good. Just need to pass the tests.

holt83:
I tried the rerun, but I still get this:

```
Parse error: ./modules/ding_sections/tests/lib/DINGTestPageAbstraction.class.inc:51
    49|    */
    50|   public function waitForElement($selector, $time = self::WAIT_FOR_ELEMENT, $force = TRUE) {
  > 51|     for ($second = 0; $second++) {
    52|       if ($second >= $time) {
    53|         if ($force) {
Unexpected ')', expecting ';' in ./modules/ding_sections/tests/lib/DINGTestPageAbstraction.class.inc on line 51
Exited with code 1
```

holt83:
Ah nevermind. I was missing this [commit](https://github.com/ding2/ding2/commit/2cbff91f36d8cf5379a97d08c54ce02b5eddaffa), so I just did a rebase, which hopefully will fix the tests

cableman:
Reviewed and awaiting release

PR #1317, opened 2019-01-07
1744: Set default carousels on frontpage

xendk:
#### Link to issue

https://platform.dandigbib.org/issues/1744

#### Description

The default export of the frontpage didn't have any carousels defined.


DDBdrift:
https://platform.dandigbib.org/issues/1744 | Resolved (tag version)

PR #1319, opened 2019-01-08
4049 by Inlead: Fix misspelling.

inlead:
#### Link to issue

https://platform.dandigbib.org/issues/4049

#### Description

Within the sourcecode there is misspellings of "No mails sent", which is given as "No mails send". This string is presented/displayed to the user.

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1323, opened 2019-01-11
3901: Webtrekk MWA fixes

holt83:
#### Link to issue

#### Description

Some small changes was requested by the business:

1. The Page parameter p_mat_category was incorrectly using a dc:subject DKABM field for the value. Instead it should only show whether the material is fiction or nonfiction (or none at all).

The only reliable way to check this, is by using the 'sk' marker in the DK5-classification. This value gets stripped in OpenSearchTingObject::getClassification(), so it's going to be a bit "hacky" to get to this. Either go directly to the record or look for empty string/FALSE. I choose the latter approach in this PR.

Alternatively, we should add a new getCategory() method to the TingObjectInterface, which purpose is to return fiction/nonfiction or FALSE if not applicable.

2. We have chosen to also track the size of current search-page in the parameter p_s_Size.

#### Checklist

- [ ] My code passes our continuous integration process. If not then I have added a comment explaining why this change should be exempt from the code standards and process

### DIFF
--- a/modules/ding_webtrekk/ding_webtrekk.module
+++ b/modules/ding_webtrekk/ding_webtrekk.module
@@ -117,6 +117,9 @@ function ding_webtrekk_page_alter(&$page) {
       // Track the page of the search result.
       $params['p_s_Page'] = $search_result->getSearchRequest()->getPage();
 
+      // Trak the size of the search result.
+      $params['p_s_Size'] = $search_result->getNumCollections();
+
       // Track the sort used in the search request.
       if (!empty(ting_search_current_results()->getSearchRequest()->getSorts())) {
         $current_sort = $search_result->getSearchRequest()->getSorts()[0];
--- a/modules/ding_webtrekk/ding_webtrekk.module
+++ b/modules/ding_webtrekk/ding_webtrekk.module
@@ -117,7 +117,9 @@ function ding_webtrekk_page_alter(&$page) {
       // Track the page of the search result.
       $params['p_s_Page'] = $search_result->getSearchRequest()->getPage();
 
-      // Trak the size of the search result.
+      // Track the size of the search result. We already track the total size
+      // of the search result in OSSr parameter, but this is the number of
+      // results on the current search page.
       $params['p_s_Size'] = $search_result->getNumCollections();
 
       // Track the sort used in the search request.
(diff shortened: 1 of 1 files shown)


modules/ding_webtrekk/ding_webtrekk.module
```diff
@@ -117,6 +117,9 @@ function ding_webtrekk_page_alter(&$page) {
       // Track the page of the search result.
       $params['p_s_Page'] = $search_result->getSearchRequest()->getPage();
 
+      // Trak the size of the search result.
+      $params['p_s_Size'] = $search_result->getNumCollections();
```
kasperg: [`getNumCollections()` is limited by paging](https://github.com/ding2/ding2/blob/96b6c0422177dd5c6987a963ce84cad26e8af262/modules/ting/src/Search/TingSearchResultInterface.php#L15-L33). It seems as if the right way to go would be to use `getNumTotalObjects()`? 

NB: I see that the pull request comment states that we want to return the size of the current search page and we already track the size of the search result in `OSSr`. If that is the case then please update the comment accordingly.

modules/ding_webtrekk/ding_webtrekk.module
```diff
@@ -189,12 +192,30 @@ function ding_webtrekk_ding_entity_view($entity, $view_mode) {
...
+    // If classification is 'sk' OpenSearchTingObject returns empty string.
+    // See: OpenSearchTingObject::getClassification().
+    if ($classification === '') {
+      $category = 'fiktion';
+    }
+    // Else, if not FALSE, the material has a DK5 classification and is
+    // nonfiction.
+    elseif ($classification !== FALSE) {
+      $category = 'nonfiktion';
+    }
```
kasperg: This is really hacky in many ways. From a very vague description of `getClassification()` to returning an empty string instead of `FALSE` for fiction etc.

I am on the fence regarding whether it makes sense to implement a new method but what might introduce problems of its own. Let's just go with this.

holt83: Agreed, it's hacky. I also thought about letting `getClassification()` just return the 'sk' marker, but I guess then other code will have to check for this to avoid showing it in frontend.

modules/ding_webtrekk/ding_webtrekk.module
holt83: I updated the comment.

modules/ding_webtrekk/ding_webtrekk.module
holt83: @kasperg I have learned that it's more complicated to determine genre category than by just looking for the 'sk' marker. Some foreign materials can be non-fiction even though they have a classification number and not 'sk'.

I really think we need a method to determine the genre category. I realise that we can't just update the interface, but how about we add it to `OpenSearchTingObject` and make this openseach specific for now?

PR #1326, opened 2019-01-16
3982: Use Views lite pager on admin node view

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/3982

#### Description

The bpi and ding_base module is adding additional filters to the admin_views_node view resulting in the associated queries getting more complex and slow. Some of the larger libraries has even experienced timeouts.

The main problem appears to be the count-query performed by the pager as reported in the issue. Appearently these queries have scalability problems on innodb databases. I found a module which tries to solve this problem:

https://www.drupal.org/project/views_litepager

With the changes in this PR the pager provided by this module is used on admin_views_node.

The main disadvantage is that you can only navigate one page at a time, where before you could jump five pages a time. The description in the module also mentions that you can't navigate to the last page, but I don't think this is an issue, since you can just sort by asc/desc depending on which sort is currently used.

#### Screenshot of the result

skaermbillede fra 2019-01-16 23 04 36
#### Checklist

Alternatively, we could look at the bpi_status filter added in `bpi_features_views_default_views_alter()`. In my tests, I found this filter to be the main culprit in the performance issues. After removing this, the performance was consistently much better  (from 200-300ms to  <  10ms). The filter could be removed or we could try to improve the perfomance. 

So if the business decides that the tradeoffs are not worth it, the next step would be to improve the perfomance of this filter or remove it. I also noticed that the workflow module, which supplies this filter, is a bit outdated.

### DIFF
--- a/ding2.make
+++ b/ding2.make
@@ -431,6 +431,9 @@ projects[views][version] = "3.20"
 projects[views_bulk_operations][subdir] = "contrib"
 projects[views_bulk_operations][version] = "3.3"
 
+projects[views_litepager][subdir] = "contrib"
+projects[views_litepager][version] = "3.0"
+
 projects[views_responsive_grid][subdir] = "contrib"
 projects[views_responsive_grid][version] = "1.3"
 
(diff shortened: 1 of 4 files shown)


kasperg:
Closed in favour of #1331.

PR #1328, opened 2019-01-22
1353: Use hash on carousel id when generating lock and cache id

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/1353

#### Description

TingSearchCarouselQuery is converted to a string ID by taking the configuration, converting it to JSON and base64 encoding it.

This string ID is then used as suffix for the lock and cache ids, when fetching items for the carousel. The problem is that these IDs can get very long preventing the module from obtaining a lock, since lock ids are restricted to 255 characters in the DB. As a consequence the search was never actually performed and the carousel just kept requesting new items because it always returned offset 0. (See screenshot).

Another effect was that the carousel items was never cached, but the problem with lock id is much more serious since it prevents the carousel from functioning at all.

I propose that we instead use a hash of the TingSearchCarouselQuery string id as suffix for the lock and cache id, to prevent it form getting to long,

#### Screenshot of the result

image

#### Checklist

### DIFF
--- a/modules/ting_search_carousel/ting_search_carousel.module
+++ b/modules/ting_search_carousel/ting_search_carousel.module
@@ -79,7 +79,10 @@ function ting_search_carousel_ctools_plugin_directory($module, $plugin) {
  *   'next_offset', which is the next offset to use or -1.
  */
 function ting_search_carousel_get_entities(TingSearchCarouselQuery $query, $start, $count, $query_only = FALSE) {
-  $query_id = (string) $query;
+  // Extract an id representing the carousel query to use in lock and cache id.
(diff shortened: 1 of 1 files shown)


cableman:
Hej Simon

Jeg skriver lige til dig denne vej, da jeg ved at jeg har an-givet det samme problem til Thomas fra reload i en anden ticket (kan bare ikke find den ticket).

Så er jeg ikke sikker på at jeg forstår hvad relationen er til https://platform.dandigbib.org/issues/3922 (Fejl i visningen af nyheder)?

Mvh
  Jesper K

cableman:
Reviewed and approved :-D

holt83:
Thanks! Sorry about the confusing wrong link hehe :)

PR #1333, opened 2019-01-25
4050: Ensure ding_webtrekk is activated on existing installs

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/4050

#### Description

Apparently the business also wants to ensure that ding_webtrekk is activated on existing installs.

#### Checklist


holt83:
All the Scrutinizer errors seems to come from the `ding_place2book.module` file and this PR only adds an update function to `ding2.install`

JrgenGNielsen:
reviewed and approved

PR #1335, opened 2019-01-28
4080: Add empty alt tag on logo, for WCAG.

rasben:
#### Link to issue

https://platform.dandigbib.org/issues/4080

#### Description
To help screenreaders, an empty alt tag is required.

#### Checklist


rasben:
The scrutinizer error is unrelated:
It mentions issues in "modules/ding_place2book/ding_place2book.module", but I've done nothing related to that file

JrgenGNielsen:
reviewed and approved

PR #1338, opened 2019-01-28
4083: Extended search should have a screenreader label for select.

rasben:
#### Link to issue

https://platform.dandigbib.org/issues/4083

#### Description
For WCAG reasons, there should be a label for the `<select>` element in the extended search form. It should be visible only to screenreaders.

#### Checklist


rasben:
The scrutinizer error is unrelated:
It mentions issues in "modules/ding_place2book/ding_place2book.module", but I've done nothing related to that file

JrgenGNielsen:
reviewed and approved

PR #1339, opened 2019-01-28
4116: Add clearfix so wrapper doesn't collapse

xendk:
#### Link to issue

https://platform.dandigbib.org/issues/4116

#### Description

Add clearfix so wrapper doesn't collapse.

Not using so many 100% width and left floated divs would be better,
but no time for that.

#### Checklist


xendk:
Huh, thought I did that, must have slipped

PR #1340, opened 2019-01-29
4121: Make sure that language suffix is displayed.

rasben:
#### Link to issue

https://platform.dandigbib.org/issues/4121

#### Description

The language suffix was never displayed, as `$entity->getType()` returned the type with an uppercase, whilst the `variable_get('ting_language_type_title_suffix', []);` array had lower case elements.

#### Screenshot of the result

It adds the "(Norsk)":

screenshot 2019-01-29 at 11 43 41

#### Checklist


rasben:
The scrutinizer error has nothing to do with my PR - I have made no changes in the file it complains about.

JrgenGNielsen:
reviewed and approved

PR #1342, opened 2019-01-29
3860 by Inlead: Make original and copy messages the same.

bozeak:
#### Link to issue

https://platform.dandigbib.org/issues/3860

#### Description

Original message and copy message sent from contact form had different headers. This commit fixes this issue so now the original and copy messages have the same headers.

#### Checklist


inlead:

PR #1343, opened 2019-01-29
3638 by Inlead: Validate image dropdown only on BPI push.

inlead:
#### Link to issue

https://platform.dandigbib.org/issues/3638

#### Description

Removed the visible requirement of a field, and moved it to error message upon push to BPI is clicked.

#### Screenshot of the result

pasted_image_at_2019-01-29__4_57_pm

#### Checklist


inlead:

PR #1344, opened 2019-01-30
4115 by Inlead: Get carousels info from panes.

inlead:
#### Link to issue

https://platform.dandigbib.org/issues/4115

#### Description

Carousel panes info is missing in ding_app_variables output.

#### Checklist


inlead:

holt83:
Doing it this way:

1. It will load all carousel panel panes in the system. Not just those on the frontpage. Is this intentionally?
2. If the carousel panel panes are defined in code in an exported panel page that is not overriden, the configurations will not be present in the database.

It's possible to load only the carousels defined on the frontpage.
```
$ding_frontpage = page_manager_get_page_cache('page-ding_frontpage');
$handler = reset($ding_frontpage->handlers);
$display = panels_panel_context_get_display($handler);
 // The panes are stored in a property called 'content'.
foreach ($display->content as &$pane) {
  // Add country code as a value to vary by.
  if ($pane->type == 'carousel' && $pane->subtype == 'carousel) {
    // Get the config
  }
}
```

It's also possible to load all carousel panel panes regardless of them being in db/code.

guddo:
1. Yes.
2. I guess this is expected? What's the example of having carousels defined in code?

holt83:
Well, isn't it a bit much to just return all carousels? Where are all these carousels shown in the app?

guddo:
1. I don't think so. It's carousel that is manually added to frontpage and sections.
2. You wan't to manage configurations like CQL-query in code?

holt83:
1. But what if you have several section with several carousels? Wont it be difficult for libraries to control which carousels are shown in app?

2. With the new ability to filter only available materiales I think there's a use case for this, where you have   a more general cql like for example: "facet.fictionSubject=krimi and holdingsitem.accessionDate>=NOW-60DAYS"

guddo:
1. I don't know how and what is controlled nor displayed in the app, I just know what is exposed in this module.
2. But this is not search carousels which you are able to insert on a panel..

holt83:
> 2. But this is not search carousels which you are able to insert on a panel..

Yes, you are able to insert these on a panel.

guddo:
> Yes, you are able to insert these on a panel.

So the scenario you set is: someone develops a module, which uses the carousel functionality and the config array. This module has the configuration exported into code and the module exposes the carousel as a pane, which the editor can then insert on a panel. Is that correctly understood?
If so; how realistic is this scenario?

holt83:
> So the scenario you set is: someone develops a module, which uses the carousel functionality and the config array. This module has the configuration exported into code and the module exposes the carousel as a pane, which the editor can then insert on a panel. Is that correctly understood?

No, you can't export a panel config separately. But nevermind :)

I'm more concerned with the lag of control of which carousels are being exported and where they are being shown.

holt83:
Look, it really wasn't my intention to start a discussion about this :) 

I made some observations about the way that you choose to implement this. I show you some example code of how to more correctly work with the Panels/Ctools API. In the end I really don't care how you choose to do it. I can just patch it in our own code base.

It's a case of take it or leave it :)

modules/ding_app_variables/ding_app_variables.module
```diff
@@ -28,7 +28,14 @@ function ding_app_variables_menu() {
  *   List of all variables to the output.
  */
 function ding_app_variables_display() {
-  $carousel_configs = variable_get('ting_carousel_search_queries');
+  $carousels = array(
+    'carousel',
+  );
```
cableman: So we create an array with one element to loop over that one element in the next lines. Why not

```
$carousel_configs[] = _ding_app_variables_get_carousel_configs('carousel');
```

modules/ding_app_variables/ding_app_variables.module
```diff
@@ -50,3 +57,34 @@ function ding_app_variables_display() {
...
+ *
+ * @param string $carousel
+ *   Pass subtype of panel pane.
+ *
+ * @return array
+ *   List of carousels pane configurations.
+ */
+function _ding_app_variables_get_carousel_configs($carousel) {
+  $query = db_select('panels_pane', 'pp');
+  $result = $query
```
cableman: I think that it would be a bette solution to load the configuration through the panels "API" as suggested by @holt83 

This would ensure all use-cases and be a more robust solution.

@kasperg be free to comment if you disagree whit this.

PR #1345, opened 2019-01-30
3667: Enable dropdown function, for header search.

rasben:
#### Link to issue

https://platform.dandigbib.org/issues/3667

#### Description
In relation to a previous revision where the header
structure was moved around, the functionality of
being able to fold/unfold the search line disappeared.

This commit looks to fix that, and also tries to clean up
all the mess in regards to the various combinations that
the header can be displayed in.
#### Screenshot of the result

The biggest issue of these changes was making sure that everything worked across all the various combinations.
Taking screenshots of all of that would take forever, but essentially it should just look like the production sites - except the two elements should finally be switched in the header.

See here:
https://github.com/ding2/ding2/pull/1322
and
https://platform.dandigbib.org/issues/3667

#### Checklist

If you test this (and you are very welcome!!) then make sure to test it across all the different combinations - here are some:
- There are 3 different styles of the search bar, which can be set at admin/config/ting/search
- If you are on a search or frontpage, the bar will always be visible
- If you are on any other page, the extended search bars will be hidden with a fold out
- ... and then you need to test all of the above across responsiveness
- ... AND THEN you need to test it if it also works with a submenu
- ... AND THEN you need to test all of the above again, to see if it also works with sub-links in the main menu

### DIFF
--- a/modules/ting_field_search/ting_field_search.module
+++ b/modules/ting_field_search/ting_field_search.module
@@ -441,8 +441,8 @@ function ting_field_search_preprocess_html(&$vars) {
     }
 
     // If the profile has visibility all always expand the search field.
-    if (ting_field_search_is_profile_visible($profile)) {
-      $vars['classes_array'][] = 'extended-search-is-open';
+    if (!ting_field_search_is_profile_visible($profile)) {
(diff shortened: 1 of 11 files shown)


rasben:
The scrutinizer error is unrelated to my changes

PR #1351, opened 2019-02-04
4088: Screenreader / focus tweaks for userlogin.

rasben:
#### Link to issue

https://platform.dandigbib.org/issues/4088

#### Description

- Makes sure the hidden login form is also hidden for screenreaders when it's closed, using aria-hidden
- When the login form is hidden, remove ability to tabbing to it
- When it is open, set focus on first element

#### Checklist


rasben:
The Scrutinizer error is not relevant as the ddbasic is a global function which the JS functionality on the site uses.

JrgenGNielsen:
reviewed and approved

PR #1353, opened 2019-02-04
4089: Make sure linkboxes have hovereffect on focus.

rasben:
#### Link to issue

https://platform.dandigbib.org/issues/4089

#### Description

Currently, the linkboxes does not show the hover effect when you keyboard-tab to them.
This fixes that.

It's these boxes we're talking about:

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1356, opened 2019-02-06
4097: W3C Invalid syntax fixes

rasben:
#### Link to issue

https://platform.dandigbib.org/issues/4097

#### Description

These are minor fixes which makes the W3C HTML Validator happy

#### Checklist

### DIFF
--- a/modules/ding_nodelist/templates/ding_nodelist.ding_event.rolltab.tpl.php
+++ b/modules/ding_nodelist/templates/ding_nodelist.ding_event.rolltab.tpl.php
@@ -2,28 +2,7 @@
 
 /**
  * @file
- * Template file for taxonomy-like layout.
+ * Template file for ding_news rolltab layout.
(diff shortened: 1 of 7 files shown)


rasben:
Scrutinizer complains about my nesting in the template file, but I feel like implementing the requested changes would hurt readability. What is your opinion?

JrgenGNielsen:
you can try the alternative syntax for control structures:
<?php if ($foo): ?>
foo
<?php endif; ?>
- otherwise, when in doubt, stick to Drupal standard :-)

rasben:
Ok i've fixed it - but now scrutinizer complains about this which I dont think makes any sense:

screenshot 2019-02-15 at 11 06 16

So it wants me to do:

```php
<?php endif;
?>
```

which looks mad

JrgenGNielsen:
reviewed and approved

PR #1359, opened 2019-02-08
3951 by Inlead: Add more metadata to RSS feed.

bozeak:
#### Link to issue

https://platform.dandigbib.org/issues/3951

#### Description

This commit contain changes and features export which adds Event CT "Place/Room" and  "Organizer" fields to Ding App Content RSS feed.

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1361, opened 2019-02-08
4100: Entities in a carousel should have a <h3> title.

rasben:
#### Link to issue

https://platform.dandigbib.org/issues/4100

#### Description

For WCAG reasons, the elements within a carousel should use `<h3>` instead of `<h2>` for their titles.
We'll need to override a suffix/prefix element to achieve this.

#### Checklist


JrgenGNielsen:
reviewed and approved

xendk:
See Redmine issue for comment.

rasben:
I will close this PR as @xendk has a better solution @ https://github.com/ding2/ding2/pull/1426

PR #1364, opened 2019-02-12
4139: Minor changes to opensearch cache

cableman:
#### Link to issue

https://platform.dandigbib.org/issues/4139

#### Description

In working with ereolen.dk we had issues with ting object that contained content from other objects and cache misses after the first view (which should not happen as the first view populated the cache).

#### Screenshot of the result

No screenshot

#### Checklist

Linked to https://github.com/ding2/ting-client/pull/29

### DIFF
--- a/modules/opensearch/opensearch.client.inc
+++ b/modules/opensearch/opensearch.client.inc
@@ -58,7 +58,7 @@ function opensearch_get_object($object_id, $collection = FALSE, $with_relations 
     // Execute the request.
     $result = opensearch_execute($request);
 
-    return $collection || is_null($result) ? $result : reset($result);
+    return ($collection || is_null($result)) ? $result : $result[$object_id];
   }
 
   return NULL;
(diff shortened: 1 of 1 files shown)


JrgenGNielsen:
reviewed and approved

modules/opensearch/opensearch.client.inc
```diff
@@ -554,13 +565,18 @@ function _opensearch_predict_cache(array $collections, TingClientRequest $reques
         $object_request->setProfile($profile);
       }
 
+      // Build both requests for local id and PID.
       $object_id_request = clone $object_request;
       $object_id_request->setObjectId($object->id);
       $local_id_request = clone $object_request;
       $local_id_request->setLocalId($object->localId);
```
holt83: @cableman I know this is not from this PR, but isn't this a bit dangerous? The localId is not guaranteed to unique over all owners, so we are really not sure this object is the correct cache response for this localID

PR #1369, opened 2019-02-19
3916: Refactor event list filter to be independent of filter order

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/3916

#### Description

The motivation for the changes in this PR started with the goal to support including multi day events in the app feed. 

In https://github.com/ding2/ding2/pull/871 the ability to show multi day events on the frontpage and on the event list was added. Normally events are only shown when start date is less than or equal to current date. But in above PR a settings was added to ding_event content type, that gives administrators the ability to show these events also when end date is less than or equal to current date.

Unfortunately, this was implemented in a way that is dependent on the order of the views filters from the view UI. Several issues with this has been reported. For example see: https://platform.dandigbib.org/issues/3442. I also found out that if you disable the "Show only promoted events" no events will be shown at all, since a filter is removed from the view messing up the order of the filters.

Now this can actually all be set up in Views UI with conditional groups (se my screenshot below), but on the event list we have an exposed filter on the start date, and we want the input from the exposed filter to also take the event **end** date into account, if the event is set up to be shown on multiple days. It's not possible to have an exposed filter which takes another field into account based on some condition.

So what I've done in this PR is:

1. Refactor our SQL-rewriting to support multi day events, so that it isn't dependent on the order of the filters. In connection with this, I have moved the rewriting to hook_views_query_alter() instead, since it seemed easier to get to the conditions this way. This fixes all the related problems mentioned above.
2. I have configured the ding_app_content_rss event feed, so that it can show multi day events if the event is set to do so. I have done this directly in the Views configuration in the UI, since for this view we don't need an exposed filter, so there's no need to deal with the complexity of SQL-rewriting.

#### Screenshot: multi day event configuration on Views UI

3916-mutliday-setup-in-view

#### Checklist

### DIFF
--- a/modules/ding_app_content_rss/ding_app_content_rss.views_default.inc
+++ b/modules/ding_app_content_rss/ding_app_content_rss.views_default.inc
@@ -314,6 +314,78 @@ function ding_app_content_rss_views_default_views() {
   $handler->display->display_options['sorts']['field_ding_event_date_value']['id'] = 'field_ding_event_date_value';
   $handler->display->display_options['sorts']['field_ding_event_date_value']['table'] = 'field_data_field_ding_event_date';
   $handler->display->display_options['sorts']['field_ding_event_date_value']['field'] = 'field_ding_event_date_value';
+  $handler->display->display_options['defaults']['filter_groups'] = FALSE;
+  $handler->display->display_options['filter_groups']['operator'] = 'OR';
+  $handler->display->display_options['filter_groups']['groups'] = array(
+    1 => 'AND',
+    2 => 'AND',
+  );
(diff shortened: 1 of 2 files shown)


JrgenGNielsen:
"I have done this directly in the Views configuration in the UI"
- this is exported to ding_app_content_rss.views_default.inc, right?

holt83:
> "I have done this directly in the Views configuration in the UI"
> 
> * this is exported to ding_app_content_rss.views_default.inc, right?

Yes it should all be exported in this PR

JrgenGNielsen:
reviewed and approved

PR #1370, opened 2019-02-19
4159: Check if WAYF was used in pin validation

cableman:
#### Link to issue

#### Description

Check if the user was logged in with WAYF and disable pin-code length check. If not WAYF login will not be possible.

This may not be the bedst solution in relation to other SSO. So other than removing the validation I'm open for other suggestions.

#### Screenshot of the result

No screenshot

#### Checklist

No comments


cableman:
Fast note: The pin is not empty it's the WAYF shared password from settings.php.

kasperg:
... but supplied using the login form which does not trigger ``ding_user_element_validate_pincode`.

PR #1372, opened 2019-02-22
3966 by Inlead: Define mobile flags for CTs.

NicolaeGrigoriu:
#### Link to issue

http://platform.dandigbib.org/issues/3966

#### Description

In https://github.com/ding2/ding2/pull/1095 the mobile viewing was enabled by default for News and Events.

However, the change did not include a definition that this configuration is part of the respective modules' configuration. It should be added to the info file of the modules so that we ensure that the configuration is not removed inadvertently through Feature Updates eg. https://github.com/ding2/ding2/pull/1266.

#### Checklist


JrgenGNielsen:
reviewet og godkendt

PR #1376, opened 2019-02-26
3901: Fix custom click parameters for several events

holt83:
#### Link to issue

#### Description

Apparently it's not supported to use variables in object literals.
As a consequence all of the webtrekk events handled client side were not
sending the correct wtk ID.

The easiest fix seemed to be to instantiate the object first and then use bracket notation with the variables.

#### Checklist

### DIFF
--- a/modules/ding_webtrekk/ding_webtrekk.js
+++ b/modules/ding_webtrekk/ding_webtrekk.js
@@ -27,12 +27,19 @@
     return url + seperator + urlParameter;
   };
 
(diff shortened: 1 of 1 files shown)


holt83:
I added new pushEvent utility function in the client, to deal with undefined wts-variable in a centralized place.

holt83:
Scrutinizer might still complain, but at least I have handled the situation now.

JrgenGNielsen:
reviewed and approved

PR #1378, opened 2019-02-28
4159: Remove antibot protection from user_login_block

holt83:
#### Link to issue

#### Description
Remove antibot protection from user_login_block.

This was causing issues with login via gatewayf since it submits the
form programmatically.

#### Checklist


holt83:
Closed in favor of #1379

PR #1379, opened 2019-02-28
4159: Antibot prevents WAYF login

cableman:
#### Link to issue

#### Description

Ding Gate WAYF submits the login_block_form as part of it's login process. Which get's flags by antibot module as spam.

So NemID login succeeds, but drupal login fails and the user ends up at the registry as new user page.  

#### Screenshot of the result

No screenshot.

#### Checklist

No comments.


cableman:
Circleci writes "Minor Service Outage" and the tests complains that Chrome is not there!

PR #1380, opened 2019-02-28
4194: opensearch cache local ids

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/4194

#### Description

The uncached id's were always passed to setObjectIds(). As a result requests based on local ids will fail.

An example is when we translate local ids from the status lists to PID by finding the namespace after looking up the local ids in getObject. These will always end up here:

https://github.com/ding2/ding2/blob/master/modules/opensearch/includes/opensearch.search.inc#L631 

As a consequence we are currently using the wrong namespace for a lot of materials in status lists.

There were also some additional problems were wouldn't hit the cache, and the objects returned wasn't keyed by PID when local ids was used in the request.

I realize that I'm setting both setObjectIds() and setLocalIds() when TingObjectRequest doesn't actually support requests with a mix of both types of identifiers. But currently I think no request are made with a mix. It's either full local or full pid. So it's no problem. But opensearch getObject does actually support it, so we could make it work if need be. The code should be ready for it if TingObjectRequest is modified to support it.

#### Checklist

I also sneaked in a small correction to opensearch_get_objects(). Opensearch execute doesn't return empty array when no results, but NULL, giving rise to "Unsupported operand" error in the current state.

### DIFF
--- a/modules/opensearch/opensearch.client.inc
+++ b/modules/opensearch/opensearch.client.inc
@@ -447,6 +447,16 @@ function opensearch_execute(TingClientRequest $request) {
       $request->setLocalIds(array_keys($uncached_local_ids));
     }
 
+    // The $cached_local_ids array is now keyed by local id and we needed that
+    // to get the uncached local ids. But in the response all objects should be
+    // keyed PID, so here we convert the array of cached "local" objects to be
+    // keyed by their PID.
+    $local_object_pids = [];
+    foreach ($cached_local_ids as $object) {
+      $local_object_pids[] = $object->id;
+    }
+    $cached_local_ids = array_combine($local_object_pids, $cached_local_ids);
+
     $cached_objects = $cached_object_ids + $cached_local_ids;
 
     // If there are no ids to fetch then there is no need to issue a request.
@@ -455,10 +465,6 @@ function opensearch_execute(TingClientRequest $request) {
       _opensearch_cache($request, $cached_objects);
       return $cached_objects;
     }
-
-    // Re-set uncached ids in the request so the missing will be fetched from
-    // the data-well.
-    $request->setObjectIds(array_keys($uncached_ids));
   }
 
   try {
--- a/modules/opensearch/opensearch.client.inc
+++ b/modules/opensearch/opensearch.client.inc
@@ -92,13 +92,6 @@ function opensearch_get_objects(array $ids, $local = FALSE) {
     $request->setProfile($profile);
   }
 
-  if ($local) {
-    // Ensure local ids are integer values and not strings. Else we will not
-    // hit the cache, since they will be converted to integers in our caching
-    // logic in opensearch_execute(), if they are passed as strings.
-    $ids = array_map('intval', $ids);
-  }
-
   // Opensearch getObject complains if it has to return more than 200 records,
   // with the message 'getObject can fetch up to 200 records. '. But it seems
   // it also have trouble fething over 169 records, which appears to be a bug.
@@ -450,7 +443,16 @@ function opensearch_execute(TingClientRequest $request) {
     }
     $uncached_local_ids = array_diff_key($local_ids, $cached_local_ids);
     if (!empty($uncached_local_ids)) {
-      $request->setLocalIds(array_keys($uncached_local_ids));
+      $uncached_local_ids = array_keys($uncached_local_ids);
+      // Datawell ids needs to be strings, since for example the id '06606059'
+      // is not the same as '6606059'. But PHP might have converted some of the
+      // local ids to integers when using them as array keys. For example the id
+      // '20174870' would be converted to integer when using as key. It's
+      // important that the request object is exactly the same when using it as
+      // a base for cache ID, so here we ensure that all local ids are added to
+      // the request as string values.
+      $uncached_local_ids = array_map('strval', $uncached_local_ids);
+      $request->setLocalIds($uncached_local_ids);
     }
 
     // The $cached_local_ids array is now keyed by local id and we needed that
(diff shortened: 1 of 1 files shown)


modules/opensearch/opensearch.client.inc
```diff
@@ -105,7 +112,10 @@ function opensearch_get_objects(array $ids, $local = FALSE) {
     else {
       $request->setObjectIds($chunk);
     }
-    $objects += opensearch_execute($request);
+    $result = opensearch_execute($request);
+    if (is_array($result)) {
```
kasperg: I do not understand the check here - and perhaps not the previous state of the code either. According to the DocBlock `opensearch_execute` returns `\TingClientSearchResult|bool`. This change is based on the option that the function will return an array.

If the DocBlock is wrong then it should be updated accordingly.

modules/opensearch/opensearch.client.inc
```diff
@@ -413,17 +421,47 @@ function opensearch_execute(TingClientRequest $request) {
...
+    }
+    $uncached_local_ids = array_diff_key($local_ids, $cached_local_ids);
+    if (!empty($uncached_local_ids)) {
+      $request->setLocalIds(array_keys($uncached_local_ids));
+    }
+
+    // The $cached_local_ids array is now keyed by local id and we needed that
+    // to get the uncached local ids. But in the response all objects should be
+    // keyed PID, so here we convert the array of cached "local" objects to be
+    // keyed by their PID.
```
kasperg: :+1:

modules/opensearch/opensearch.client.inc
```diff
@@ -92,6 +92,13 @@ function opensearch_get_objects(array $ids, $local = FALSE) {
     $request->setProfile($profile);
   }
 
+  if ($local) {
+    // Ensure local ids are integer values and not strings. Else we will not
+    // hit the cache, since they will be converted to integers in our caching
+    // logic in opensearch_execute(), if they are passed as strings.
+    $ids = array_map('intval', $ids);
```
kasperg: I wonder if this the right level for addressing such an issue or it should be done lower down the call stack.

modules/opensearch/opensearch.client.inc
holt83: It either returns an array or NULL. Will update DocBlock.

holt83: @kasperg I have updated the docblock to describe exactly what is being returned based on the type of request passed to opensearch_execute().

holt83:
I have updated the PR:

1. opensearch_execute() docblock now correctly describes what it returns based on the type of $request.

2. I discovered that local ids have to be strings. This is also documented in the docblock for opensearch_get_objects(). It's also important that the ids are strings when generating cache id, so instead of converting them to intval I now convert them to string values before passing them on to the request object. More about this in code/commit comments.

3. I discovered an error where _opensearch_cache() would sometimes return the wrong cached response. I found out this happens because we merged the cached response before setting cache. resulting in a response that doesn't fit the cache.

For example making the following requests:

1. opensearch_get_objects(['06606059'], TRUE);
2. opensearch_get_objects(['06606059', '06472214'], TRUE);

In the second request it will find the '06606059' and make a request for '06472214'. Afterwards it would merge in the response for '06606059' and put it in cache. This means that a single object request now would point to a response of two objects, giving rise to all sorts of hard to explain behavior.

modules/opensearch/opensearch.client.inc
holt83: I actually found out that they need to be strings. More in comment below and in code/commit messages.

modules/opensearch/opensearch.client.inc
```diff
@@ -454,19 +497,20 @@ function opensearch_execute(TingClientRequest $request) {
...
-      // Merge in any cached objects. The map uses ids as keys. The order does
-      // not matter for object requests.
-      $response = array_merge($cached_objects, $response);
-    }
-
     // Update the cache with the raw response object from the data well for
     // faster processing. Also set the static cache.
     _opensearch_cache($request, $response);
 
+    // Merge any cached objecst that we found earlier with the result from the
```
cableman: If this was moved above and the request pid/local was reset. It could be store in cache. Else the store partial search it not likely to be search for again?

holt83: I get what you're saying, but I think it's a minor issue. In the next request for the same objects, the response will be reassembled from the cached objects and the full response will actually [be stored in the cache](https://github.com/vejlebib/ding2/blob/4194-opensearch-cache-local-ids/modules/opensearch/opensearch.client.inc#L473)

I think that if we should make this even better, we should instead store the individual objects in the cache. That way there's a much higher chance that a given object request will be able to re-assemble the response from cache without requesting opensearch.

holt83: Wait.. you're right. It will actually make a new request if the $response we put in cache has more that one object. 

So we need to change something here. I can follow your suggestion and restore the original ids in the request or I can store the individual objects in cache as I suggested.

What do you think?

holt83: I realize that we don't have much time, so I've just started working on you suggestion now. Will update PR soon.

cableman: The individual objects will be stored in the predict cache function.

holt83: Unfortunately only for collection and search request and not getObject request for mutliple objects :(

I Updated the PR with your suggestion to restore original request before setting cache. A nice side effects of this was that I could remove the code that converted the ids to string values, since we remember the original array.

cableman: :-)

modules/opensearch/opensearch.client.inc
kasperg: Thanks. Much appreciated.

PR #1381, opened 2019-03-04
3912: Underline links

xendk:
https://platform.dandigbib.org/issues/3912

#### Description

Adds underline to links in body on event, news and groups.

#### Screenshot of the result

image

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1383, opened 2019-03-07
4123 by Inlead: Stopped requests to eKurser.

bozeak:
#### Link to issue

https://platform.dandigbib.org/issues/4123

#### Description

eKurser embedded materials are not visible. This commit adds "group_oembed" fieldgroup to ting_object's "Right column" field group, so it will be displayed under "Materials display" field.

#### Screenshot of the result

image

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1384, opened 2019-03-07
3936: Fix missing titles for ILL on loan/reservation

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/3936

#### Description

In the latest versions a lot of libraries has experienced the "Error: unknown/missing/inaccessible record:" problem with materials on statuslists.

I have posted a pretty hacky attempt to fix this here: https://github.com/ding2/ding2/pull/1293, but it relies on a weird way of handling not-found object in opensearch getObject. We are in contact with DBC and they will update opensearch to handle it in a better way.

This should only be a problem with ILL loans/reservations since we should be able to always get an entity for materials in the libraries own catalog.

During analysis in https://platform.dandigbib.org/issues/3936 we have found out that we actually shouldn't try to make a request against opensearch at all for ILL, since the ID we recieve from FBS API isn't guaranteed to be a faust and hence there's a risk for collisions with wrong materials from other sources than the library catalog.

So I believe a better fix for this right now is to **not** make an opensearch request for ILL. This should fix all the cases with  "Error: unknown/missing/inaccessible record:" on loan/reservation.

The main approach is that we don't send a ding_entity_id from the providers when we have an ILL-bibliographic record on the response from FBS API. Ding_loan/reservation will check for the entity and if provider didn't set ding_entity_id, NULL will be returned. Before we were actually sending a wrong ding_entity_id in some cases.

We will still have problem with debts, since we have no idea whether it's ILL there. But we will handle that in another case, since we will have to wait for Systematic to (hopefully) update the API with more ILL-information.

Off course we should also try to fix the "Errors as real object"-problem that my previous PR tried to address, but we will also handle that in another case, since we will have to wait for DBC to update opensearch.

#### Checklist

I found a weird quirk in the DingProviderReservation and DingProviderLoan objects, where it would try to make a very problematic opensearch request, when the object has been serialized in ding_session_cache(). More about that in the related commit message.

Tbh I'm not really a fan of these classes, and have a very hard time seeing the usefulness of all the complexity in these, but I have tried to fix this without changing too much.

I did experience much better performance on the status lists after fixing it, because it doesn't make the very problematic request agaisnt opensearch.

### DIFF
--- a/modules/ding_reservation/ding_reservation.module
+++ b/modules/ding_reservation/ding_reservation.module
@@ -490,7 +490,8 @@ function ding_reservation_reservations_form($form, &$form_state, $reservations, 
   }
 
   foreach ($reservations as $reservation) {
-    $mat_type = $creators = $title = NULL;
+    $title = t('Title not available');
+    $creators = $mat_type = $cover = NULL;
 
     // If the provider was able to produce an entity we'll use that since we can
     // show more information and link to the entity.
@@ -501,6 +502,7 @@ function ding_reservation_reservations_form($form, &$form_state, $reservations, 
       $title = l($entity->getTitle(), $uri['path']);
       $creators = $entity->getCreators();
       $mat_type = $entity->getType();
+      $cover = field_view_field('ting_object', $entity, 'ting_cover', 'user_list');
     }
     // Else the provider has the option to supply a display name and author.
     elseif (!empty($reservation->display_name)) {
@@ -521,7 +523,7 @@ function ding_reservation_reservations_form($form, &$form_state, $reservations, 
           '#creators' => $creators,
           '#material_type' => $mat_type,
           '#title' => $title,
-          '#cover' => field_view_field('ting_object', $entity, 'ting_cover', 'user_list'),
+          '#cover' => $cover,
           '#information' => array(
             'pickup_id' => array(
               'label' => t('Pickup id'),
--- a/modules/ding_reservation/ding_reservation.module
+++ b/modules/ding_reservation/ding_reservation.module
@@ -561,7 +561,7 @@ function ding_reservation_reservations_form($form, &$form_state, $reservations, 
           '#creators' => $creators,
           '#material_type' => $mat_type,
           '#title' => $title,
-          '#cover' => field_view_field('ting_object', $entity, 'ting_cover', 'user_list'),
+          '#cover' => $cover,
           '#information' => array(
             'queue_number' => array(
               'label' => t('Queue number'),
@@ -609,7 +609,7 @@ function ding_reservation_reservations_form($form, &$form_state, $reservations, 
           '#creators' => $creators,
           '#material_type' => $mat_type,
           '#title' => $title,
-          '#cover' => field_view_field('ting_object', $entity, 'ting_cover', 'user_list'),
+          '#cover' => $cover,
           '#information' => array(
             'ill_status' => array(
               'label' => t('Status'),
--- a/modules/ding_reservation/ding_reservation.module
+++ b/modules/ding_reservation/ding_reservation.module
@@ -495,7 +495,7 @@ function ding_reservation_reservations_form($form, &$form_state, $reservations, 
 
     // If the provider was able to produce an entity we'll use that since we can
     // show more information and link to the entity.
-    if (!empty($reservation->ding_entity_id)) {
+    if (is_object($reservation->entity)) {
       $entity = $reservation->entity;
 
       $uri = entity_uri('ting_object', $entity);
--- a/modules/ding_reservation/ding_reservation.module
+++ b/modules/ding_reservation/ding_reservation.module
@@ -509,9 +509,6 @@ function ding_reservation_reservations_form($form, &$form_state, $reservations, 
       $title = $reservation->display_name;
       $creators = $reservation->author;
     }
-    else {
-      $title = t('Title not available');
-    }
 
     $pickup_branch = ding_provider_invoke('reservation', 'branch_name', $reservation->pickup_branch_id);
     switch ($reservation->reservation_type) {
--- a/modules/ding_reservation/ding_reservation.module
+++ b/modules/ding_reservation/ding_reservation.module
@@ -490,25 +490,26 @@ function ding_reservation_reservations_form($form, &$form_state, $reservations, 
   }
 
   foreach ($reservations as $reservation) {
-    /** @var \TingEntity $entity */
-    $entity = $reservation->entity;
-    if (!is_object($entity) || NULL === $entity->getTingObject()) {
-      $entity = ding_provider_get_pseudo_entity($reservation->ding_entity_id);
-      // Don't link the title as the data well do not known this title. It may
-      // be a inter library loan.
-      $title = (!empty($entity) && $entity->getTitle()) ? $entity->getTitle() : $reservation->display_name;
-    }
-    else {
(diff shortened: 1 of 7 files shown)


holt83:
I get "Only string literals should be passed to t() where possible" in Scrutinizer, but it's something I've added in this PR.

modules/ding_reservation/ding_reservation.module
```diff
@@ -490,25 +490,28 @@ function ding_reservation_reservations_form($form, &$form_state, $reservations,
...
+      $creators = $entity->getCreators();
+      $mat_type = $entity->getType();
+      $cover = field_view_field('ting_object', $entity, 'ting_cover', 'user_list');
+    }
+    // Else the provider has the option to supply a display name and author.
+    elseif (!empty($reservation->display_name)) {
+      $title = $reservation->display_name;
+      $creators = $reservation->author;
+    }
+    else {
```
cableman: Is this else needed will it not default to this from line 493?

holt83:
@cableman Sorry, I can see the description/ticket history is a bit confusing.

It will use a data-well object if the provider sets a ding_entity_id. And FBS only sets a ding_entity_id if the loan/reservation is **not** ILL. That is if there's **no** ilBibliographicRecord present.

[This call](https://github.com/ding2/ding2/pull/1384/files#diff-415a10980167e5f354fa41ea380ee169R498) will produce a TingEntity object if a valid ding_entity_id was passed to the provider, which should be the case for all non-ILL loan/reservations.

modules/ding_reservation/ding_reservation.module
holt83: I missed that one. It's not needed. Will remove.

holt83:
Updated PR to remove unneeded else statement.

GitCop:

 * Commit: 41977d6262592eb04f20b4f90409a41315ca600a

---

GitCop:

 * Commit: 41977d6262592eb04f20b4f90409a41315ca600a

---

holt83:
Scrutinizer: 

- It complains about wrong identation, but I can't see anything wrong. Perhaps it doesn't like bracket notation for arrays? In any case I think we should ignore.

- The other errors are not related to the changes in this, but I have tried to fix them anyway.

modules/ding_webtrekk/ding_webtrekk.module
```diff
@@ -280,7 +280,15 @@ function ding_webtrekk_form_ding_loan_loans_form_alter(&$form, &$form_state) {
...
     if ($loan->renewable) {
-      $material_ids[$loan->id] = $loan->getEntity()->getId();
+      if (is_object($loan->entity)) {
+        $material_ids[$loan->id] = $loan->entity->getId();
+      }
+      // For interlibrary loans the material may not reachable in the well and
+      // therefore we can't get a well ID to display. We'll note that it's an
+      // ILL and use the title instead.
+      else {
+        $material_ids[$loan->id] = 'Interlibrary Loan: ' . $loan->display_name;
```
cableman: This should be wrapped un t() to enable translation of the title also into webtrekk I think.

holt83: You're right. I'll update PR.

holt83:
Updated PR with @cableman suggestion.

I tried to fix Scrutinizer, but I'm unsure how to handle:
> Multi-line assignments must have the equal sign on the second line

Also it complains about identation in loans.inc, but when I look it seems fine. Perhaps Scrutinizer doesn't like brackets instead of array() in this situation?

modules/ding_webtrekk/ding_webtrekk.module
```diff
@@ -287,7 +287,9 @@ function ding_webtrekk_form_ding_loan_loans_form_alter(&$form, &$form_state) {
       // therefore we can't get a well ID to display. We'll note that it's an
       // ILL and use the title instead.
       else {
-        $material_ids[$loan->id] = 'Interlibrary Loan: ' . $loan->display_name;
+        $material_ids[$loan->id] = t('Interlibrary Loan: %title', [
```
JrgenGNielsen: don't use the [ ] syntax - use array() instead

holt83: Are you sure about that? From the README:

JrgenGNielsen: det er hvad der står i [Code Guidelines](https://github.com/ding2/ding2/blob/master/docs/code_guidelines.md). Ellers så skal dén vist opdateres... ;-P

GitCop:

 * Commit: 7f5ab9f8f6231cce37e84e4cb5e39f5ddc885d65

---

GitCop:

 * Commit: 7f5ab9f8f6231cce37e84e4cb5e39f5ddc885d65

---

modules/ding_webtrekk/ding_webtrekk.module
holt83: You're right, but as I mentioned we do use bracket notation in a lot of other places and in new PR's they are often used.

Anyway, I updated the PR to use array() notation.

JrgenGNielsen:
could I get you to fix the GitCop error? also, try to see if the array() notation fixes the Scrutinizer errors

holt83:
My bad with the git cop. It's fixed now.

holt83:
Indentation issues were fixed but scrutinizer still complains about:

> Multi-line assignments must have the equal sign on the second line

I can move the equal sign down a line, but do we want to do this: https://www.drupal.org/project/coder/issues/2710067

holt83:
Ok, I just updated the PR to do exactly as Scrutinizer even though [I agree it looks less readable](https://www.drupal.org/project/coder/issues/2710067). 

So Scrutinizer and everything looks fine now. Let's get on with this :)

JrgenGNielsen:
reviewed and approved

modules/fbs/includes/fbs.loan.inc
```diff
@@ -48,13 +48,22 @@ function fbs_loan_list($account, $reset = FALSE) {
...
           'materials_number' => $loan->loanDetails->materialItemNumber,
         );
 
+        // With the FBS provider we have to use FBS data for ILL, since we can't
+        // rely on the record ID being unique in these cases.
+        if (!empty($loan->loanDetails->ilBibliographicRecord)) {
+          $loan_data['display_name'] = $loan->loanDetails->ilBibliographicRecord->title;
+          $loan_data['author'] = $loan->loanDetails->ilBibliographicRecord->author;
+        }
+        elseif (isset($ids[$loan->loanDetails->recordId])) {
```
guddo: @holt83 Lots of code seems to rely in the presence of `ding_entity_id`. In this clause you might be in a situation where this critical value is not set., for example here: https://github.com/ding2/ding2/blob/master/modules/p2/ding_serendipity/ding_serendipity.module#L419-L437.

Maybe this line should not be wrapped in an `elseif`?

holt83: I think that ding_serendipity should check for an entity like for example ding_loan does and just skip the loan if it's not present. It means the provider didn't return an entity for some reason. With FBS as provider it's because the loan is an ILL and it should not be on any ding_list and therefore ding_serendipity should ignore it. 

Example here: https://github.com/ding2/ding2/blob/master/modules/ding_loan/plugins/content_types/loans.inc#L179

PR #1387, opened 2019-03-14
3974 by Inlead: Sorting of Events is changed

NicolaeGrigoriu:
#### Link to issue

http://platform.dandigbib.org/issues/3974

#### Description

At some point between the latest releases, the sorting/display of Events have changed.
**Release 29 (7.x-4.5.x):**
image

**Release 30 (7.x-4.6.x):**
image

#### Screenshot of the result

image

#### Checklist

The problem is not in the releases. This behavior occurs when 2 or more nodes in the view doesn't have images. It tries to stack them in a single column.

### DIFF
--- a/themes/ddbasic/scripts/view.js
+++ b/themes/ddbasic/scripts/view.js
@@ -207,14 +207,6 @@
                 has_image = false;
               }
 
-              // If odd and has image.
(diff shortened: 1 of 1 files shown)


JrgenGNielsen:
Scrutinizer error is unrelated to this PR.
reviewed and approved

PR #1388, opened 2019-03-18
3756 by Inlead: Add Material list to Event and Page CTs

NicolaeGrigoriu:
#### Link to issue

https://platform.dandigbib.org/issues/3756

#### Description

Currently it seems like its only News that supports referencing of Materials to be displayed as carousel.

Event CT and Page CT should also support this.

#### Screenshot of the result
**Pace CT**
image

**Event CT**
image

#### Checklist


guddo:
@cableman This is the layout used on News CT at the moment.

PR #1393, opened 2019-03-22
1404: Dont delete field data on p2b API error

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/1404

#### Description

I discovered this while testing the new integration module on our production database:

If something goes wrong communicating with p2b API when updating existing event that is sync'ed with p2b, the code will override the field data with empty array, destroying the the link between p2b and CMS.

This PR introduces some better handling of these situations.

#### Screenshot of the result

#### Checklist

### DIFF
--- a/modules/ding_place2book/ding_place2book.fields.inc
+++ b/modules/ding_place2book/ding_place2book.fields.inc
@@ -414,13 +414,19 @@ function ding_place2book_field_presave($entity_type, $entity, $field, $instance,
     case 'ding_p2b':
       if (!empty($items)) {
         $settings = reset($items);
+        $synchronized_items = [];
         // Ensure that both 'synchronize' and 'do_sync' is set and TRUE, before
         // making any requests to p2b API. This is to ensure we don't make lots
         // of unneeded requests when nodes are saved programmatically or updated
         // in bulk via the UI.
         // See ding_place2book_field_widget_form() for more info.
         if ($settings['synchronize'] && !empty($settings['do_sync'])) {
-          $items = _ding_place2book_sync_p2b_entities($entity, $settings);
+          $synchronized_items = _ding_place2book_sync_p2b_entities($entity, $settings);
+        }
+
+        // Only update field data if synchronizationn was successful.
+        if (!empty($synchronized_items)) {
+          $items = $synchronized_items;
         }
         else {
           $items = array(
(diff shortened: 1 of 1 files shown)


modules/ding_place2book/ding_place2book.fields.inc
```diff
@@ -414,13 +414,20 @@ function ding_place2book_field_presave($entity_type, $entity, $field, $instance,
...
+        // Boolean indicating if the p2b synchronizationn was successful.
+        $synchronized = FALSE;
         // Ensure that both 'synchronize' and 'do_sync' is set and TRUE, before
         // making any requests to p2b API. This is to ensure we don't make lots
         // of unneeded requests when nodes are saved programmatically or updated
         // in bulk via the UI.
         // See ding_place2book_field_widget_form() for more info.
         if ($settings['synchronize'] && !empty($settings['do_sync'])) {
-          $items = _ding_place2book_sync_p2b_entities($entity, $settings);
+          $synchronized = _ding_place2book_sync_p2b_entities($entity, $settings);
```
kasperg: This seems odd to me. According to the DocBlock `_ding_place2book_sync_p2b_entities` returns an array. This change seems to assume that it can be a boolean. Does the DocBlock need updating or are we handling a case where we might get an empty array returned in case on an error?

Could we simplify the change to something like:

```php
$synchronized_items = _ding_place2book_sync_p2b_entities($entity, $settings);
// Only update field data if synchronization was successful.
if (!empty($synchronized_items)) {
  $items = $synchronized_items;
}
```

modules/ding_place2book/ding_place2book.fields.inc
```diff
@@ -551,11 +558,19 @@ function _ding_place2book_sync_p2b_entities($entity, array $settings) {
...
+          'event_maker_id' => $event_maker_id,
+          'synchronize' => 1,
+        ),
+      );
     }
     catch (Exception $ex) {
       watchdog_exception('ding_place2book', $ex);
       drupal_set_message($ex->getMessage(), 'error');
-      return $items;
+      return FALSE;
```
kasperg: This seems to indicate that the DocBlock should be updated or that we should just make it explicit that returning an empty array means that no items are sync'ed.

modules/ding_place2book/ding_place2book.fields.inc
holt83: With your suggestion we don't have to change return value and update docblock, so I think we should go with that. I'll update PR.

PR #1396, opened 2019-03-25
4055 by Inlead: DAMS should be enabled by default.

bozeak:
#### Link to issue

http://platform.dandigbib.org/issues/4055

#### Description

Adds hook_update() into profile which will enable Ding DAMS module.
**NOTE:** This PR must be merged **only after** https://github.com/ding2/ding2/pull/1332 is merged.

#### Checklist

### DIFF
--- a/ding2.install
+++ b/ding2.install
@@ -726,3 +726,10 @@ function ding2_update_7063() {
     'ding_nodelist',
   ));
 }
+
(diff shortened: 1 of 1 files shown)


guddo:
@bozeak What about new installations? A dependency should be defined somewhere.

PR #1397, opened 2019-03-25
4196: Ensure fixed header elements doesn't overlap in-page anchors

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/4196

#### Description

Ensure that our fixed header elements doesn't overlap in-page anchors.

#### Screenshot of the result - before
4196-before-fix

#### Screenshot of the result - after
4196-aftere-fix

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1398, opened 2019-03-26
4150: Fixed ting-object accordion wrapper

b14sha:
#### Link to issue

https://platform.dandigbib.org/issues/4150

#### Description

In some cases the field-group-format-wrapper has a zero height in the 'group-on-this-site' accordion section of a ting-object. This causes the accordion to fail. A simple clearfix on the element solves the issue.

#### Screenshot of the result

Screenshot 2019-03-26 at 14 19 31


JrgenGNielsen:
reviewed and approved

PR #1401, opened 2019-03-29
4199 by Inlead: Nodelist styling changes.

bozeak:
#### Link to issue

https://platform.dandigbib.org/issues/4199

#### Description

Configuration forms rendered in Ding IPE popup takes default site's theme, which is not adapted yet.
In this commit are added styles which are loaded when popup is opened, so those styles are not breaking anything else.

#### Screenshot of the result

ads

#### Checklist


JrgenGNielsen:
reviewet og godkendt

PR #1405, opened 2019-04-05
4257 by Inlead: Various layout optimizations for ticket buttons.

inlead:
#### Link to issue

https://platform.dandigbib.org/issues/4257

#### Description

- Green ticket-buttons have white underlined text
- Waiting list buttons are now yellow instead of green
- Non-P2b ticket-buttons are now looking like P2b ticket-buttons (green coloured)
- Complete P2b ticket-button is now clickable
- Ticket sale start date and time is shown inside button

#### Checklist

### DIFF
--- a/modules/ding_place2book/ding_place2book.module
+++ b/modules/ding_place2book/ding_place2book.module
@@ -690,7 +690,7 @@ function ding_place2book_get_event_state($event_maker_id, $event_id) {
         }
       }
       else {
-        // Else we have no other options than to fetchs the tickets and look
+        // Else we have no other options than to fetch the tickets and look
         // for open or upcoming sale periods.
         $sale_periods = [];
         foreach ($p2b->getPrices($event_maker_id, $event_id) as $price) {
@@ -716,23 +716,27 @@ function ding_place2book_get_event_state($event_maker_id, $event_id) {
           return $p['end'] >= $now;
         });
 
-        // Find the closest sale period.
-        $begin_values = array_column($sale_periods, 'begin');
-        $min_period_begin = min($begin_values);
-        $min_period = array_filter($sale_periods, function ($sale_period) use ($min_period_begin) {
-          return ($sale_period['begin']) == $min_period_begin;
-        });
-        $min_period = array_shift($min_period);
+        // Check if we still have any active sale periods. If such are - go on!
+        if (!empty($sale_periods)) {
+          // Find the closest sale period.
+          $begin_values = array_column($sale_periods, 'begin');
+          // Get the closest sale period. In case when event has upcoming sales,
+          // its start value will be used in button label.
+          $min_period_begin = min($begin_values);
+          $min_period = array_filter($sale_periods, function ($sale_period) use ($min_period_begin) {
+            return ($sale_period['begin']) == $min_period_begin;
+          });
+          $min_period = array_shift($min_period);
 
-        if ($now >= $min_period['begin'] && $now <= $min_period['end']) {
-          // We have found an open sales period. No need to continue iterating
-          // the ticket types.
+          // We have found an open sales period.
           $ticket_link_status = 'open';
-        }
-        elseif ($now < $min_period['begin']) {
-          // We haven't found any open sales period yet, but at least there's
-          // an upcoming, so set that status for now.
-          $ticket_link_status = 'upcoming';
+
+          // Check for 'upcoming' sale.
+          if ($now < $min_period['begin']) {
+            // We haven't found any open sales period yet, but at least there's
+            // an upcoming, so set that status for now.
+            $ticket_link_status = 'upcoming';
+          }
         }
 
         // If no upcoming or open sales period was found for the published event
--- a/modules/ding_place2book/ding_place2book.module
+++ b/modules/ding_place2book/ding_place2book.module
@@ -661,6 +661,9 @@ function ding_place2book_get_event_state($event_maker_id, $event_id) {
     $meta_data = $event->meta_data;
     $state = $event->state;
     $available_tickets = $meta_data->available_tickets;
+    $data = '';
+    $class = '';
+    $type = '';
 
     // Try to derive a status for the ticket link in frontend by looking at the
     // p2b event's state, ticket types and sale periods.
@@ -730,36 +733,63 @@ function ding_place2book_get_event_state($event_maker_id, $event_id) {
     if (!empty($ticket_link_status)) {
       switch ($ticket_link_status) {
         case 'open':
-          $data = l(t('Book a ticket'), $event->links->sales_location, array('attributes' => array('target' => '_blank')));
           $class = 'ticket-available';
+          $data = l(t('Book a ticket'), $event->links->sales_location, array(
+            'attributes' => array(
+              'target' => '_blank',
+              'class' => array('button', $class),
+            ),
+          ));
+          $type = 'link';
           break;
 
         case 'closed':
-          $data = t('Not open for ticket sale');
           $class = 'ticket-notice';
+          $data = t('Not open for ticket sale');
+          $type = 'markup';
           break;
 
         case 'sold-out':
-          $data = t('Sold out');
           $class = 'ticket-warning';
+          $data = t('Sold out');
+          $type = 'markup';
           break;
 
         case 'upcoming':
-          $data = t('Ticket sale has not yet started for this event');
+          $sale_start = $price->sale_begin_at;
+          $date = date('d.m.Y', strtotime($sale_start));
+          $time = date('H:i', strtotime($sale_start));
           $class = 'ticket-notice';
+          $data = t("Ticket sale starts @date at @time", array(
+            '@date' => $date,
+            '@time' => $time,
+          ));
+          $type = 'markup';
           break;
 
         case 'waiting-list':
-          $data = l(t('Waiting list open'), $event->links->sales_location, array('attributes' => array('target' => '_blank')));
           $class = 'ticket-notice';
+          $data = l(t('Waiting list open'), $event->links->sales_location, array(
+            'attributes' => array(
+              'target' => '_blank',
+              'class' => array('button', $class),
+            ),
+          ));
+          $type = 'link';
           break;
 
         case 'event-over':
           $data = t('The event has already taken place');
-          $class = '';
+          $type = 'markup';
           break;
       }
-      $output = "<span class='place2book button {$class}'>{$data}</span>";
+
+      if ($type == 'link') {
+        $output = $data;
+      }
+      elseif ($type == 'markup') {
+        $output = "<div class='place2book button {$class}'>{$data}</div>";
+      }
       print ajax_render($output);
     }
   }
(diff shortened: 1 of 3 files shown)


modules/ding_place2book/ding_place2book.module
```diff
@@ -571,36 +574,63 @@ function ding_place2book_get_event_state($event_maker_id, $event_id) {
...
         case 'sold-out':
-          $data = t('Sold out');
           $class = 'ticket-warning';
+          $data = t('Sold out');
+          $type = 'markup';
           break;
 
         case 'upcoming':
-          $data = t('Ticket sale has not yet started for this event');
+          $sale_start = $price->sale_begin_at;
```
holt83: So where does this $price variable come from? Is it the last assignment to $price in the foreach further up? Are we even sure this is the next open sales period?

I think we need to remember the closest next sales period, when we find an upcoming sales period when iterating the sales periods.

modules/ding_place2book/ding_place2book.module
```diff
@@ -536,29 +540,37 @@ function ding_place2book_get_event_state($event_maker_id, $event_id) {
             'end' => strtotime($price->sale_end_at),
           ];
         }
+
         // Some events, for example those migrated from the old API, will have
         // sales period on the event and not on the individual tickets.
-        if (isset($event->sale_open_at) && isset($event->sale_close_at)) {
+        if (isset($event->sale_open_at) && isset($event->sale_close_at) && !$sale_periods) {
```
holt83: I'm wondering about the `!$sale_periods` condition. 

I don't think it's needed and we don't know if an event can both have sale period on the event while also having sale periods on individual tickets. In any case, I think it will best to just remove it.

holt83:
After taking a closer look at the code, I'm not sure it will work.

What you do now is you iterate all sales periods and find the min-period. Then you check if $now is before the min-period. But what if the sales-period with the min_period has already passed? You can not know that, and they will be present on the event anyway.

I would prefer if you would not destroy the original loop, because it's a bit complicated and we will probably mess something up. The code is tested and tried in production so no need to mess with that too much IMO.

My suggestion is that each time you find an upcoming sales period you remember that as the next one. If you find another one and that one has sales_begin **lower** that the previous one you remember that instead. Finally, after iterating all sales periods you should know whether there's any upcoming sales period and the begin date of the closest one.

themes/ddbasic/sass/components/node/event.scss
```diff
@@ -257,10 +257,18 @@
...
             &:hover {
               background-color: $green-text;
-              color: lighten($black, 15);
+              color: lighten($white, 15);
+            }
+
+            a {
+              &:link, &:visited, &:hover, &:active {
+                color: $white;
+                text-decoration: underline;
```
kasperg: I cannot seem to grok this change. It targets generic links inside `.tickets-available` but that class only seems to be set directly on links on L590. Am I missing something?

modules/ding_place2book/ding_place2book.module
```diff
@@ -545,20 +550,27 @@ function ding_place2book_get_event_state($event_maker_id, $event_id) {
           ];
         }
 
+        // Find the closest sale period.
+        $begin_values = array_column($sale_periods, 'begin');
+        $min_period_begin = min($begin_values);
+        $min_period = array_filter($sale_periods, function ($sale_period) use ($min_period_begin) {
+          return ($sale_period['begin']) == $min_period_begin;
+        });
+        $min_period = array_shift($min_period);
```
kasperg: :+1:

modules/ding_place2book/ding_place2book.module
```diff
@@ -545,20 +550,27 @@ function ding_place2book_get_event_state($event_maker_id, $event_id) {
           ];
         }
 
+        // Find the closest sale period.
+        $begin_values = array_column($sale_periods, 'begin');
+        $min_period_begin = min($begin_values);
```
kasperg: This selects the lowest/first sales period. Is there a risk that one sales period is in the past (and thus over) while there may be another current open one?

Example:
```php
$sales_periods = [
  [ 'begin' => 1554076800, 'end' => 1556582400 ], // <-- this should be selected through April 2019
  [ 'begin' => 0, 'end' => 1 ], // <-- this would be selected
];

File: themes/ddbasic/sass/components/node/event.scss (hunk 3 above)
Comment (guddo) [reply themes/ddbasic/sass/components/node/event.scss]: ticket-available is set on L590 @ modules/ding_place2book/ding_place2book.module  and L309 @ themes/ddbasic/template.node.php, right. What do you think is wrong?

File: modules/ding_place2book/ding_place2book.module (hunk 5 above)
Comment (holt83) [reply modules/ding_place2book/ding_place2book.module]: Yes, there is chance for this to happen.

I would also like to note that if we find an active sales periods there's really no need to continue with all this. The status will just be sales open. Hence my suggestion to keep the original loop, which handled this pretty well IMO.

There was a lot of gotchas when i wrote the code. I'm not sure it's a good idea to change it too much.

File: modules/ding_place2book/ding_place2book.module [hunk 6]
```diff
@@ -546,19 +551,31 @@ function ding_place2book_get_event_state($event_maker_id, $event_id) {
...
-            $ticket_link_status = 'open';
-            break;
-          }
-          elseif ($now < $sale_period['begin']) {
-            // We haven't found any open sales period yet,q but at least there's
-            // an upcoming, so set that status for now.
-            $ticket_link_status = 'upcoming';
-          }
+
+        // Filter out periods which are already over.
```
bozeak: @kasperg Related to https://github.com/ding2/ding2/pull/1405/files#r278920822. Added additional filtration, which will wipeout already passed periods, so we will have only current and future periods.
Does it make sense?

themes/ddbasic/sass/components/node/event.scss
kasperg: @guddo As far as I can tell the change above will result in the following slightly abbreviated selectors:

```css
.button.ticket-available a:link { color: $white; text-decoration: underline; }
``` 

This targets the following HTML element:

```html
<div class="parent">
  <div class="button ticket-available">
    <a href="http://domain.com/">...</a>
  </div>
</div>
```

All the places where the `ticket-available` class is used it seems to be used directly on the link e.g.

```php
$class = 'ticket-available';
$data = l(t('Book a ticket'), $event->links->sales_location, array(
  'attributes' => array(
    'target' => '_blank',
    'class' => array('button', $class),
  ),
));
```

I assume this results in the following markup which is not targeted by the above selectors.

```html
<a class="button ticket-available" href="http://domain.com/">...</a>
```

If that is correct then the change above does not seem to do anything. I am not sure whether it needs to be updated or removed.

modules/ding_place2book/ding_place2book.module
kasperg: Yep. Thanks for elaborating.

modules/ding_place2book/ding_place2book.module
```diff
@@ -681,19 +686,31 @@ function ding_place2book_get_event_state($event_maker_id, $event_id) {
...
+
+        // Find the closest sale period.
+        $begin_values = array_column($sale_periods, 'begin');
+        $min_period_begin = min($begin_values);
+        $min_period = array_filter($sale_periods, function ($sale_period) use ($min_period_begin) {
+          return ($sale_period['begin']) == $min_period_begin;
+        });
+        $min_period = array_shift($min_period);
+
+        if ($now >= $min_period['begin'] && $now <= $min_period['end']) {
```
holt83: How will this behave if:

1. All sales periods are over and there are no $min_period?
2. The event is published and non-passive, but somehow there's no sales periods on the ticket and the event and hence also no $min_peroid.

Also a bit nitpicking here: wouldn't it be more simple to just check `$now < $min_period['begin']` first and then 'upcoming' otherwise just 'open', since you have already verified that all the sales periods left are not over?

bozeak: Checked one more time.

> All sales periods are over.

If such situation happens when the status will be set to "event-over" or "closed".

> The event is published and non-passive, but somehow there's no sales periods on the ticket.

In such case the request to P2B will not return anything and this simply will not proceed, because once the module is enabled and configured - it will force the authors to define at least one price even the event is free.

holt83: > If such situation happens when the status will be set to "event-over" or "closed".

No it won't. Just because all the sales periods are over doesn't mean the event is over or complete. You will get a notice in these cases,

> In such case the request to P2B will not return anything and this simply will not proceed, Once the module is enabled and configured - it will force the authors to define at least one price even the event is free.

I'm not sure what you mean it will not return anything(?), but it's possible to delete a price in place2book GUI after creation and in these situations you will also get a notice.

themes/ddbasic/sass/components/node/event.scss
guddo: You are on to something here. It's quite hard for us to test now that we don't have P2b credentials and events with various states anymore. @holt83 do you have a running environment and able to check it out?

holt83: Yes, I will have a look at it tomorrow.

I was planning of doing a thorough test of all the states again after these changes anyway :)

holt83: @guddo, @kasperg Kasper is right: the selector does not apply, since the classes is added directly to the anchor tag. 

The text color IS white, because you already set that [further up](https://github.com/ding2/ding2/pull/1405/files#diff-bbdbc25e57d7a11faa06cf1676d62b5bR265), but the text is not underlined.

modules/ding_place2book/ding_place2book.module
holt83: @bozeak I just tested this: if all sales_periods are over active **is not false** and $min_period will be FALSE in these cases.

holt83: So event though the documentation states that active takes sales periods into consideration it's not doing that for the ticket types sales periods. Hence the event is still active event though there are no active sales periods. In our case active just tells us whether the event is paused/unpaused.

Even though it does create a notice it seems that the ticket-status will be correctly set to closed anyway.

guddo:
@holt83 Will you make changes to this PR or should we? If we should; can you elaborate what we need to change?

holt83:
@guddo Even though the code appears to output the correct status anyway, I don't think it's a good idea to rely on that. 

It's probably best to verify that there's in fact any sales periods that are not over before trying to find the min period.

modules/ding_place2book/ding_place2book.module
holt83: Even though the code appears to output the correct status anyway, I don't think it's a good idea to rely on that.

It's probably best to verify that there's in fact any sales periods that are not over before trying to find the min period.

holt83:
Thank for the updates... Is this ready for another review by the core team?

guddo:
It should be, yes.

holt83:
Ok, I'll update the case.

kasperg:
:+1: LGTM.

PR #1407, opened 2019-04-12
4267: Fix offset for user menu in header

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/4267

#### Description

Fixes wrong offset for user menu in header. Screenshot of the problem in the issue above.

Also fixes wrong offset of the menu when admin menu with or without shortcuts are used. In these cases we don't need the extra height, since admin_menu itself adds an offset.

#### Screenshot of the result

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1410, opened 2019-04-25
4261: Fix renew all loans button

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/4261

#### Description

Something has changed so that using a click handler for these checkboxes
breaks the functionality of the select all renew checkbox and therefore
also the global select all renew button (which just reuses the select
all checkbox).

The problem was the triggering of click handler on the local select all
checkboxes, from the change event handler on the global select all. This
would simuluate a real click changing state of the checkbox and mess up
the logic in the click handler. For example, if we have just set the box status 
to checked with the prop method, it would immediately change it back in 
browser click handler, before invoking our change handler.

I don't know why this hasn't been a problem before, but I suspect the jquery
update have caused our logic to behave a bit different.

The easiest fix seem to be to use a change handler instead on the local
select all checkboxes. This will not break any functionality since a
real click from a user will also fire the change handler. And now we can
trigger the change handler instead, which doesn't change the state of
the checkbox. It also seems more consistent since a change handler is
alos used on the other checkboxes.

#### Screenshot of the result

#### Checklist

### DIFF
--- a/modules/ding_loan/js/ding_loan.js
+++ b/modules/ding_loan/js/ding_loan.js
@@ -16,7 +16,7 @@
     $('.action-buttons input[type=submit]').prop('disabled', 'disabled');
 
     // Handle select all checkboxes.
-    var sub_select_all = $('.select-all input[type=checkbox]').click(function() {
+    var sub_select_all = $('.select-all input[type=checkbox]').change(function() {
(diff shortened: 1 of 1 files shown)


JrgenGNielsen:
reviewed and approved

PR #1411, opened 2019-04-26
4286: Fix event price display

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/4286

#### Description

Fix the display of event price to work with the price field after it has been converted to a text field.

- If price is empty show "Free" text
- Only show currency if price is numeric.

#### Screenshot of the result

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1412, opened 2019-04-29
4173 by Inlead: Re-export Section panel template.

inlead:
#### Link to issue

#### Description

Re-exported the default Section panel variant with the new required panel as per defined in the issue.

#### Screenshot of the result

image

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1413, opened 2019-04-29
1340: Fix kultunaut image export

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/1340

#### Description

Ensure list image from the event is exported to `kultunaut_picture_uri`. (see: http://developer.place2book.com/v1/events/)

Also updates the image style used to be better quality than (120px/120px), by using the same crop and scale dimensions, that the list image uses in CMS on frontpage.

I think scaling width to 1200px is too much, when looking at other images and how they are displayed on kultunaut. 548x308 is also a lot better than 120x120, so I think this will be sufficient.

### Screenshots of the result

1340-successful-kultunaut-image-export

#### Checklist

I removed the feature export of user permission, since these permissions are already correctly managed by secure_permissions module via hook implementation in ding_permission.


guddo:
Does it make sense to rename this image style as it is not `_square` anymore?

holt83:
> Does it make sense to rename this image style as it is not _square anymore?

I think it does, but then we would still need to keep the old image style to ensure already exported URLs still work? Or do some redirecting to the new URLs. I think maybe it's not worth it?

guddo:
I would keep the `_square` and change it to 308x308 (or whatever fits the default list) and then add your new style as an addition instead.

holt83:
> I would keep the _square and change it to 308x308 (or whatever fits the default list) and then add your new style as an addition instead.

Done and done! :)

guddo:
A note about why you are removing the user permission would be nice, as it seems unrelated to the task itself!

holt83:
It's in commit message, but I can post here as well:

"These permissions are already correctly managed by secure_permissions
module via hook implementation in ding_permission."

JrgenGNielsen:
reviewed and approved

PR #1414, opened 2019-04-30
4207: Update EU Cookie Compliance module to 1.28

kasperg:
#### Link to issue

https://platform.dandigbib.org/issues/4207

#### Description

This addresses a cross site scripting security issue:
https://www.drupal.org/sa-contrib-2019-033

The update is a major leap in versions where the functionality of the
module has been extended to take GDPR and other issues into account.

We do not have to do anything special regarding consent here so we
just need to configure the module to behave as it did previously.
With the new configuration options this boils down to:

1. Using the default consent method
2. Not showing any disagreement button

The updated module uses `I18n_variable` to store internationalized
settings. To provide a consistent configuration we also use this
module for new installs. Instead of defining all configuration
options we modify the defaults to work as we want to. This allows us
to highlight what is important in our case.

Our custom template has been updated to both include the close button
and be in line with the latest changes to the module.

Finally we have to explicitly specify paragraphs in our cookie text
to be in block format. For some reason the updated module thinks
they should be inline.

In the current form the button does nothing - probably because of a
race condition. The behavior was called before the popup was created.

To fix this we instead listen for an event emitted by the module and
register our click handler then.

Click logic is replicated from 
`Drupal.eu_cookie_compliance.declineAction` to ensure that the close
animation is similar to the open animation. Before it was too fast.

#### Screenshot of the result

#### Checklist

The update only required an update to version 1.26 but since this is a major leap in versions anyhow and [1.26 had a bug which affected us](https://www.drupal.org/project/eu_cookie_compliance/issues/3038087) I chose to upgrade to the latest version.

Regarding Scrutinizer failures:

- [Comments in `ding2.install`](https://scrutinizer-ci.com/g/ding2/ding2/inspections/c34c7f5d-5591-4d35-a5ac-4a12d913d331/issues/files/ding2.install?status=new&orderField=path&order=asc&honorSelectedPaths=0) should be disregarded as it is related to code which has not been changed as a part of this PR
- I think [comments in `eu-cookie-compliance-popup-info.tpl.php`](https://scrutinizer-ci.com/g/ding2/ding2/inspections/c34c7f5d-5591-4d35-a5ac-4a12d913d331/issues/files/themes/ddbasic/templates/eu-cookie-compliance-popup-info.tpl.php?status=new&orderField=path&order=asc&honorSelectedPaths=0) should be disregarded. The code follows standard procedures for template files.

### DIFF
--- a/ding2.info
+++ b/ding2.info
@@ -21,6 +21,7 @@ dependencies[] = field_ui
 ; Helper modules.
 dependencies[] = module_filter
 dependencies[] = menu_position
+dependencies[] = i18n_variable
 dependencies[] = eu_cookie_compliance
 dependencies[] = view_unpublished
 dependencies[] = redirect
--- a/ding2.install
+++ b/ding2.install
@@ -199,6 +199,12 @@ function ding2_update_dependencies() {
     // provider is present.
     'ting' => 7014,
   );
+  // Update hook which sets custom configuration for EU Cookie Compliance.
+  $dependencies['ding2'][7064] = array(
+    // Update hook that updates the configuration format for EU Cookie
+    // Compliance module.
+    'eu_cookie_compliance' => 7010,
+  );
 
   return $dependencies;
 }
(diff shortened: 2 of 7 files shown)


JrgenGNielsen:
reviewed and approved

PR #1415, opened 2019-04-30
4237: Clean up in infomedia error handling

cableman:
#### Link to issue

https://platform.dandigbib.org/issues/4237

#### Description

Infomedia throws exception as it tries to login to FBS without pin-code which is not available (and not at all available when using singe-sign-on services). Code removed was from 2011.

Also fixed error handling that was wrong (sorry for that one) which now is handled by exceptions. 

#### Screenshot of the result

Screenshot 2019-04-30 at 10 58 41

#### Checklist

No comments.


kasperg:
I assume that the Behat test failures are not related to the change at hand.

cableman:
You are right the behat work on next to last commit, but failed at other the last commit. Tried to re-run but it's not related to this PR.

PR #1418, opened 2019-05-06
4295: Disable ratings on search results

kasperg:
#### Link to issue

https://platform.dandigbib.org/issues/4214

#### Description

Disable ratings on search results

The current implementation loads ratings synchronously for each
result. This increases response times considerably.

To address this we hide the ratings field on search results. This is
also in line with UX recommendations.

If we want to reintroduce ratings they should be loaded
asynchronously instead.

This change is the result of hiding the ratings field group and 
fields on the search result view mode for ting objects and doing a 
features export.

Field instances and definitions are stored in the database.
Consequently we need to do an explicit feature revert for our changes
to take effect.

In the process we also have to fix programmatic feature reverting.
In the current form the feature object would be passed to
module_exists() which expects a string. If not it will return false
and features will not be reverted.

To address this we separate the feature name and the feature object
into separate variables.

#### Screenshot of the result

Search Ting | danmark | Ding2 2019-05-06 14-50-28

#### Checklist

### DIFF
--- a/ding2.install
+++ b/ding2.install
@@ -726,3 +726,11 @@ function ding2_update_7063() {
     'ding_nodelist',
   ));
 }
+
+/**
+ * Revert Ding Entity Rating feature to hide ratings on search results.
+ */
+function ding2_update_7065() {
+  require_once drupal_get_path('profile', 'ding2') . '/ding2.install_callbacks.inc';
+  _ding2_features_revert('ding_entity_rating');
+}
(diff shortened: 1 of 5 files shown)


JrgenGNielsen:
Reviewed and approved. 
Aber dabei: behat_tests fails. But seems unrelated to this issue.

kasperg:
Closed in favour of #1421.

PR #1419, opened 2019-05-08
4095: Better accessability color contrast

rasben:
#### Link to issue

https://platform.dandigbib.org/issues/4095

#### Description

A lot of the buttons on the website have poor color contrast, with very light font colors used on light backgrounds and vice versa

This looks to fixes this, whilst trying not to introduce any new colors.

#### Screenshot of the result

Basically affects all buttons on the website, but if I have done my job correctly you wont notice the difference unless you compare directly.

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1420, opened 2019-05-08
4091: Skip to content link.

rasben:
#### Link to issue

https://platform.dandigbib.org/issues/4091

#### Description

This is a visually hidden link which skips to the main content of the page.
It's placed at the top of the page, and it's useful for screenreader-users as a way of skipping past any base layout, such as the header.
The styling which hides it visually is based on the same way that core themes usually do it.

#### Checklist


rasben:
Scrutinizer fails, but it is unrelated failures. It's because I am editing the ding2-site-template.tpl.php file and then old issues pops up.

JrgenGNielsen:
reviewed and approved

PR #1426, opened 2019-05-10
4100 carousel h2 to h3 take two

xendk:
#### Link to issue

https://platform.dandigbib.org/issues/4100

#### Description

From #1361:

> For WCAG reasons, the elements within a carousel should use `<h3>` instead of `<h2>` for their titles.

This does it by extending the field formatter with configuration for which HTML element to use, and configure the teaser display to use it.

Supersedes #1361

#### Checklist


xendk:
Scrutinizer error is existing problem.

JrgenGNielsen:
reviewed and approved

PR #1428, opened 2019-05-10
4093: Better general focus styling on the site.

rasben:
#### Link to issue

https://platform.dandigbib.org/issues/4093

#### Description

Generally just better styling for focus, such as keyboard focus.

#### Checklist


JrgenGNielsen:
behat error is unrelated to this issue.
reviewed and approved

PR #1430, opened 2019-05-14
4329: Fixed pincode description float

b14sha:
#### Link to issue

https://platform.dandigbib.org/issues/4329

#### Description

The form description element needed a width and float to be correctly displayed

#### Screenshot of the result

Screenshot 2019-05-14 at 14 51 04

### DIFF
--- a/themes/ddbasic/sass/components/form/form-user-profile.scss
+++ b/themes/ddbasic/sass/components/form/form-user-profile.scss
@@ -91,7 +91,7 @@
         clear: left;
       }
 
-      #edit-pass--2 {
+      fieldset[id^='edit-pass'] {
         width: 100%;
         float: left;
         padding: 20px 0;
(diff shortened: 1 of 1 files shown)


themes/ddbasic/sass/components/form/form-user-profile.scss
```diff
@@ -91,7 +91,7 @@
         clear: left;
       }
 
-      #edit-pass--2 {
+      fieldset[id^='edit-pass'] {
```
holt83: Isn't it possible to just have `[id^='edit-pass']` without the fieldset?

JrgenGNielsen: True. Personally, though, I like knowing what kind of element we're looking for...

JrgenGNielsen:
reviewed and approved

themes/ddbasic/sass/components/form/form-user-profile.scss
holt83: Good point :)

PR #1431, opened 2019-05-15
4332: Align search results

b14sha:
#### Link to issue

https://platform.dandigbib.org/issues/4332

#### Description

Adjusted the sort select element on search-results to align with top of ting-search-backend-engines-form.

Also changed the font weight to use the correct typography.

#### Screenshot of the result

Screenshot 2019-05-15 at 11 42 05

#### Comments
The commit fail on ci/circleci: behat_tests, but since the PR only includes small changes to 2 scss files - i dont see why this PR should fail?


JrgenGNielsen:
reviewed and approved

PR #1433, opened 2019-05-15
4279 by Inlead: Do not make menu default when creating Section.

bozeak:
#### Link to issue

https://platform.dandigbib.org/issues/4279

#### Description

By default on Section creation there is an option to create menu item in main menu. As it was required, this commit disables this action so by default checkbox will be unchecked.

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1436, opened 2019-05-20
4168 by Inlead: Add Eresource CT as supported Nodelist CT.

bozeak:
#### Link to issue

https://platform.dandigbib.org/issues/4168

#### Description

Added support of Eresources CT content into nodelist module. This commit contains changes for clean installs/update for enabling Eresources support and templates for each type of widgets present in current distribution.

#### Checklist


JrgenGNielsen:
could you fix the Scrutinizer errors?

guddo:
Fixed. The rest of the mentions does not seem to be related to this changeset.

JrgenGNielsen:
reviewed and approved

PR #1439, opened 2019-05-23
4363: Updated Drupal makefile

cableman:
#### Link to issue

https://platform.dandigbib.org/issues/4363

#### Description

Removed not needed patches and updated core version. Also changed from using ssh to https to enable clone inside docker container on non Linux boxes.

#### Screenshot of the result

Non

#### Checklist

Non

### DIFF
--- a/drupal.make
+++ b/drupal.make
@@ -3,17 +3,15 @@ api = 2
 
 ; Core
 projects[drupal][type] = core
-projects[drupal][version] = 7.60
-projects[drupal][patch][] = "http://drupal.org/files/issues/menu-get-item-rebuild-1232346-45.patch"
-projects[drupal][patch][] = "http://drupal.org/files/ssl-socket-transports-1879970-13.patch"
-projects[drupal][patch][] = "http://www.drupal.org/files/issues/1232416-autocomplete-for-drupal7x53.patch"
-projects[drupal][patch][] = "http://drupal.org/files/issues/translate_role_names-2205581-1.patch"
-projects[drupal][patch][] = "http://www.drupal.org/files/issues/drupal-tabledrag-scroll-2843240-36.patch"
-projects[drupal][patch][] = "patches/drupal_core.robots.txt.ding2.patch"
+projects[drupal][version] = 7.67
+projects[drupal][patch][] = "https://drupal.org/files/issues/menu-get-item-rebuild-1232346-45.patch"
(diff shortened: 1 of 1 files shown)


holt83:
> Also changed from using ssh to https to enable clone inside docker container on non Linux boxes.

Just out of curiosity: can you elaborate on this?

What do you mean by "docker container on non Linux boxes"? Is it when you use "Docker on windows" with Hyper-V or something? Or is it Windows containers?

cableman:
When using docker on mac/windows you can't forward you ssh agent into docker (which you can in Linux). So when running make inside docker:

`docker run -v ${PWD}:/app itkdev/drush6:latest make --concurrency=1 --contrib-destination=profiles/ding2/ --working-copy drupal.make htdocs2`

It fails saying that it can't clone `git@github.com:ding2/ding2.git` but it works over https with `https://github.com/ding2/ding2.git`.

Also github recommend using https: https://help.github.com/en/articles/which-remote-url-should-i-use

holt83:
Ok, thanks :) 

They used to say "We strongly recommend using an SSH connection when interacting with GitHub". See: https://stackoverflow.com/questions/11041729/why-does-github-recommend-https-over-ssh

But I can see why using https is better.

A sidenote: I believe you could get around this on Windows by using docker client in WSL and have it connect to a remote docker daemon in a VM or "Docker for windows" exposed on a local port.

drupal.make
```diff
@@ -3,17 +3,15 @@ api = 2
 
 ; Core
 projects[drupal][type] = core
-projects[drupal][version] = 7.60
-projects[drupal][patch][] = "http://drupal.org/files/issues/menu-get-item-rebuild-1232346-45.patch"
-projects[drupal][patch][] = "http://drupal.org/files/ssl-socket-transports-1879970-13.patch"
```
kasperg: This patch has been removed but as far as I can tell the corresponding issue has not been resolved. Is this intentional?

cableman: It was an patch I crated back in the time to fix issues with certificates and ALMA. But it have been resolved in other ways and is not an issue for the current setup of DDB CMS services.

GitCop:

 * Commit: edeb88d2071cc30e19cbbcf14c0fbfc23ff7bd14

---

GitCop:

 * Commit: edeb88d2071cc30e19cbbcf14c0fbfc23ff7bd14

---

PR #1442, opened 2019-05-29
4094: Langcode on node content.

rasben:
#### Link to issue

https://platform.dandigbib.org/issues/4094

#### Description

Currently, some libraries have english pages.
The issue is that these pages, if read by a screen readers, or index by a search engine, will be treated as danish content, as the only `lang` property on the page is on `<html lang="da"`

Looking at the page, this is actually not the problem -  the content around the nodes are actually in danish, such as the menu and footer, so really we only want a `lang="da"` on the actual content.

This PR:
- Adds a selector to the node form, where the editor can choose a language:
Screenshot 2019-05-29 at 14 22 09

- Based on the chosen language, it sets a `lang="[LANGCODE]"` on the relevant node content wrapper.

#### Checklist

### DIFF
--- a/themes/ddbasic/template.node.php
+++ b/themes/ddbasic/template.node.php
@@ -11,7 +11,22 @@
  * Override or insert variables into the node templates.
  */
 function ddbasic_preprocess_node(&$variables, $hook) {
-  //
+  $installed_languages = language_list();
+
+  // If the node has a language associated, and its part of our installed
+  // languages (The value can also be LANGUAGE_NONE), then we will add the
+  // lang="LANGCODE" attribute to the attributes/content_attributes.
+  // This way, if the screenreader reads it, it will read the content in the
+  // correct prounication.
+  if (!empty($variables['language']) &&
+      !empty($installed_languages[$variables['language']])) {
+    // We're doing it to both attributes and content_attributes as it seems
+    // fairly random whether or not a template uses both or not.
+    // Even if a template does use both, it does no damage.
+    $variables['attributes_array']['lang'] = 'en';
+    $variables['content_attributes_array']['lang'] = 'en';
+  }
+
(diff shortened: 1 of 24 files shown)


rasben:
I've already fixed a bunch of scrutinizer errors, **but the remaining errors are unrelated to my changes** - and I've tried to fix them in the past too which just makes scrutinizer more angry..

JrgenGNielsen:
reviewed and approved

themes/ddbasic/template.node.php
```diff
@@ -11,7 +11,22 @@
...
+  // lang="LANGCODE" attribute to the attributes/content_attributes.
+  // This way, if the screenreader reads it, it will read the content in the
+  // correct prounication.
+  if (!empty($variables['language']) &&
+      !empty($installed_languages[$variables['language']])) {
+    // We're doing it to both attributes and content_attributes as it seems
+    // fairly random whether or not a template uses both or not.
+    // Even if a template does use both, it does no damage.
+    $variables['attributes_array']['lang'] = 'en';
+    $variables['content_attributes_array']['lang'] = 'en';
```
holt83: @rasben Shouldn't this be:

```
$variables['attributes_array']['lang'] = $variables['language'];
$variables['content_attributes_array']['lang'] = $variables['language'];
```

rasben: ?! yes you are totally right @holt83, this must have been a test value that i forgot to revert
I'll create a PR straight away and test if it still works

rasben: @holt83 https://github.com/ding2/ding2/pull/1477

holt83: @rasben thanks for the fast response and fix 👍

PR #1443, opened 2019-05-29
4322: Do not display async empty campaign plus pane wrapper

kasperg:
#### Link to issue

https://platform.dandigbib.org/issues/4322

#### Description

The default panel pane markup surrounding campaign plus panel panes
has styling. It will be rendered when using asynchronous loading
but only the content of the pane is replaced with actual content. If 
there are no campaigns to be shown in the panel pane this means that 
the surrounding markup remains. As it is styled with a padding and 
background color this leaves an empty grey box.

To avoid this we hide the entire wrapper by default and only show
it if any actual campaign markup has been loaded.

#### Screenshot of the result

screencast 2019-05-29 14-58-32

#### Checklist

### DIFF
--- a/modules/ding_campaign_plus/js/ding_campaign_plus.js
+++ b/modules/ding_campaign_plus/js/ding_campaign_plus.js
@@ -13,10 +13,18 @@
     attach: function (context) {
       $('[data-ding-campaign-plus-cid]', context).once(function () {
         var campaign = $(this);
+        // Hide the wrapper by default. We do not know if we have any actual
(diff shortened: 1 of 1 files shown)


holt83:
👍 for this! 

Was actually just battling with this myself. Also had the problem on frontpage, which was even more annoying. And with Varnish/page cache these grey areas will just stick around.

Just tested and it fixed both my cases.

JrgenGNielsen:
reviewed and approved

PR #1447, opened 2019-06-03
4331: Campaign plus basic search term fixes

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/4331

#### Description

This PR contains two fixes to the matching in campaign plus basic search term trigger, which both prevent it from working it all.

1. There was a typo on the sub-type of basic campaign. For search term the key is `search_term` not `search_request`. This typo prevented the code from fetching the correct campaigns in the DB. Fixed in: 80f88146ca9536eb59091adb4c063d17c0cecd36

2. The mathing of the user entered search term and the matching value in the DB for the campaign was incorrect/the other way around. Instead of taking the comma-separated value from the DB and splitting into terms it took the user entered search. As a consequence, campaigns with several comma-separated terms was not working and also a single search term with several words was not working. Fixed in 141c81a3e6c243fca45bedad43a1f5c141ccf8eb

Also allow administrative users to use the special "\<front\>"-value for basic path trigger, to target frontpage.

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1452, opened 2019-06-06
4379: Catch exception when checking reservability

holt83:
#### Link to issue

#### Description

An exception can be thrown if provider is down or if its not
configureed correctly. Since we call fbs_service directly, we
need to handle exception ourselves.

The reservability check should always be done asynchronously with AJAX and therefore should not be able to take the whole frontpage down. I havn't had time to investigate why this happens, but I can confirm this fixes the WSOD on frontpage, when issues with FBS-provider.

#### Screenshot of the result

#### Checklist

### DIFF
--- a/modules/fbs/fbs.module
+++ b/modules/fbs/fbs.module
@@ -99,10 +99,17 @@ function fbs_ding_entity_is($object, $class) {
       // display and has a chance of calling opensearch to get marc record. We
       // need none of this and only want to know if the material is reservable
       // and if it has any periodical information.
-      $result = fbs_service()->Catalog->getHoldings(
-        fbs_service()->agencyId,
-        $object->localId
-      );
+      try {
+        $result = fbs_service()->Catalog->getHoldings(
+          fbs_service()->agencyId,
+          $object->localId
+        );
+      }
+      catch(Exception $e) {
+	watchdog_exception('fbs', $e);
+        return FALSE;
+      }
+
       $result = reset($result);
       $holdings = reset($result->holdings);
 


modules/fbs/fbs.module
```diff
@@ -99,10 +99,17 @@ function fbs_ding_entity_is($object, $class) {
...
-        fbs_service()->agencyId,
-        $object->localId
-      );
+      try {
+        $result = fbs_service()->Catalog->getHoldings(
+          fbs_service()->agencyId,
+          $object->localId
+        );
+      }
+      catch(Exception $e) {
```
kasperg:
```suggestion
      catch (Exception $e) {
```

modules/fbs/fbs.module
```diff
@@ -99,10 +99,17 @@ function fbs_ding_entity_is($object, $class) {
...
+	watchdog_exception('fbs', $e);
```
kasperg: Indentation seems off here?

modules/fbs/fbs.module
```diff
@@ -99,10 +99,17 @@ function fbs_ding_entity_is($object, $class) {
...
+        return FALSE;
```
kasperg: Should we return `NULL` here since we actually cannot tell anything about the reservability of a material when FBS returns an error?

[`is()` will return `FALSE` if no hook returns `TRUE`](https://github.com/ding2/ding2/blob/c0273fbf051044e9ac6971cad104df70cc4c7786/modules/ding_entity/ding_entity.module#L372-L388).

PR #1454, opened 2019-06-07
4256: Fix missing tabrolls

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/4256

#### Description

The tabrolls are missing because of a bug that appears in a very specific setup with a non-required entity reference field used as a non-required relationship on a view:

https://www.drupal.org/project/drupal/issues/1349080

The reason why this error hasn't surfaced before is that the node_access table hasn't been rebuild after enabling view_unpublished and therefore still contains the global view all access record that the node module adds on install. On those sites where it's suddenly appearing something must have triggered a rebuild.

There's a patch provided for Drupal 7 in the issue, but it's a bit extensive and has some test fails. When reading through the comments in above issue, it doesn't seem like the Drupal 7 patches are tested and "approved" by the other participants.

I therefore think our best cause of action right now, is to just disable the SQL-rewriting for this specific problematic view. I list some reasons for this approach in this comment: https://platform.dandigbib.org/issues/4256#note-25 

#### Screenshot of the result

#### Checklist

We could consider triggering a node_access_rebuild during next release in an update function, to ensure node_access table is consistent on all tables.

I'm really not sure what, if any, problems would surface/occure if doing this/not doing this.

### DIFF
--- a/modules/ding_tabroll/ding_tabroll.views_default.inc
+++ b/modules/ding_tabroll/ding_tabroll.views_default.inc
@@ -28,7 +28,7 @@ function ding_tabroll_views_default_views() {
   $handler->display->display_options['access']['type'] = 'none';
   $handler->display->display_options['cache']['type'] = 'none';
   $handler->display->display_options['query']['type'] = 'views_query';
-  $handler->display->display_options['query']['options']['query_comment'] = FALSE;
+  $handler->display->display_options['query']['options']['disable_sql_rewrite'] = TRUE;
   $handler->display->display_options['exposed_form']['type'] = 'basic';
   $handler->display->display_options['exposed_form']['options']['reset_button_label'] = 'Revert';
   $handler->display->display_options['pager']['type'] = 'some';


modules/ding_tabroll/ding_tabroll.views_default.inc
```diff
@@ -28,7 +28,7 @@ function ding_tabroll_views_default_views() {
   $handler->display->display_options['access']['type'] = 'none';
   $handler->display->display_options['cache']['type'] = 'none';
   $handler->display->display_options['query']['type'] = 'views_query';
-  $handler->display->display_options['query']['options']['query_comment'] = FALSE;
```
holt83: This may seem unrelated, but this is how the view is exported with `drush fu`.

holt83:
@kasperg apologize for not responding in time (vacation)

Should we roll with the patch instead and/or should I add the commit messages in a new PR?

PR #1459, opened 2019-06-13
3265: Date slider

BirkAndMe:
#### Link to issue

#### Description

Year span on search.

#### Checklist


kasperg:
Closed because no good solution could be provided without resulting in extra search requests.

PR #1460, opened 2019-06-13
1460: Reservation buttons overhaul Round 2

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/1460

Continuation of the work in https://github.com/ding2/ding2/pull/1139

#### Description

This PR contains changes requested in previous PR and some small corrections:

1. Optimization of reservability checks. Now checks multiple materials at once - both in frontend and backend.
2. Do not check reservability for periodical reserve buttons to avoid making unnecessary requests.
3. Handle FBS errors when checking reservability to prevent fatal PHP error. 

#### Screenshot of the result

#### Checklist

### DIFF
--- a/modules/ding_entity/ding_entity.api.php
+++ b/modules/ding_entity/ding_entity.api.php
@@ -63,5 +63,54 @@ function hook_ding_entity_buttons($type, $entity, $view_mode, $widget) {
 }
 
 /**
+ * Help determine whether entities belongs to given pseudo-classes.
+ *
+ * Note that there's also a single entity version of this hook named
+ * hook_ding_entity_is(), which is invoked from within a ding entity object and
+ * therefore only support one entity at a time. This hook is therefore better to
+ * implement if acting on multiple entities and the implementation can achieve
+ * better performance by recieving all entities at once.
+ *
+ * @param array $entities
+ *   The entities that needs to be checked against the pseudo-class.
+ * @param string $class
+ *   The pseudo class to check for.
+ *
+ * @return array
+ *   The implementation must return an array with an entry for each of the
+ *   passed entities keyed by their ding_entity_id. If the implementation
+ *   doesn't have a say for a particular entity and/or pseudo-class it must set
+ *   NULL to return something for that entity and signal that it is neutral.
+ */
+function hook_ding_entity_entities_is(array $entities, $class) {
+  $return = [];
+  // Loop through each passed entity and check if it belongs to $class.
+  foreach ($entities as $entity) {
+    // Here we check if the entities belong to a pseudo class called 'book' to
+    // give an example, but it could be anything. This way the implementation
+    // can "invent" and add new pseudo-classes just by implementing this hook.
+    if ($class == 'book') {
+      // If we are certain this entity belongs to the pseudo class set TRUE:
+      if ($entity->getType() == 'book') {
+        $return[$entity->ding_entity_id] = TRUE;
+      }
+      // If we are absolutely certain this entity does not belong to the class and
+      // wanna overwrite any module who says so, we set FALSE for this entity:
+      elseif (strpos($entity->getType(), 'book') === FALSE) {
+        $return[$entity->ding_entity_id] = FALSE;
+      }
+    }
+    if (!isset($return[$entity->ding_entity_id])) {
+      // If we don't know about the pseudo class or don't have sufficient
+      // information about the entity to have a say, we can set a neutral response
+      // by setting NULL for this entity. This way we don't interfere with the
+      // final outcome and leaves more flexibility to other modules.
+      $return[$entity->ding_entity_id] = NULL;
+    }
+  }
+  return $return;
+}
+
+/**
  * @} End of "addtogroup hooks".
  */
--- a/modules/ding_entity/ding_entity.api.php
+++ b/modules/ding_entity/ding_entity.api.php
@@ -57,8 +57,15 @@ function hook_ding_entity_collection_view(&$object, $view_mode) {
  * Return an array of buttons to add.
  */
 function hook_ding_entity_buttons($type, $entity, $view_mode, $widget) {
-  if ($type == 'ding_entity' && $entity->is('reservable')) {
-    return array(ding_provider_get_form('ding_reservation_reserve_form', new DingReservationReservableEntity($entity), TRUE));
+  // Adds resevation buttons is entity is resevable.
+  if ($type == 'ding_entity' && $entity->is('library_material')) {
+    if (ding_provider_implements('reservation', 'is_reservable')) {
+      $reservable = ding_provider_invoke('reservation', 'is_reservable', [$entity->localId]);
+      $reservable = reset($reservable);
+      if ($reservable) {
+        return array(ding_provider_get_form('ding_reservation_reserve_form', new DingReservationReservableEntity($entity), TRUE));
+      }
+    }
   }
 }
 
--- a/modules/ding_entity/ding_entity.api.php
+++ b/modules/ding_entity/ding_entity.api.php
@@ -63,54 +63,5 @@ function hook_ding_entity_buttons($type, $entity, $view_mode, $widget) {
 }
 
 /**
- * Help determine whether entities belongs to given pseudo-classes.
- *
- * Note that there's also a single entity version of this hook named
- * hook_ding_entity_is(), which is invoked from within a ding entity object and
- * therefore only support one entity at a time. This hook is therefore better to
- * implement if acting on multiple entities and the implementation can achieve
- * better performance by recieving all entities at once.
- *
- * @param array $entities
- *   The entities that needs to be checked against the pseudo-class.
- * @param string $class
- *   The pseudo class to check for.
- *
- * @return array
- *   The implementation must return an array with an entry for each of the
- *   passed entities keyed by their ding_entity_id. If the implementation
- *   doesn't have a say for a particular entity and/or pseudo-class it must set
- *   NULL to return something for that entity and signal that it is neutral.
- */
-function hook_ding_entity_entities_is(array $entities, $class) {
-  $return = [];
-  // Loop through each passed entity and check if it belongs to $class.
-  foreach ($entities as $entity) {
-    // Here we check if the entities belong to a pseudo class called 'book' to
-    // give an example, but it could be anything. This way the implementation
-    // can "invent" and add new pseudo-classes just by implementing this hook.
-    if ($class == 'book') {
-      // If we are certain this entity belongs to the pseudo class set TRUE:
-      if ($entity->getType() == 'book') {
-        $return[$entity->ding_entity_id] = TRUE;
-      }
-      // If we are absolutely certain this entity does not belong to the class and
-      // wanna overwrite any module who says so, we set FALSE for this entity:
-      elseif (strpos($entity->getType(), 'book') === FALSE) {
-        $return[$entity->ding_entity_id] = FALSE;
-      }
-    }
-    if (!isset($return[$entity->ding_entity_id])) {
-      // If we don't know about the pseudo class or don't have sufficient
-      // information about the entity to have a say, we can set a neutral response
-      // by setting NULL for this entity. This way we don't interfere with the
-      // final outcome and leaves more flexibility to other modules.
-      $return[$entity->ding_entity_id] = NULL;
-    }
-  }
-  return $return;
-}
-
-/**
  * @} End of "addtogroup hooks".
  */
--- a/modules/ding_entity/ding_entity.api.php
+++ b/modules/ding_entity/ding_entity.api.php
@@ -59,12 +59,9 @@ function hook_ding_entity_collection_view(&$object, $view_mode) {
 function hook_ding_entity_buttons($type, $entity, $view_mode, $widget) {
   // Adds resevation buttons is entity is resevable.
   if ($type == 'ding_entity' && $entity->is('library_material')) {
-    if (ding_provider_implements('reservation', 'is_reservable')) {
-      $reservable = ding_provider_invoke('reservation', 'is_reservable', [$entity->localId]);
-      $reservable = reset($reservable);
-      if ($reservable) {
-        return array(ding_provider_get_form('ding_reservation_reserve_form', new DingReservationReservableEntity($entity), TRUE));
-      }
+    $reservable = ding_provider_invoke('reservation', 'is_reservable', [$entity->localId]);
+    if (!empty($reservable[$entity->localId])) {
+      return array(ding_provider_get_form('ding_reservation_reserve_form', new DingReservationReservableEntity($entity), TRUE));
     }
   }
 }
--- a/modules/ding_provider/connie/connie.reservation.inc
+++ b/modules/ding_provider/connie/connie.reservation.inc
@@ -187,3 +187,13 @@ function connie_reservation_branch_name($branch_id) {
   }
   return NULL;
 }
+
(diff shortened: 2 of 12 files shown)


modules/ding_entity/ding_entity.api.php
```diff
@@ -62,6 +62,55 @@ function hook_ding_entity_buttons($type, $entity, $view_mode, $widget) {
...
+ * @param string $class
+ *   The pseudo class to check for.
+ *
+ * @return array
+ *   The implementation must return an array with an entry for each of the
+ *   passed entities keyed by their ding_entity_id. If the implementation
+ *   doesn't have a say for a particular entity and/or pseudo-class it must set
+ *   NULL to return something for that entity and signal that it is neutral.
+ */
+function hook_ding_entity_entities_is(array $entities, $class) {
```
kasperg: I understand what you are trying to do here but I am not sure I agree about the approach.

Bundling multiple checks into a single request is nice and efficient but it makes it very hard for developers to know whether they should 1) use 2) implement `ding_entity_entities_is()` or `ding_entity_is()` for their modules.

One approach to this would be to fall back to `ding_entity_is()` for modules and classes which does not support `ding_entity_entities_is()`. This quickly becomes very complex and it may not be possible at all.

Another approach could be to implement a new provider hook á la `fbs_reservation_is_reservable($provider_ids)` where the implementation for FBS would be about the same as in `fbs_ding_entity_entities_is()`.

A third approach might be to use the `reservable` info from the `availability_items` provider hook. If implemented correctly by the provider this could actually provide the correct information. Note that this creates a dependency from ding_reservation against an availability provider making the whole situation a bit more muddy.

Personally I am leaning towards a new provider hook.

I could also be OK with multiple calls to `->is('reservable')` although it may decrease performance.

Whatever you decide on please remember to clean up. #1139 also introduced changes to `connie` and elsewhere which may not be needed anymore.

modules/ding_reservation/js/ding_reservation_reservability.js
```diff
@@ -8,19 +8,26 @@
 
   Drupal.behaviors.ding_reservation = {
     attach: function(context) {
-      $(".ting-object .reserve-button", context).once('check-reservability', function() {
-        var reserveButton = $(this);
-        var entityId = reserveButton.data("entity-id");
+      var entityIds = [];
+      var selector = '.ting-object .reserve-button.check-reservability';
+      $(selector, context).once('check-reservability', function() {
```
kasperg: If this class is purely for JS reasons then please prefix `check-reservability` with `js-`. With that in mind I think we can also drop the `.reserve-button` class from the selector - perhaps even `.ting-object`.

modules/ding_entity/ding_entity.api.php
holt83: Thanks for the good explanation 👍 I understand your concerns.

I think it's important to get the performance benefit from the bundling. After reading your feedback I also lean towards a new provider hook, so I will implement that instead.

Regarding the changes to connie from #1139: I added the entity_is implementation to ensure to get reservation buttons when testing with that module. I havn't really used this module a lot, so this might be wrong. Would you say that it's not needed?

modules/ding_reservation/js/ding_reservation_reservability.js
holt83: Done

modules/ding_entity/ding_entity.api.php
holt83: Done

holt83:
@kasperg I updated the PR based:

- Use provider function for checking reservability instead of ding_entity pseudo classes check.
- Add js prefix to check-reservability
- Updated/cleaned up Connie
- Remove unneeded stuff

GitCop:

 * Commit: 2359f6ae8a567bbc3db2e46353aa4379438d8520

---

GitCop:

 * Commit: 2359f6ae8a567bbc3db2e46353aa4379438d8520

---

modules/ding_reservation/ding_reservation.module
```diff
@@ -67,7 +67,7 @@ function ding_reservation_menu() {
     'access arguments' => array('perform reservation'),
   );
 
-  $items['ding_reservation/%ding_entity/is_reservable'] = array(
+  $items['ding_reservation/%/is_reservable'] = array(
```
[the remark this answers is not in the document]
cableman: Found the following tabel (so IE and Edge may have issues):

Browser    |  Address bar   
------------|------------------------------
Chrome      |    32779        
Android      |    8192           
Firefox         |   >64k          
Safari            | >64k           
IE11                | 2047           
Edge 16          |2047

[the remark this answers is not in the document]
holt83: Interesting question :)

I have tried to do some calculations:

We have 31 chars in "ding_reservation//is_reservable" and let's say the domain part is 50 chars to be sure. Then we have 1.966 chars left. Each faust is 8 chars and we also need a comma so:

1.966 / 9 = 218,44

So we should be able to handle 218 fausts even in IE11 and Edge. 

This could be a problem on frontpage with a lot carousels. But a lot of them are loaded async at different timings, so there will not be one single request that checks reservability for alle materials on frontpage. Rather it will be split up in 2-3 requests. 
Also considering that the materials in the carousels are lazy loaded on demand themselves I don't think carousels will cause issues with the limitation.

Another case to consider is news/events with material lists. If an administrator adds over 218 materials to a list there could be an issue in IE/Edge, since these will all be checked in one single request.

I guess we could split the request up in max 200 fausts in the client. What do you think?

modules/ding_reservation/ding_reservation.module
```diff
@@ -143,9 +152,9 @@ function ding_reservation_ding_entity_buttons($type, $entity, $view_mode, $widge
                   'action-button',
                   'reserve-button',
                   'use-ajax',
+                  'js-check-reservability',
                 ),
-                'id' => 'reservation-' . $entity->id,
-                'data-entity-id' => $entity->id,
+                'data-local-id' => $entity->localId,
```
kasperg: :+1:

modules/ding_entity/ding_entity.api.php
```diff
@@ -57,8 +57,15 @@ function hook_ding_entity_collection_view(&$object, $view_mode) {
...
  */
 function hook_ding_entity_buttons($type, $entity, $view_mode, $widget) {
-  if ($type == 'ding_entity' && $entity->is('reservable')) {
-    return array(ding_provider_get_form('ding_reservation_reserve_form', new DingReservationReservableEntity($entity), TRUE));
+  // Adds resevation buttons is entity is resevable.
+  if ($type == 'ding_entity' && $entity->is('library_material')) {
+    if (ding_provider_implements('reservation', 'is_reservable')) {
+      $reservable = ding_provider_invoke('reservation', 'is_reservable', [$entity->localId]);
+      $reservable = reset($reservable);
+      if ($reservable) {
```
kasperg: I know what your intention is here but for correctness it would probably be a good idea to replace the two lines with something that ensures that we are checking availability for the same id that we passed into the function:

```php
if (!empty($reservable[$entity->localId])) {
```

modules/ding_reservation/ding_reservation.module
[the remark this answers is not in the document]
kasperg: You cannot assume what the length of the local ids is. For example the length of local ids for Icelandic libraries using DDELibra Aleph is 20 chars. For other library systems it might be something else.

You could chunk requests dynamically in the client based on length of local ids and browser but in increases the complexity somewhat. Another approach would be to convert it to a POST request.

If you come up with a method which chucks requests using GET then also remember to set appropriate headers which prevent caching of the response.

modules/ding_reservation/ding_reservation.module
```diff
@@ -113,10 +113,19 @@ function ding_reservation_ding_entity_menu(&$items, $type, $path, $index) {
...
+ * AJAX callback used to check the reservability of the passed entity ids.
  */
-function ding_reservation_is_reservable_ajax_callback($entity) {
-  drupal_json_output(array('reservable' => $entity->is('reservable')));
+function ding_reservation_is_reservable_ajax_callback($local_ids) {
+  $local_ids = explode(',', $local_ids);
+  $result = [];
+
+  // The reservation provider decides reservability.
+  if (ding_provider_implements('reservation', 'is_reservable')) {
```
kasperg: As mentioned previously I think we can remove this check and simplify the code.

modules/ding_reservation/ding_reservation.module
```diff
@@ -164,10 +173,14 @@ function ding_reservation_ding_entity_buttons($type, $entity, $view_mode, $widge
...
         // The last parameter to the form below (TRUE) hides the provider
         // options in the form (interest period and branch).
-        if ($entity->is('reservable')) {
-          $button = array(
-            ding_provider_get_form('ding_reservation_reserve_form', new DingReservationReservableEntity($entity), TRUE),
-          );
+        if (ding_provider_implements('reservation', 'is_reservable')) {
+          $reservable = ding_provider_invoke('reservation', 'is_reservable', [$entity->localId]);
+          $reservable = reset($reservable);
+          if ($reservable) {
```
kasperg: See comments elsewhere.

modules/ting/ting.admin.inc
```diff
@@ -136,7 +136,7 @@ function ting_admin_reservable_settings($form_state) {
   $form['reservation_settings'] = array(
     '#type' => 'fieldset',
     '#title' => t('Reservation settings'),
-
+    '#tree' => FALSE,
```
kasperg: Is this needed?

themes/ddbasic/template.php
```diff
@@ -829,15 +829,15 @@ function ddbasic_process_ting_object(&$vars) {
...
+          if ($vars['object']->is('library_material')) {
             $vars['content']['group_text']['reserve_button'] = ding_reservation_ding_entity_buttons(
               'ding_entity',
               $vars['object'],
               $vars['elements']['#view_mode'],
               'ajax'
             );
           }
-          if ($vars['object']->online_url) {
+          if ($vars['object']->is('online')) {
```
kasperg: :+1:

modules/ding_reservation/ding_reservation.module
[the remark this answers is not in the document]
holt83: I think I'll go with the POST request.

But why prevent caching if I was using GET? Reservability is not something that changes often, so I see no problem with the response being cached. Care to elaborate why you think this is necessary?

modules/ding_provider/connie/connie.reservation.inc
```diff
@@ -187,3 +187,13 @@ function connie_reservation_branch_name($branch_id) {
...
 }
+
+/**
+ * Implements provider reservation, is_reservable.
+ */
+function connie_reservation_is_reservable($provider_ids) {
+  // Make everything reservable.
+  $return = array_fill(0, count($provider_ids), TRUE);
+  $return = array_combine($provider_ids, $return);
+  return $return;
```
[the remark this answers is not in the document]
holt83: Thanks for the tip 👍

modules/ding_entity/ding_entity.api.php
```diff
@@ -57,8 +57,15 @@ function hook_ding_entity_collection_view(&$object, $view_mode) {
  * Return an array of buttons to add.
  */
 function hook_ding_entity_buttons($type, $entity, $view_mode, $widget) {
-  if ($type == 'ding_entity' && $entity->is('reservable')) {
-    return array(ding_provider_get_form('ding_reservation_reserve_form', new DingReservationReservableEntity($entity), TRUE));
+  // Adds resevation buttons is entity is resevable.
+  if ($type == 'ding_entity' && $entity->is('library_material')) {
+    if (ding_provider_implements('reservation', 'is_reservable')) {
```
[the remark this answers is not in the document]
holt83: Will remove

modules/ding_entity/ding_entity.api.php
holt83: 👍

modules/ding_reservation/ding_reservation.module
holt83: I agree with you that it shouldn't be necessary.  

Just want to mention that ding_reservation uses this pattern in a lot of other places. Do you have any guidelines on when to use/not to use?

modules/ting/ting.admin.inc
holt83: Last time [you questioned why I removed it](https://github.com/ding2/ding2/pull/1139#discussion_r292157752). Then i told you it was because it's default, but that maybe it would be good to keep for clarity. No response, but no biggie I just added it again.

modules/ding_reservation/ding_reservation.module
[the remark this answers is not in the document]
holt83: > I do not know what it is for Ereolen

Doesn't matter. FBS know nothing about these ids and only operate on fausts. So it doesn't make sense to look up these ids.

modules/ding_reservation/ding_reservation.module
holt83: Edit: After looking at code I see that the implements-pattern is only used when the provider callback is of optional nature.

modules/ding_reservation/ding_reservation.module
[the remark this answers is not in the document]
holt83: Fixed! Using post request instead.

modules/ding_provider/connie/connie.reservation.inc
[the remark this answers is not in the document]
holt83: Fixed!

modules/ding_entity/ding_entity.api.php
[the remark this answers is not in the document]
holt83: Fixed!

modules/ding_entity/ding_entity.api.php
holt83: Fixed!

modules/ding_reservation/ding_reservation.module
holt83: Fixed!

modules/ding_reservation/ding_reservation.module
holt83: Fixed!

holt83:
@kasperg PR is updated. I also rebased to get Scutinizer working.

holt83:
We discovered an additional issue with referenced materials on ding_news content missing reservation button, which we decided to handle here also: https://platform.dandigbib.org/issues/4415#note-17 

The problem is that ting_reference is using custom AJAX-callback which doesn't account for attached scripts during the rendering process in the callback.

I have chosen to include the reservability checking script manually in this case. I think it makes sense that it's the modules doing something non-standard Drupal, that should take care that the correct scripts is added.

An alternative solution would be just include the reservablilty checking script on every page in ding_reservation module.

modules/ding_reservation/ding_reservation.module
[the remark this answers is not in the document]
kasperg: > But why prevent caching if I was using GET? Reservability is not something that changes often, so I see no problem with the response being cached. Care to elaborate why you think this is necessary?

You are right. I was mixing up reservability and availability.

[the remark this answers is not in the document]
holt83: > As afar as I know Ereolen uses Publizon. Not FBS.

Ah ok. Wasn't aware of that.

kasperg:
I agree with the your solution to the missing JavaScript library.

PR #1461, opened 2019-06-14
4379: Prevent carousel timeouts

kasperg:
#### Link to issue

https://platform.dandigbib.org/issues/4379

#### Description

This PR fixes two issues with the current implementation of search carousels to prevent timeouts when retrieving carousel items:

1. Fix progressing through search result for carousel covers
2. Correct identification of search carousel offset/size mismatch 

Please check commit comments for additional details.

#### Checklist


cableman:
When I think over it I think i have seen this fix (or something like this) in the eReolen carrousel code.

PR #1462, opened 2019-06-18
4290: Make sure contact email body shows user email.

rasben:
#### Link to issue

https://platform.dandigbib.org/issues/4290

#### Description

If a user is logged in and fills out the contact form, the username is not displayed in the body, only the user link.
This is not great for the workflow of the librarians, so this adds the email regardless.

See examples here of how it is right now:

#### Checklist

### DIFF
--- a/ding2.install
+++ b/ding2.install
@@ -829,3 +829,10 @@ function ding2_update_7072() {
   // Clear cache before continuing update.
   drupal_flush_all_caches();
 }
+
+/**
+ * Update translations.
+ */
+function ding2_update_7073() {
+  ding2_translation_update();
+}
--- a/modules/ding_contact/ding_contact.module
+++ b/modules/ding_contact/ding_contact.module
@@ -84,6 +84,21 @@ function ding_contact_mail_alter(&$message) {
       $message['headers']['From'] = $from_value;
       $message['headers']['Reply-To'] = $message['params']['mail'];
       $message['headers']['Sender'] = $sitename;
+
+      // If it is not an anonymous user that has filled out the form,
+      // we want ot make sure their email is included in the message.
+      if (!empty($message['params']['sender']->uid)) {
+        $message['body'][0] .= ' ' . t(
+          "The user's email is !sender-mail.",
+          array(
+            '!sender-mail' => $message['params']['sender']->mail,
+          ),
+          array(
+            'langcode' => $message['language']->language,
+          )
+        );
+      }
+
       break;
   }
 }
--- a/translations/da.po
+++ b/translations/da.po
@@ -61388,6 +61388,9 @@ msgstr "\"Hvis du ønsker at der skal vises yderligere tekst på den globale \"e
 msgid "\"!sender-name (!sender-url) sent a message using the contact form at \"!form-url."
 msgstr "\"!sender-name (!sender-url) sendte en besked via kontaktformularen på \"!form-url."
 
+msgid "The user's email is !sender-mail."
+msgstr "Brugerens email er !sender-mail."
+
 msgid "\"!sender-name (!sender-url) has sent you a message via your contact \"form (!form-url) at !site-name."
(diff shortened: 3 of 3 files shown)


cableman:
I'm not user that I understand this. This fix puts a link to the users account on the site, but what we want is the users mail adresse (as the one that answers these mails don't have access to users account on the site).

cableman:
Okay, so the first screenshot is the result and the last the before change?

rasben:
@cableman, as i write "See examples here of how it is right now:" - the screenshots are just how it works right now, I havent added any screenshots of what it looks like when its fixed

In the standard contact.module, a logged in user will return an email looking like this:

**"Benjamin Rasmussen (https://www.viborgbib.dk/user/1337) sendte en besked via kontaktformularen på https://www.viborgbib.dk/contact"**

But thats not cool for the ones that has to answer the email, as Benjamin's email will only appear in the email-headers reply-to. If the email is forwarded to another person, that only contact will also disappear.


My fix changes it, so it should now say:

**"Benjamin Rasmussen (benjamin@reload.dk - https://www.viborgbib.dk/user/1337) sendte en besked via kontaktformularen på https://www.viborgbib.dk/contact"**

You can see the contact.module developers originale line-of-thought here:

`'!sender-url' => $params['sender']->uid ? url('user/' . $params['sender']->uid, array('absolute' => TRUE, 'language' => $language)) : $params['sender']->mail`

"If they dont have a user, then we'll use the email that they gave us - but if they do have a user, it would be super useful just to link to that user!" - but sadly, thats not really how it works in real life :)

--
I can see that there is a conflict now - I'll fix it ASAP, but this should still be reviewable :)

guddo:
Do you really wanna patch core with these ding2-specific changes, instead of altering within `ding_contact`?

rasben:
@guddo obviously I never want to patch core if I can avoid it, but I seem to recall (This PR is 3 months old now, so forgive my memory..) that I had no way of hooking into, without a core patch.

You are more than welcome to come with a change suggestion, if you see a way to do it :)

guddo:
Did you try `ding_contact_mail_alter()`?

rasben:
Well, that means I have to totally override the message that is being set by core, but i 'spose that is a better solution :)

I've tweaked my PR here:
https://github.com/ding2/ding2/pull/1462/commits/9eacd0c77df892e60ff3ea5f12bf1c7c6d71285f

So, now when a logged-in user sends a message, it will display the same message, regardless whether they're logged in or not:

**"!sender-name (!sender-mail) sent a message using the contact form at !form-url."**

guddo:
If it requires the complete body to be overridden like this in code, I’m not sure I think it’s the best solution - since it basically makes the usage of the core module redundant. A template file could be accepted I guess..

What about just altering the existing variable by imploding name and mail, within the function you already changed?

rasben:
>What about just altering the existing variable by imploding name and mail, within the function you already changed?

Uh, you mean to do some kind preg_match/regex to replace the relevant part of the string with an email?
Because that is exactly the kind of thing that is fragile, hard to read and has a higher chance of breaking fantastically one day..

guddo:
I just mean altering the variable that already exist (let’s call it ‘!sender-url’) and is used by core, eg. ‘!sender-url = “!sender-mail (!sender-mail)”’.

rasben:
But once the `ding_contact_mail_alter()` hook is run, the hook_mail() has already generated the string - I cant just switch out the variable, because its a compiled string.

These are the variables i have access to in ding_contact_mail_alter, with the $message parameter:

image

guddo:
I see. Maybe extend the `$message['body'][1]` with a string about the users email instead then, eg. `"[!sender-name]'s email address is: !sender-mail."`? Info would be duplicated for the logged-in users, however we are not overriding the complete body of the message generated by an external module (which might change in the future).

PR #1466, opened 2019-06-21
4382: Make dibs payments unique.

rasben:
#### Link to issue

https://platform.dandigbib.org/issues/4382

#### Description

By adding the uniqueoid=yes parameter to the payment URL, we make sure that if is a payment with an identical ID already has been submitted, it will be rejected.

#### Checklist

As I mention in the redmine Issue, I havent found any way of testing this locally, meaning we need to **be extra careful in our payments tests.**

### DIFF
--- a/modules/ding_dibs/ding_dibs.module
+++ b/modules/ding_dibs/ding_dibs.module
@@ -211,6 +211,10 @@ function ding_dibs_capture($delta = NULL, array &$transaction = NULL) {
   $param['orderid'] = $transaction['payment_order_id'];
   $param['transact'] = $transaction['payment_transaction_id'];
   $param['textreply'] = TRUE;
+  // If this parameter is present, the parameter orderid has to be unique
+  // compared to all other orderid's used by the merchant.
+  // If the orderid isn't unique, the call will be declined by a reason=7.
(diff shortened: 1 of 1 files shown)


modules/ding_dibs/ding_dibs.module
```diff
@@ -211,6 +211,10 @@ function ding_dibs_capture($delta = NULL, array &$transaction = NULL) {
   $param['orderid'] = $transaction['payment_order_id'];
   $param['transact'] = $transaction['payment_transaction_id'];
   $param['textreply'] = TRUE;
+  // If this parameter is present, the parameter orderid has to be unique
+  // compared to all other orderid's used by the merchant.
+  // If the orderid isn't unique, the call will be declined by a reason=7.
+  $param['uniqueoid'] = 'yes';
```
rasben: Could it ever be possible that an orderid might show up on upgrade-fps/testing environments?
Because if that can happen, then a payment would possibly be rejected on production with this change

rasben:
PLEASE check out my worries/comments on the issue before you review this:
https://platform.dandigbib.org/issues/4382

The actual code change here should be fine, but it has consequences that I'd like help with :)

PR #1468, opened 2019-06-25
4173 by Inlead: Use default panel variant from database.

bozeak:
#### Link to issue

#### Description

New sections must listen to default panel variant which is stored in database and not the one from code.

#### Checklist


JrgenGNielsen:
Behat test errors are unrelated. Reviewed and approved

PR #1472, opened 2019-06-26
4391 by Inlead: Place2book: Default values are overriding given values.

NicolaeGrigoriu:
#### Link to issue

https://platform.dandigbib.org/issues/4391

#### Description

Once you have created an event and defined values in Maintain copy, Kultunaut export and Passive event as you wish, then it stores the event fine and correct. But if you edit the event, it sets the default values again even if you do not want it to change

**Example 1:**

I have in the standards for p2b under /admin/config/ding/place2book/defaults checked Maintain copy, Kultunaut export and Passive arrangement.

I create an event that I want there to be ticket to. So I remove the checkmark in "Passive".

I save the event.

I would like to edit the body-field of the event, so I edit the event.

I now notice that "Passive" is again marked in the standards.

If I accidentally save, the ticket link is gone, and I cannot undo to a non-passive event, and if tickets have been sold, I cannot delete and start over.

**Example 2:**

I have in the standards for p2b under /admin/config/ding/place2book/defaults checked Maintain copy, and Kultunaut export

I want to make an event where there should not be tickets nor sent to Kultunaut. So I remove the checkmark in Maintain copy and Kultunaut export.

I save the event.

I would like to edit the body-field of the event, so I edit the event.

I now notice that Maintenance copy and Kultunaut export are checked again by the standards.

This means that a ticket will be created for an event, which should not be ticketed and also sent incorrectly to Kultunaut.

**Solution:** 

As mentioned in the API documentation in the PUT request the header input Content-Type should be application/json. Plus, get passive value direct from the event entity.

Could not reproduce example 2.

#### Screenshot of the result

image

#### Checklist

### DIFF
--- a/modules/ding_place2book/ding_place2book.fields.inc
+++ b/modules/ding_place2book/ding_place2book.fields.inc
@@ -46,9 +46,9 @@ function ding_place2book_field_widget_info() {
  * Implements hook_field_widget_form().
  */
 function ding_place2book_field_widget_form(&$form, &$form_state, $field, $instance, $langcode, $items, $delta, $element) {
-  $event = $event_id = $event_maker_id = $synchronize = NULL;
+  $event = $event_id = $event_maker_id = $synchronize = $passive = NULL;
   if (!empty($items)) {
-    list($event_id, $event_maker_id, $synchronize) = array_values(reset($items));
+    list($event_id, $event_maker_id, $synchronize, $passive) = array_values(reset($items));
     try {
       $p2b = ding_place2book_instance();
       $event = $p2b->getEvent($event_maker_id, $event_id);
@@ -177,8 +177,7 @@ function ding_place2book_field_widget_form(&$form, &$form_state, $field, $instan
         ),
       );
 
-      $passive = 0;
-      if (!isset($meta_data->event_state)) {
+      if ($passive === NULL) {
         $passive = variable_get('ding_p2b_default_passive', FALSE);
       }
       elseif ($meta_data->event_state == 'passive') {
@@ -474,6 +473,7 @@ function ding_place2book_field_presave($entity_type, $entity, $field, $instance,
             ),
           );
         }
+        $items[0]['passive'] = $settings['passive'];
       }
       break;
   }
--- a/modules/ding_place2book/ding_place2book.fields.inc
+++ b/modules/ding_place2book/ding_place2book.fields.inc
@@ -276,14 +276,12 @@ function ding_place2book_field_widget_form(&$form, &$form_state, $field, $instan
         $element['place2book']['prices_wrapper']['prices'][$key]['sale_begin_at'] = array(
           '#type' => 'date_popup',
           '#default_value' => $sale_begin_at,
-          '#states' => $required_state,
         );
 
         $sale_end_at = !empty($price->sale_end_at) ? date('Y-m-d H:i:s', strtotime($price->sale_end_at)) : '';
         $element['place2book']['prices_wrapper']['prices'][$key]['sale_end_at'] = array(
           '#type' => 'date_popup',
           '#default_value' => $sale_end_at,
-          '#states' => $required_state,
         );
 
         $element['place2book']['prices_wrapper']['prices'][$key]['id'] = array(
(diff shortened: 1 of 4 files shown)


GitCop:

 * Commit: 44138889eff049feb2e7e3b224198559d31f2aa5

---

GitCop:

 * Commit: 44138889eff049feb2e7e3b224198559d31f2aa5

---

holt83:
Why not use the passive boolean returned from the API instead of adding af new field?

modules/ding_place2book/ding_place2book.fields.inc
```diff
@@ -276,12 +276,14 @@ function ding_place2book_field_widget_form(&$form, &$form_state, $field, $instan
         $element['place2book']['prices_wrapper']['prices'][$key]['sale_begin_at'] = array(
           '#type' => 'date_popup',
           '#default_value' => $sale_begin_at,
+          '#states' => $required_state,
```
holt83: Seems unrelated?

modules/ding_place2book/lib/p2b/P2b.php
```diff
@@ -559,7 +559,7 @@ private function p2bRequest($url, $type, array $params = array(), $code = 200) {
     );
 
     if (!empty($params)) {
-      if ($type == 'POST') {
+      if ($type == 'POST' || $type == 'PUT') {
```
holt83: Seems unrelated?

NicolaeGrigoriu: As mentioned in the documentation http://developer.place2book.com/v1/events/#update-an-event. Header options should be of the type json

holt83: Sorry, I just can see how this solves the problem of default settings overriding actual values?

modules/ding_place2book/ding_place2book.fields.inc
holt83: Thanks for fix 👍

modules/ding_place2book/lib/p2b/P2b.php
NicolaeGrigoriu: when the node form is saved for the first time the POST request is sent to the p2b server with the `passive` property in a json object. This property is received on GET request and assigend in the line 180(https://github.com/ding2/ding2/pull/1472/files#diff-8f3bc34f993926671354976be58fae7aR180). 
But when the node is updated the PUT method is used. If no Content-Type json is set then it seems that an empty object is sent. So, the entity is not updated on the p2b side.

holt83: Ok, thanks for elaborating. So the problem wasn't really defaults overriding, but that the event was never updated?

When looking at how we determine the passive state now it just looks weird:

```
      $passive = $event->passive ?? 0;
      if (!isset($meta_data->event_state)) {
        $passive = variable_get('ding_p2b_default_passive', FALSE);
      }
      elseif ($meta_data->event_state == 'passive') {
        $passive = 1;
      }
```

So it looks like passive state can be found two places and that the one on$meta_data->event_state takes precedence. And if the there's nothing set on $meta_data->event_state it always fallback to default value regardless of the value on @event->passive.

Is this really correct? I would think one of them would be the correct one to get the passive state.

NicolaeGrigoriu: It seems that the problem is not in the module but in the place2book service. [See video](https://www.dropbox.com/s/utreqbku41wg87h/p2b.mp4?dl=0). When event is firstly created it saves all the needed fields. But when a price is created for a specific event then it recives the published status.

NicolaeGrigoriu: One more thing, when `$event->passive == TRUE` then `$meta_data->event_state == 'passive'` and `$event->state == 'passive'`.

holt83: > One more thing, when `$event->passive == TRUE` then `$meta_data->event_state == 'passive'` and `$event->state == 'passive'`.

Ok. Why not just check one of them then? It seems excessive and needlessly complex to check them all.

holt83: > It seems that the problem is not in the module but in the place2book service. [See video](https://www.dropbox.com/s/utreqbku41wg87h/p2b.mp4?dl=0). When event is firstly created it saves all the needed fields. But when a price is created for a specific event then it receives the published status. Is this on purpose?

Thanks for the vid. Really helpful 👍 

I guess it's because it doesn't make sense to have tickets on a passive event, so p2b automatically publishes the event. So yeah, I think it's on purpose.

NicolaeGrigoriu: If so, then the fix for the node update should be enough, I guess :) Everything else was reverted to initial state.

PR #1473, opened 2019-06-26
4001 by Inlead: Use validation as in ting_reference field.

AnatolyMuntean:
#### Link to issue

https://platform.dandigbib.org/issues/4001

#### Description

Change the validation mechanism of imported bpi materials to use same logic as in the ting_reference field. This makes sure that there's no different logic when validating materials, therefore producing unexpected results. For example, when after syndicating content the form reports validation errors regarding bpi materials, when, in fact, these materials do exist.

#### Screenshot of the result

None.

#### Checklist

- [x] My code passes continuous integration process. If not then I have added a comment explaining why this change should be exempt from the code standards and process.

None.


JrgenGNielsen:
reviewed and approved

PR #1476, opened 2019-06-26
4394 by Inlead: Place2book: Clone of events does not create new ticket.

NicolaeGrigoriu:
#### Link to issue

https://platform.dandigbib.org/issues/4394

#### Description

_Solution_

On the clone stage(node/{node id}/clone/confirm page) form submit it is copied the old event from the p2b and are removed all relations(id of event and prices ids).

#### Checklist

### DIFF
--- a/modules/ding_place2book/ding_place2book.fields.inc
+++ b/modules/ding_place2book/ding_place2book.fields.inc
@@ -99,7 +99,8 @@ function ding_place2book_field_widget_form(&$form, &$form_state, $field, $instan
         '#value' => TRUE,
       );
 
-      if (!empty($event)) {
+      // Hide prepopulated form if this is clone of the event.
+      if (!empty($event) && arg(2) != 'clone') {
         $begin_at = date('d/m/Y H:i:s', strtotime($event->begin_at));
         $end_at = date('d/m/Y H:i:s', strtotime($event->end_at));
         $element['place2book']['event_data'] = array(
@@ -461,7 +462,7 @@ function ding_place2book_field_presave($entity_type, $entity, $field, $instance,
           $synchronized_items = _ding_place2book_sync_p2b_entities($entity, $settings);
         }
 
-        // Only update field data if synchronizationn was successful.
+        // Only update field data if synchronization was successful.
         if (!empty($synchronized_items)) {
           $items = $synchronized_items;
         }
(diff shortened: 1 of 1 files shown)


GitCop:

 * Commit: 331b87c0ea057e747b712a05450cc541962e9982

---

GitCop:

 * Commit: 331b87c0ea057e747b712a05450cc541962e9982

---

holt83:
Isnt't this a bit hacky? On the node clone module description page it mentions a hook which can be used if we want to modify the cloned node: https://www.drupal.org/project/node_clone

Have you tried using this?

NicolaeGrigoriu:
I agree that this solution is not elegant. When the _prepopulate_ method is used then the `hook_clone_node_alter` is triggered before `hook_field_widget_form`. In this case, when unset the relation on the clone_node_alter the prices form will be empty. If the _save-edit_ method is used then the prices will be attached to the original event and will required another hack to clear the things up.

modules/ding_place2book/ding_place2book.fields.inc
```diff
@@ -566,9 +566,19 @@ function _ding_place2book_sync_p2b_entities($entity, array $settings) {
   $items = [];
 
   $event_id = !empty($settings['event_id']) ? $settings['event_id'] : FALSE;
+  $prices = $settings['prices_wrapper']['prices'];
   $event_maker_id = FALSE;
+
   // For clarity.
   $is_new = !$event_id;
+  if (arg(2) == 'clone') {
```
kasperg: Having a dependency on the page url deep down in a internal function does not seem like a good idea.

Is there something we can do using the clone api e.g. [`hook_clone_node_alter`](https://git.drupalcode.org/project/node_clone/blob/7.x-1.x/clone.api.php#L22)?

NicolaeGrigoriu: The `hook_clone_node_alter` is called before the node form is rendered. As it is working now it is not possible to use this hook. As a solution we can store event id in a hidden form field and use it when make field presave(the POST/PUT/DELETE requests). This field will be populated on the node edit form. And on node clone this id will be removed so, the prices will be attached to the newly created entity.

holt83:
I have made an alternative solution using hook_node_clone_alter() that also takes care of removing event ID in: https://platform.dandigbib.org/issues/4394

kasperg:
Thanks both of you. I prefer the solution in #1488 and am closing this in favour of it.

PR #1478, opened 2019-07-04
4429: Allow for variable length pincodes

xendk:
#### Link to issue

https://platform.dandigbib.org/issues/4429

#### Description

Allow for configuring maximum and minimun pincode lengths.

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1480, opened 2019-07-08
3573: Webtrekk campaign plus tracking

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/3573

#### Description

Adds Webtrekk tracking of ding_campaign_plus campaigns.

To be consistent we have chosen to always track these with events, since it's not possible to track with page parameters when they are not in cache and are loaded async via AJAX.

Instead of using a custom event I have used Drupal behaviors. We can then have an attach function which works both when campaigns are attached on page load and when they are attached async.

Currently we send the following parameters when a campaign is viewed:

```
define('DING_WEBTREKK_PARAMETER_CAMPAIGN_PLUS_TITLE', 62);
define('DING_WEBTREKK_PARAMETER_CAMPAIGN_PLUS_ID', 63);
```

I have tested that the campaign tracking with event works both when campaigns are added async (not in cache) and at page load (in cache).

In addition to tracking when a campaign is viewed we also add the `u_navigatedby` URL-parameter to the campaign destination link with the campaing title as value, to be able to link campaigns and associated page views. `u_navigtatedby` is also used on material links in carousels and is setup to work correctly in Webtrekk. The previously used `wt_mc` is not.

These changes also fixes a somewhat critical bug, where every campaign link with a URL-parameter present would not work, since the tracking parameter was always appended with '?'.

Also adds a new debugging mode, which outputs data about events when they are triggered and before they are send to Webtrekk. This information could be found by looking at network requests in dev tools, but in console we can output the information with better formatting. And it's easier to see/find. See screenshots below:

#### Screenshot of new debug mode

3573-webtrekk-campaign-tracking-debug-mode1

#### Screenshot of campaign plus tracking with debug mode

3573-webtrekk-campaign-tracking-debug-mode2

#### Checklist

There's a couple of TODOs that will be taken care off :)


holt83:
Seems like Scrutinizer is getting in stuck on "Installing Code" in every PR lately. Will investigate.

holt83:
I asked Scrutinizer about our issue and got following answer:

> Hi Simon 👋, 
> thanks for your message. 
> It seems like you have an older configuration file with old tools that are not supported anymore. I would recommend to delete the .scrutinizer.yml file from your repository and let the platform generate a new configuration file for your project. You can then use this new file and make tweaks to it.
> Let me know if that helps solve this.

holt83:
PR is ready for review

holt83:
I have tried to fix Scrutinizer config in #1496

modules/ding_campaign_plus/js/ding_campaign_plus.js
```diff
@@ -24,6 +24,8 @@
           if (response) {
             campaign.replaceWith(response);
             wrapper.slideDown('fast');
+            // We add new content so attach behaviors.
+            Drupal.attachBehaviors($('<div></div>').html(response));
```
kasperg: Attaching behaviors to something that is not bound to the document seems odd to me although I am not sure about how `replaceWith()` handles the response by reference or by value.

How about attaching behaviors to the wrapper instead?

modules/ding_webtrekk/ding_webtrekk.admin.inc
```diff
@@ -22,6 +22,12 @@ function ding_webtrekk_admin_settings_form() {
     '#description' => t('Enter your TagIntegration customer ID here.'),
     '#default_value' => variable_get('webtrekk_ti_id', ''),
   ];
-  
+  $form['webtrekk_debug'] = [
+    '#type' => 'checkbox',
+    '#title' => t('Enable debug mode'),
+    '#description' => t('In debug mode all events and associated data are logged in the JavaScript console before they are send.'),
+    '#default_value' => variable_get('webtrekk_debug', FALSE),
+  ];
```
kasperg: :+1:

modules/ding_webtrekk/ding_webtrekk.js
```diff
@@ -28,6 +28,10 @@
   };
 
   var pushEvent = function(event, eventData) {
+    if (Drupal.settings.dingWebtrekk.debugMode) {
+      console.log('Ding webtrekk - Pushing "' + event +  '"-event with data:');
+      console.log(JSON.stringify(eventData, null, 2))
```
kasperg: [`console.log`](https://developer.mozilla.org/en-US/docs/Web/API/console#Outputting_text_to_the_console) accepts all sorts of arguments. Is there a reason this is not just:

```js
console.log('Ding webtrekk - Pushing ', event, "-event with data:', eventData);
```

modules/ding_webtrekk/ding_webtrekk.js
```diff
@@ -165,6 +169,22 @@
...
+      // In these cases we can't use page parameters, since we don't have the
+      // required information at initial page load and we'll have to use an
+      // event. To be consistent we track every campaign with event.
+      $('.node-ding-campaign-plus', context) .once('js-ding-webtrekk', function() {
+        var eventData = $(this).data('ding-webtrekk-event');
+        // Sending a 'click'-event to webtrekk in this case may seem weird, but
+        // currently our Webtrekk setup only supports this type of event. If our
+        // setup is changed in the future it may be more appropiate to send
+        // another type of event. The important thing now is that we get the
+        // data to Webtrekk.
```
kasperg: :+1:

modules/ding_webtrekk/ding_webtrekk.module
```diff
@@ -255,6 +258,42 @@ function ding_webtrekk_ding_entity_view($entity, $view_mode) {
...
+ */
+function ding_webtrekk_process_node(&$variables) {
+  $node = $variables['node'];
+  if ($node->type == 'ding_campaign_plus') {
+    // Add u_navigatedby URL-parameter to campaign URLs, to be able to track
+    // page views triggered from campaigns. We use a process for this because
+    // campaign plus adds the variable in preprocess.
+    $url = &$variables['campaign_url'];
+    $url .= parse_url($url, PHP_URL_QUERY) !== NULL ? '&' : '?';
+    $url .= 'u_navigatedby=' . check_plain(rawurlencode($node->title));
```
kasperg: Instead of relying on string manipulation you could do the following using [url api functions](https://api.drupal.org/api/drupal/includes%21common.inc/function/drupal_parse_url/7.x):

```php
$url = drupal_parse_url($variables['campaign_url']);
$url['query']['u_navigatedby'] = $node->title;
$variables['campaign_url'] = url(
  $url['path'],
  [
    'query' => $url['query'],
    'fragment' => $url['fragment']
  ]
);
```

themes/ddbasic/templates/node/node--ding-campaign-plus--image.tpl.php
```diff
@@ -83,14 +83,8 @@
 ?>
 <div class="<?php print $classes; ?>" <?php print $attributes; ?>>
   <div class="ding-campaign ding-campaign--content campaign-image  <?php print $campaign_type; ?>">
-    <a href="<?php print $field_ding_campaign_plus_link[LANGUAGE_NONE][0]['url']; ?>?wt_mc=<?php print $wt_mc_id; ?>" target="__blank">
+    <a href="<?php print $campaign_url; ?>" target="__blank">
```
kasperg: 👍

modules/ding_webtrekk/ding_webtrekk.module
holt83: Yep, this is the way to do it :) I see now that my code will also make a wrong URL if the original URL has fragment. Will fix!

modules/ding_webtrekk/ding_webtrekk.js
holt83: If I remember correctly the JSON.stringify produced cleaner more readable output. I'l make some screenshots..

modules/ding_campaign_plus/js/ding_campaign_plus.js
holt83: Had some trouble getting this to work otherwise. Using the wrapper might work. will try it out!

modules/ding_webtrekk/ding_webtrekk.js
holt83: **JSON.stringify**
3573-debug-json-stringify

**Console.log argument**
3573-debug-console-log-argument

holt83: I think JSON.stringify looks better in this case. The collapsible object structure is a bit overkill, since we're only interesting in seeing the data in the data layer object.

modules/ding_campaign_plus/js/ding_campaign_plus.js
holt83: The wrapper has HTML which is not in response and is already in the DOM, so I opted for finding the nearest parent() of the campaign we just inserted and use that in attachBehaviors.

Went down bit of a rabbit hole with this. If you take a look at how Drupal core attaches behaviors when new content is added with ajax commands in insert. From Drupal core /includes/ajax.js (around lines 514-534):

```
// We don't know what response.data contains: it might be a string of text
// without HTML, so don't rely on jQuery correctly iterpreting
// $(response.data) as new HTML rather than a CSS selector. Also, if
// response.data contains top-level text nodes, they get lost with either
// $(response.data) or $('<div></div>').replaceWith(response.data).
var new_content_wrapped = $('<div></div>').html(response.data);
var new_content = new_content_wrapped.contents();

// For legacy reasons, the effects processing code assumes that new_content
// consists of a single top-level element. Also, it has not been
// sufficiently tested whether attachBehaviors() can be successfully called
// with a context object that includes top-level text nodes. However, to
// give developers full control of the HTML appearing in the page, and to
// enable Ajax content to be inserted in places where DIV elements are not
// allowed (e.g., within TABLE, TR, and SPAN parents), we check if the new
// content satisfies the requirement of a single top-level element, and
// only use the container DIV created above when it doesn't. For more
// information, please see http://drupal.org/node/736066.
if (new_content.length != 1 || new_content.get(0).nodeType != 1) {
  new_content = new_content_wrapped;
}
```

So it looks like it's also using the '<div></div>'-wrapper in certain cases, but it appears to not be because of the same reasons that I'm experiencing.

I just can't understand how it works when it's not using '<div></div>'-wrapper. The find method does ignore the root-element: http://danielnouri.org/notes/2011/03/14/a-jquery-find-that-also-finds-the-root-element/

Anyways, http://drupal.org/node/736066 is a rather interesting discussion about how to attach behaviors when new content is added.

modules/ding_webtrekk/ding_webtrekk.module
holt83: Fixed

holt83:
@kasperg PR is updated and I've answered comments.

I also discovered that i had forgot to remove a now unused file: ding_campaign_plus_trekk.js.

modules/ding_webtrekk/ding_webtrekk.js
holt83: Fixed. Using console arguments and JSON.stringify

holt83:
Regarding Scrutinizer:

> All variables defined by your module must be prefixed with your module's name to avoid name collisions with others. Expected start with "ding_webtrekk" but found "webtrekk_debug"

I think we should ignore this. I could add "ding"-prefix, but this will make it inconsistent with the other variables `webtrekk_ti_domain` and `webtrekk_ti_id` and adding prefixes to these will be problematic, because the values of those are probably hardcoded in local settings.php.

modules/ding_campaign_plus/js/ding_campaign_plus.js
kasperg: :+1: Thanks for the elaboration 😅

modules/ding_webtrekk/ding_webtrekk.js
kasperg: :+1:

modules/ding_webtrekk/ding_webtrekk.module
kasperg: :+1:

PR #1482, opened 2019-07-10
4399 by Inlead: P2b - Do not set ticket link if node is passive.

NicolaeGrigoriu:
#### Link to issue

https://platform.dandigbib.org/issues/4399

#### Description

In the hook_node_presave() implementation (ding_place2book_node_presave) we set field_ding_ticket_link with sales link to P2b, if the event is synced with P2b. We should add a check here, so that we only set this if event is not passive.

#### Checklist

This will work properly when the #1472 PR is merged. Since, on each node presave a PUT request is sent to p2b service.(See discussion in PR.)

### DIFF
--- a/modules/ding_place2book/ding_place2book.module
+++ b/modules/ding_place2book/ding_place2book.module
@@ -310,11 +310,12 @@ function ding_place2book_node_presave($node) {
   try {
     $p2b = ding_place2book_instance();
     $event = $p2b->getEvent($event_maker_id, $event_id);
-
-    $node_wrapper
-      ->field_ding_event_ticket_link
-      ->url
-      ->set($event->links->sales_location);
+    if (!$event->passive) {
+      $node_wrapper
+        ->field_ding_event_ticket_link
+        ->url
+        ->set($event->links->sales_location);
+    }
   }
   catch (Exception $e) {
     watchdog_exception('ding_place2book', $ex, t('Error appeared on getting data from p2b.'));


holt83:
I think it would be good investigate if it's possible to go from a passive event to an active and vice versa.

If we can go from a passive event to an active I believe it shouldn't be a problem. It will see that the event is p2b-sync, is not passive and doesn't have p2b sales URL. So it will fetch event and update ticket link accordingly.

But what about the other way around? It's more problematic because there will be at least one price and prices doesn't make sense on passive events. So it would have to delete the prices and also make sure to clear the ticket link field. As far as I can tell this doesn't happen now.

Maybe it's not possible to go from active to passive? If so, should we disable to option to select passive for existing non-passive events?

holt83:
Also this does have one negative side effect: Basicly, it will now always try to fetch the event for passive events, since it doesn't have p2b-URL in ticket link field and is p2b-synced. I don't know if there's a way around this.

NicolaeGrigoriu:
Thanks for the feedback. According to the [API](http://developer.place2book.com/v1/events/#field-descriptions) the `active` field should not conflict with the `passive` field.
p2b_state

guddo:
@holt83 Is there something we should do in regards to your comments? :)

holt83:
@guddo Not in the position to tell you what to do :) Just trying to give my thoughts about the changes and issues at hand and start a discussion.

But I think it would be relevant to test whether whether it's possible to go from passive to active and the other way around and how the system handles that? If it's not possible to go from active to passive for example, it safes us a lot of trouble. If that is the case, I think we should also disable to option to make it passive.

JrgenGNielsen:
reviewed and approved

PR #1423, opened 2019-05-08
4319: Show red asterisk for required p2b fields

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/4319

#### Description

Adds red asterisk to required p2b price fields. See screenshot.

The price field is only required if the event is **not** passive. Therefore we can't use the FAPI #required validation, since [it only cares about the initial field setting in backend](https://www.drupal.org/project/drupal/issues/2855139#comment-12472554).

To still take advangtage of the states API, we can manually theme the required marker in backend and keep our custom validate handler. Some additional logic is required though, since the fields should only be required for existing prices or if no prices currently exist for the event in p2b (we need at least one price for non-passive event for it to work correctly in p2b).

#### Screenshot of the result

4319-p2b-required-fields

#### Checklist

- [x] My complies with [our coding guidelines](../docs/code_guidelines.md).
- [x] My code passes our static analysis suite. If not then I have added a comment explaining why this change should be exempt from the code standards and process.
- [x] My code passes our continuous integration process. If not then I have added a comment explaining why this change should be exempt from the code standards and process.

#### Additional comments or questions

Also removed the notice about price is not deletable and disabling of the remove button for prices, for new prices that is not created yet:

4319-p2b-unwanted-price-deletable

PR #1483, opened 2019-07-10
4319: Only show validation errors for price name and value

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/4319#note-18

#### Description

A minor follow up from the changes in #1423  

Only show validation errors for price name and value and not also on sales period like before.

#### Screenshot of the result

4319-p2b-required-fields-only-name-value

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1485, opened 2019-07-12
4367: Header lenght changed and min-height removed

KristianB14:
#### Link to issue

https://platform.dandigbib.org/issues/4367

#### Description

For the campaign type "text", the header length is changed to be longer and the min-height is removed. For the campaign type "text-on-image", the text width is changed to be longer.

#### Screenshot of the result

image

image

#### Checklist

No further comments.


JrgenGNielsen:
reviewed and approved

PR #1486, opened 2019-07-12
4416: Resize event called on view when images are loaded

KristianB14:
#### Link to issue

https://platform.dandigbib.org/issues/4416

#### Description

View class added to imagesLoaded function that triggers a resize event for event teasers.

#### Screenshot of the result

image

#### Checklist

No further comments.


JrgenGNielsen:
reviewed and approved

PR #1487, opened 2019-07-17
4443: Update tracking tag field label and description

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/4443

#### Description

We have decided to support only one tracking tag for campaigns, and here the title is updated accordingly from plural to single form.

Also added a description text to give administrators a better idea of what "Tracking tag" is and what it's used for. 

#### Checklist

### DIFF
--- a/modules/ding_campaign_plus/ding_campaign_plus.features.field_instance.inc
+++ b/modules/ding_campaign_plus/ding_campaign_plus.features.field_instance.inc
@@ -45,6 +45,12 @@ function ding_campaign_plus_field_default_field_instances() {
         'type' => 'hidden',
         'weight' => 0,
       ),
+      'teaser_no_overlay' => array(
+        'label' => 'above',
+        'settings' => array(),
+        'type' => 'hidden',
+        'weight' => 0,
+      ),
     ),
     'entity_type' => 'node',
     'field_name' => 'field_ding_campaign_plus_image',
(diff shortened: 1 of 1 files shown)


modules/ding_campaign_plus/ding_campaign_plus.features.field_instance.inc
```diff
@@ -45,6 +45,12 @@ function ding_campaign_plus_field_default_field_instances() {
         'type' => 'hidden',
         'weight' => 0,
       ),
+      'teaser_no_overlay' => array(
+        'label' => 'above',
+        'settings' => array(),
+        'type' => 'hidden',
+        'weight' => 0,
+      ),
```
holt83: These seem unrelated, but this is how features exports it so better keep it as is.

modules/ding_campaign_plus/ding_campaign_plus.features.field_instance.inc
```diff
@@ -383,7 +419,9 @@ Campaign box is located at the bottom of the left column, on the page types wher
     'widget' => array(
       'active' => 1,
       'module' => 'options',
-      'settings' => array(),
+      'settings' => array(
+        'match_operator' => 'CONTAINS',
+      ),
```
holt83: I was very confused as to why features was exporting this. Looking at `options_field_widget_info()` from the options module, that defines the option_select widget, it doesn't have any settings.

Instead this setting is actually added to the option_select widget by ding_nodelist module in `ding_nodelist_field_widget_info_alter()`. I think this is really weird. I can't see the setting used anywhere in ding_nodelist or related modules and does it even make sense to have contains match operator on an select field?

@guddo Do you know why this was added? Is it something we can remove to avoid confusion?

holt83:
Closed in favor of #1492

PR #1488, opened 2019-07-18
4394: Handle cloned nodes with p2b sync using node clone API

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/4394

This is an alternative solution to the one presented in: #1476 

#### Description

Handles cloned nodes with p2b sync using hook_clone_node_alter(). Need to make sure that the cloned node doesn't reference the same event and prices in p2b API.

The node clone module supports two different methods of cloning and we need to handle them differently:

- **prepopulate**: In this mode the node clone is populated but not saved before showing the widget form. In this case we need to tell the widget form that it should use the cloned event and prices that we passed from alter instead of trying to fetch event and prices from p2b API
- **save-edit**: In this mode node clone is saved before showing the form. In this case we need to simulate the items array passed to our field presave implementation and remember to set 'do_sync'.

It's also possible to batch clone nodes from admin node view via the node clone action. This case need to be treated the same way as save-edit since no widget form is shown and the clone nodes is just saved immediately. We will have `$context['method'] == 'action'` in these cases, so if method is not prepopulate, we just need to simulate items array.

I have tested that it works with both methods of cloning and with node clone action.

#### Screenshot of the result

#### Checklist

### DIFF
--- a/modules/ding_place2book/ding_place2book.fields.inc
+++ b/modules/ding_place2book/ding_place2book.fields.inc
@@ -48,15 +48,26 @@ function ding_place2book_field_widget_info() {
 function ding_place2book_field_widget_form(&$form, &$form_state, $field, $instance, $langcode, $items, $delta, $element) {
   $event = $event_id = $event_maker_id = $synchronize = NULL;
   if (!empty($items)) {
-    list($event_id, $event_maker_id, $synchronize) = array_values(reset($items));
-    try {
-      $p2b = ding_place2book_instance();
-      $event = $p2b->getEvent($event_maker_id, $event_id);
+    // If items is non-empty we might be on a node clone form.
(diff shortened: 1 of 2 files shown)


holt83:
> However I must admit that I am a bit surprised about the amount of code added here compared to how little we can delete. I understand that the handling of the event id and ticket link might have been missing from the existing implementation but it only covers a part of the total change. Was the whole prepopulate/save-edit distinction missing? 🤔

I would say the primary reason for the amount of code is that we have to support the save-edit mode and the fact that the data structure we have saved in the DB isn't at all what our sync-function expects. It expects a lot of other data from the event in p2b, so we have to use a clone of that event and simulate the items-array that is passed to sync-function.

PR #1489, opened 2019-07-18
4444: Fix music shelfmark

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/4444

#### Description

#1390 has introduced a minor error in shelfmark for music materials. The order of the if-elseif is important here, since music is a special case and we don't need to check dk5 classification fields for these materials.

Also ensure that we use the special area codes defined by danMARC2 when translating code from 039.b.
See: http://www.kat-format.dk/danMARC2/felter03.htm

#### Screenshot of wrong music shelfmark

musik-forkert-shelfmark

#### Screenshot of correct music shelfmark

musik-korrekt-shelfmark

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1490, opened 2019-07-19
4446: Ensure no data is missing when updating tickets/events

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/4446

#### Description

We have observed that some data is getting removed from p2b when updating event in CMS. Apparently p2b will remove some data if it's missing in the update request we send to API.

This PR ensures that no data is missing by merging original event and tickets with our updates.

#### Screenshot of the result

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1492, opened 2019-07-30
4445: remove tracking field

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/4445

#### Description

The business has decided that the tracking tag field adds little to no value and to remove it to keep complexity down. In this PR we remove field and every reference to it.

See also #1480 which implements campaign tracking the correct way. 

#### Screenshot of the result

#### Checklist


modules/ding_campaign_plus/ding_campaign_plus.install
```diff
@@ -66,3 +66,10 @@ function ding_campaign_plus_uninstall() {
 
   variable_del('ding_campaign_plus_cron_last_run');
 }
+
+/**
+ * Remove tracking tag field.
+ */
+function ding_campaign_plus_update_7001() {
+  field_delete_field('field_ding_campaign_plus_track');
```
kasperg: Is there a reason why this update hook does not call [`field_purge_batch()`](https://api.drupal.org/api/drupal/modules%21field%21field.crud.inc/function/field_purge_batch/7.x) like we do [elsewhere](http://github.com/ding2/ding2/blob/master/modules/fbs/fbs.install#L47-L47) [in](http://github.com/ding2/ding2/blob/master/modules/ding_campaign/ding_campaign.install#L139-L139) [similar](http://github.com/ding2/ding2/blob/master/modules/ding_staff/ding_staff.install#L115-L115) update hooks?

holt83: I wonder why we do this actually. With field_delete_field() it will be marked for deletion and purged during future cron runs. 

I think it makes sense to do if it's a virtual field or if there's little to no data. But it if it contains data, field_purge_batch() could invoke alot of hooks during the update process.

Another question would be what $batch_size to use. We might not even remove all field data depending on the number of campaigns created.

kasperg: In my opinion we prefer to catch any problems that would be caused by this when we deploy rather than when cron runs subsequently. That is way harder to identify for us.

holt83: Ok, then maybe do a light purge with a batch size of 10 or something like that?

holt83: Added a lightweight purge!

holt83:
PR is rebased and updated with a call to `field_purge_batch()`

holt83:
Also: tried to fix campaign_plus behat tests by removing references to the tag field.

PR #1498, opened 2019-08-09
4467: Adgangsplatformen single signout

kasperg:
#### Link to issue

https://platform.dandigbib.org/issues/4467

#### Description

This PR updates the Ding Adgangsplatformen module to support and use [single signout](https://github.com/DBCDK/hejmdal/blob/master/docs/single-logout.md).

A new endpoint is defined which should live up to the provided specification. The module will return status code 403 for users which should not support single signout because they have not authenticated via Adgangsplatformen. @vibjerg has nodded to this approach although it is not mentioned in the current documentation.

#### Checklist

### DIFF
--- a/modules/ding_adgangsplatformen/ding_adgangsplatformen.module
+++ b/modules/ding_adgangsplatformen/ding_adgangsplatformen.module
@@ -143,6 +143,7 @@ function ding_adgangsplatformen_get_configuration() {
     'urlAccessToken' => 'https://login.bib.dk/oauth/token/',
     'urlResourceOwnerDetails' => 'https://login.bib.dk/userinfo/',
     'urlLogout' => 'https://login.bib.dk/logout/',
+    'singleLogout' => true,
   ));
 
   // All-ways set redirect url (editing not allowed) in admin.
@@ -213,6 +214,7 @@ function ding_adgangsplatformen_logout($hooked = FALSE) {
   global $base_url;
 
   $config = ding_adgangsplatformen_get_configuration();
+  $singleLogout = $config['singleLogout'] ? 'true' : 'false';
   $token = $_SESSION['oauth2token'];
 
   // Generate logout request for the authorization service and send the request.
@@ -221,7 +223,7 @@ function ding_adgangsplatformen_logout($hooked = FALSE) {
     array(
       'external' => true,
       'query' => array(
-        'singlelogout' => 'true',
+        'singlelogout' => $singleLogout,
         'access_token' => $token,
         'redirect_uri' => $base_url,
       ),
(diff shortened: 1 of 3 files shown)


modules/ding_adgangsplatformen/ding_adgangsplatformen.module
```diff
@@ -209,7 +216,16 @@ function ding_adgangsplatformen_logout($hooked = FALSE) {
   $token = $_SESSION['oauth2token'];
 
   // Generate logout request for the authorization service and send the request.
-  $logout_url = $config['urlLogout'] . '?access_token=' . $token . '&redirect_uri=' . $base_url;
+  $logout_url = url(
+    $config['urlLogout'],
+    array(
+      'external' => true,
+      'query' => array(
+        'singlelogout' => 'true',
```
cableman: This parameter should be configurable to allow to opt-out of singe sign out.

kasperg: :+1: Done.

modules/ding_adgangsplatformen/ding_adgangsplatformen.module
```diff
@@ -225,6 +243,41 @@ function ding_adgangsplatformen_logout($hooked = FALSE) {
...
+    session_destroy();
+  }
+
+  // Logout requests must never be cached. A cached response would cause
+  // Adgangsplatformen to think the user was logged out when it fact he was not.
+  drupal_page_is_cacheable(FALSE);
+  drupal_add_http_header('Status', "$status_code $reason");
+  drupal_add_http_header(
+    'X-Frame-Options',
+    'allow-from https://login.bib.dk'
```
cableman: Maybe this should be put into configuration as an hidden variable as this endpoint have changed once before.

kasperg: @cableman: I see several options here.

1. Add a hidden variable
2. Use the hostnames of one of the urls from the OAuth provider (assuming that the iframe will be located on the same system.
3. A combination of 1 and 2 where the hidden variable falls back back on the hostname.

cableman: @kasperg I was thinking something like https://github.com/ding2/ding2/blob/master/modules/ding_adgangsplatformen/includes/ding_adgangsplatformen.admin.inc#L43 so the solution was the same for these variables. That should not be changed but might need to be in the future...

kasperg: I have tried to implement this in da9824e.

kasperg:
I think the issues identified by Scrutinizer are out of scope for this PR.

modules/ding_provider/connie/connie.auth.inc
```diff
@@ -7,7 +7,7 @@
 /**
  * Implements hook_auth_single_sign_on().
  */
-function fbs_auth_single_sign_on($name) {
+function connie_auth_single_sign_on($name) {
```
cableman: 👍

PR #1500, opened 2019-08-14
4400: Avoid double rawurlencode on "Read more"-link

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/4400

#### Description

The $node_url variable has already been passed through Drupal's url function in template_preprocess_node(), which means it has been passed through rawurlencode once. So instead of using that, we manually construct the node uri and pass that to l-function.

#### Screenshot of the result

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1508, opened 2019-08-22
4126: Alter request hook

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/4126

**THIS PR WILL NOT WORK WITHOUT** : https://github.com/ding2/ting-client/pull/30

#### Description

Ting field search needs to take care only to alter targeted requests in `opensearch_pre_execute()` implementation. For search pages that is all search requests performed in the `search_result` pane and for material views (ting_object/ting_collection) we want to target the request for the main object of the page.

This PR introduces two mechanism for controlling this:

1. A new `hook_ting_field_search_alter_request($request)` which modules can implement to help decide whether request should be altered by returning a boolean. If anyone returns FALSE the request will not altered. If none returned FALSE and at least one returns TRUE request will be altered.
2. A static variable `$alter_request` that can be used to turn altering on/off for a series of requests if that is more convenient. In ting_field_search' default implementation of `hook_ting_field_search_alter_request()` it handles the static variable and it should be set through `ting_field_search_set_alter_request()` 

**The motivation for these changes** 

In the current implementation only a static variable is used to control alteration. Unfortunately this is not sufficient: 

If using a VIP-profile on a ting_field_search profile, that has sources that is not enabled in the main VIP-profile it will result in a 404 when viewing materials from those sources with the profile (see screenshot below).

When viewing a ting_object or ting_collection the first request is made very early in the load function: `ting_object_load()`, so the ting_object/ting_collection pane is never rendered and our hook, which set the context, is not invoked. Unfortunately there's currently no hooks available that let's us do something before an entity is loaded.

I have take a different approach in `hook_ting_field_search_alter_request()` looking at the id of the TingObject og TingCollection request and comparing that to the id URL-argument of the page. If it's the same we tell it to alter the request.

#### Screenshot of error when viewing material that is not searchable with main VIP-profile

4126-material-404

#### Checklist

I have also removed an unneeded hook implementation. From commit message:

"After the work with SAL the timing of the the generation of the cache
key has changed, such this implementation is not needed anymore for
opensearch to be able to serve the correct response when using profiles.

Before hook_opensearch_pre_execute was invoked after generation of cache
key and as a result it didn't take into account any modification made in
pre_execute implementation. This couldn't be changed without moving a
lot of code around, so the easiest solution at that time was to
introduce a new cache_key hook, such that ting_field_search could modify
it based on an active profile.

Hence in this commit we remove ting_field_search cache_key_alter
implementation. Since it's not needed it's better to remove to reduce
complexity."

The question is: should we remove the alter cache key completely? I added this hook to work around the issue described above. I'm not sure if there's any other use cases for this. Maybe just leave it?

### DIFF
--- a/modules/ting_field_search/ting_field_search.module
+++ b/modules/ting_field_search/ting_field_search.module
@@ -38,20 +38,32 @@ function ting_field_search_panels_pane_prerender($pane) {
  */
 function ting_field_search_panels_pane_content_alter($content, $pane, $args, $contexts) {
   if ($pane->type == 'search_result') {
-    ting_field_search_set_alter_request(NULL);
+    ting_field_search_set_alter_request(FALSE);
   }
 }
 
 /**
- * Helper function to set/unset the alter_request static variable.
+ * Set/get the static $alter_request variable.
  *
- * @param mixed $alter_request
- *   A boolean indication whether opensearch requests should be altered.
- *   NULL to unset again.
+ * This can be used to turn altering on/off for a series of requests or if
+ * alteration is dependent on timing from some hook, it can be more convenient
+ * than implementing alter_hook().
+ *
+ * @param boolean $alter_request (optional)
+ *   Use TRUE to alter behavior and FALSE to turn off again.
+ *
+ * @return boolean
+ *   Current value of $alter_request static variable.
  */
-function ting_field_search_set_alter_request($alter_request) {
-  $static = &drupal_static('ting_field_search_alter_request', NULL);
-  $static = $alter_request;
+function ting_field_search_set_alter_request($alter_request = NULL) {
+  // Default behavior is to not alter request.
+  $static = &drupal_static('ting_field_search_alter_request', FALSE);
+
+  if (isset($alter_request)) {
+    $static = $alter_request;
+  }
+
+  return $static;
 }
 
 /**
@@ -60,8 +72,6 @@ function ting_field_search_set_alter_request($alter_request) {
  * @see ting_field_search_opensearch_pre_execute()
  */
 function ting_field_search_ting_field_search_alter_request($request) {
-  $alter_request = NULL;
-
   // For new we only support alteration of the following types of requests, so
   // return early if it's not one of these.
   $class = get_class($request);
@@ -74,13 +84,9 @@ function ting_field_search_ting_field_search_alter_request($request) {
     return FALSE;
   }
 
-  // First check if the static alter_request variable is set. This can be
-  // switched on/off with ting_field_search_set_alter_request(). If alteration
-  // is dependent on timing from some hook, this can be more handy instead of
-  // implementing hook_ting_field_search_alter_request().
-  $alter_request = drupal_static('ting_field_search_alter_request', NULL);
-  if (isset($alter_request)) {
-    return $alter_request;
+  // First check if alter request is turned on via the static variable.
+  if (ting_field_search_set_alter_request()) {
+    return TRUE;
   }
 
   // Special handling for object and collection requests. The request should be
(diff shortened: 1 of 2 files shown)


holt83:
Hmm why no behat tests starting?

holt83:
Really weird.,. when I update the vejlebib:4126-alter-request-hook this PR doesn't automatically update commits/diffs. I have to create new PR or close PR and change base branch and back again for it to pick up changes. 

Ever experienced this before?

holt83:
Ok, seems there was some problems with Github. Test started running now and commits/diffs are updated.

modules/ting_field_search/ting_field_search.module
```diff
@@ -91,54 +151,14 @@ function ting_field_search_get_active_profile() {
...
-  }
-
-  // Do we support the class of this request?
-  $class = get_class($request);
-  $supported_classes = array(
-    'TingClientObjectRequest',
-    'TingClientCollectionRequest',
-    'TingClientSearchRequest',
-  );
-  if (!in_array($class, $supported_classes)) {
```
kasperg, line 141: I do not think it is a good idea to move this code to the hook implementation. The code in this function has some expectations regarding what methods can be called on the request e.g. `setProfile()`. Some of these methods are not available on all types of requests. Consequently I think it should rely on this method on this function to filter away irrelevant request types instead of relying on a hook implementation to handle it.

modules/ting_field_search/ting_field_search.module
```diff
@@ -21,38 +21,98 @@ function ting_field_search_ctools_plugin_directory($module, $plugin) {
...
+  }
+
+  // Special handling for object and collection requests. The request should be
+  // altered if it's for the main material of the page.
+  if ($class == 'TingClientObjectRequest' || $class == 'TingClientCollectionRequest') {
+    $path = current_path();
+    if (strpos($path, 'ting/object') === 0 || strpos($path, 'ting/collection') === 0) {
+      $parts = explode('/', $path);
+      // Require precise path and non-empty id.
+      if (count($parts) == 3 && !empty($parts[2])) {
```
kasperg, line 93: I would not mind replacing these 4 LOC with a regular expression but that may be a matter of preference.

kasperg (changes requested):
:+1: It is quite tricky what is going on here but I think I get it. The comments are helpful.

I have a few comments.

modules/ting_field_search/ting_field_search.module
```diff
@@ -21,38 +21,98 @@ function ting_field_search_ctools_plugin_directory($module, $plugin) {
...
+ * Helper function to set/unset the alter_request static variable.
+ *
+ * @param mixed $alter_request
+ *   A boolean indication whether opensearch requests should be altered.
+ *   NULL to unset again.
+ */
+function ting_field_search_set_alter_request($alter_request) {
+  $static = &drupal_static('ting_field_search_alter_request', NULL);
+  $static = $alter_request;
+}
```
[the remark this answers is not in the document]
holt83: Done. Also changed default value of the static to FALSE. The default behavior is to not alter request.

modules/ting_field_search/ting_field_search.module
holt83: Done. I agree it makes more sense to have in `ting_field_opensearch_pre_execute()`.

modules/ting_field_search/ting_field_search.module
holt83: Updated to use a regex instead. Much cleaner :)

holt83:
Thanks for the comments! I have updated the with suggested changes.

Also: I merged code from upstream, did some testing and got a fatal error from ting_subsearch, since it was using the old `ting_field_search_set/get_context(`) methods. I've tried to fix that also.

I can't find other code using these functions.

kasperg (approved):
:+1: LGTM

PR #1510, opened 2019-08-29
4241 by Inlead: Created pane widget to display Sections.

bozeak:
#### Link to issue

https://platform.dandigbib.org/issues/4241

#### Description

Was added custom panel pane which allows to compose custom displays with "Section" terms, like in Nodelist module. Also this pane was added to ding_ipe_filters module.

#### Screenshot of the result

1567063264135_sections

#### Checklist

Regarding Scrutinizer issues: "Do not use the raw $form_state['input'], use $form_state['values'] instead where possible" - because we always have to keep form data between AJAX requests and for correct form re-rendering, this can't be done in another way at the moment.


JrgenGNielsen:
reviewed and approved

PR #1511, opened 2019-08-29
4494: Add agency to SSO login url

cableman:
#### Link to issue

https://platform.dandigbib.org/issues/4494

#### Description

Added option to add agency to the auth url to automatically select the library at adgangsplatformen.

#### Screenshot of the result

Added setting to `/admin/config/ding/adgangsplatformen` allow this new setting.

Screenshot 2019-08-29 at 09 33 57

#### Checklist

No comments.

### DIFF
--- a/modules/ding_adgangsplatformen/ding_adgangsplatformen.module
+++ b/modules/ding_adgangsplatformen/ding_adgangsplatformen.module
@@ -37,7 +37,7 @@ function ding_adgangsplatformen_init() {
 function ding_adgangsplatformen_menu() {
   $items = array();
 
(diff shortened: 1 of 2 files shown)


modules/ding_adgangsplatformen/ding_adgangsplatformen.module
```diff
@@ -37,7 +37,7 @@ function ding_adgangsplatformen_init() {
 function ding_adgangsplatformen_menu() {
   $items = array();
 
-  $items['admin/config/ding/auth'] = array(
```
kasperg: :+1:

PR #1513, opened 2019-09-02
4507: Ensure session variable is removed completely

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/4507

#### Description

Ensure that the messages session variable is removed completely when removing errors to allow empty search with profile.

#### Screenshot of the result

#### Checklist


holt83:
Scrutinizer complaining about an unrelated variable missing module prefix. We decided to ignore this in:  https://github.com/ding2/ding2/pull/1480#issuecomment-521687648

PR #1515, opened 2019-09-03
4381: Disable Autologout once Adgangsplatformen is installed

kasperg:
#### Link to issue

https://platform.dandigbib.org/issues/4381

#### Description

Autologout in combination with single signout from Adgangsplatformen
will log out the user from all platforms when inactive on one. That
is not helpful - especially in combination with platforms used for
passive media consumption.

We do not remove the module entirely as some libraries may want to
retain the autologout functionality. They will have to reenable it
themselves subsequently.

#### Checklist

### DIFF
--- a/modules/ding_base/ding_base.info
+++ b/modules/ding_base/ding_base.info
@@ -2,7 +2,6 @@ name = Ding base
 core = 7.x
 package = Ding!
 project = ding_base
-dependencies[] = autologout
 dependencies[] = ctools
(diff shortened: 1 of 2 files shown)


holt83:
@kasperg Do we need to hide the issues on Scrutinizer for them stop appearing?
https://scrutinizer-ci.com/docs/reviews/filtering_issues_and_false_positives

PR #1516, opened 2019-09-03
4399: Remove ticket links from existing passive events

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/4399

#### Description

In the [previous PR](https://github.com/ding2/ding2/pull/1482) we have ensured that we don't put the p2b sales URL in the ticket link field for passive events.

However, there still might be some existing passive events with ticket links, so in this PR we add an update function which find all these and ensure that the ticket link is removed. Since this can potentially result in several request to p2b API, we use the batch API to ensure no timeouts and to give partial updates after each batch.

#### Screenshot of the result

#### Checklist

### DIFF
--- a/modules/ding_place2book/ding_place2book.install
+++ b/modules/ding_place2book/ding_place2book.install
@@ -197,7 +197,7 @@ function ding_place2book_update_7007(&$sandbox) {
   $query = new EntityFieldQuery();
   $query->entityCondition('entity_type', 'node')
     ->entityCondition('bundle', 'ding_event')
-    ->fieldCondition('field_ding_event_date', 'value2', gmdate('Y-m-d H:i:s'), '>')
+    ->fieldCondition('field_ding_event_date', 'value2', gmdate(DATE_FORMAT_DATETIME), '>')
     ->fieldCondition('field_place2book', 'synchronize', '1', '=')
     ->fieldCondition('field_place2book', 'event_id', 'NULL', '!=')
     ->fieldCondition('field_place2book', 'event_maker_id', 'NULL', '!=');
--- a/modules/ding_place2book/ding_place2book.install
+++ b/modules/ding_place2book/ding_place2book.install
@@ -219,7 +219,8 @@ function ding_place2book_update_7007(&$sandbox) {
   $chunk = array_slice($nodes, $sandbox['processed'], $limit);
 
   foreach ($chunk as $nid => $node) {
-    list($event_id, $event_maker_id) = array_values($field_place2book);
+    $field_place2book = field_get_items('node', $node, 'field_place2book');
+    list($event_id, $event_maker_id) = array_values($field_place2book[0]);
     try {
       $p2b = ding_place2book_instance();
       $event = $p2b->getEvent($event_maker_id, $event_id);
--- a/modules/ding_place2book/ding_place2book.install
+++ b/modules/ding_place2book/ding_place2book.install
@@ -219,7 +219,7 @@ function ding_place2book_update_7007(&$sandbox) {
   $limit = 10;
   $chunk = array_slice($nodes, $sandbox['processed'], $limit);
 
-  foreach($chunk as $nid => $node) {
+  foreach ($chunk as $nid => $node) {
     list($event_id, $event_maker_id) = array_values($field_place2book);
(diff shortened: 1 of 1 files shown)


modules/ding_place2book/ding_place2book.install
```diff
@@ -178,10 +178,79 @@ function ding_place2book_update_7005(&$sandbox) {
...
+function ding_place2book_update_7007(&$sandbox) {
+  $nodes = [];
+
+  // Get the ding_event nodes we need to consider. We only consider those with
+  // an event end-date that hasn't passed. We can't make any assumptions about
+  // whether the nodes are publsihed or not, so no condition on $node->status.
+  $query = new EntityFieldQuery();
+  $query->entityCondition('entity_type', 'node')
+    ->entityCondition('bundle', 'ding_event')
+    ->fieldCondition('field_ding_event_date', 'value2', gmdate('Y-m-d H:i:s'), '>');
```
kasperg: Instead of this string we should probably use `DATE_FORMAT_DATETIME`.

Also: Is there any reason why you use `gmdate()` I has expected `date` to be sufficient. `gmdate()` does not seem to be used anywhere within the date module itself.

modules/ding_place2book/ding_place2book.install
```diff
@@ -178,10 +178,79 @@ function ding_place2book_update_7005(&$sandbox) {
...
+  }
+
+  // Potentially we have to make request to p2b for all fetched ding_event nodes
+  // so keep the amount of nodes we process each run low.
+  $limit = 10;
+  $chunk = array_slice($nodes, $sandbox['processed'], $limit);
+
+  foreach($chunk as $nid => $node) {
+    // Avoid making request if this node isn't synchronized.
+    if (!$field_place2book = ding_place2book_is_ding_event_node_synchronized($node)) {
```
kasperg: This seems to be based the existance of the `field_place2book` field. Could we check for that in the initial query? If not the we could filter away these nodes before the actual processing.

To the memory footprint low for each batch process we should remove ids which are not relevant for this step before loading the entities.

holt83:
This is getting a bit complex and I have considered simplifying the code in our `hook_node_presave()`, by removing this check: https://github.com/ding2/ding2/blob/master/modules/ding_place2book/ding_place2book.module#L296 

It prevents the hook from making requests to p2b API if the node already has a p2b-URL. The concern it that we can end up making a lot of requests if multiple ding_event nodes are updated in a batch.

But maybe it's not that big of a problem? The get request to p2b-API are certainly very fast. Removing the check would make the code simpler and it will check for passive everytime an event is updated and ensure ticket link field is updated accordingly. It would also make this update-function simpler, because the we cant just do `node_save($node)` and our presave will handle the rest.

In the end I chose to keep the presave like it is, but I'm on the fence about it..

modules/ding_place2book/ding_place2book.install
holt83: The date string are in UTC in the database. gmdate ensures that we get the current time as datetime string in UTC.

holt83: Will use DATE_FORMAT_DATETIME

modules/ding_place2book/ding_place2book.install
holt83: Hmm.. that may be a good idea. I'll take a look at it 👍

holt83:
@kasperg I updated the PR with you suggestions.

Let me know if you prefer another solution that `gmdate()` for handling UTC strings in db: https://github.com/ding2/ding2/pull/1516#discussion_r321598540

modules/ding_place2book/ding_place2book.install
holt83: Fixed!

modules/ding_place2book/ding_place2book.install
holt83: Fixed!

holt83:
Good catch, thanks 👍 I updated PR with a fix.

kasperg:
> I also added some small optimizations: Query is only run once now on initial call and we only load `$chunk` nodes each pass with `entity_load()`.

Excellent!

PR #1519, opened 2019-09-06
3543: Fixed argument for bpi_validate_material

rimi-itk:
#### Link to issue

https://platform.dandigbib.org/issues/3543

#### Description

Materials shared via the BPI web service are validated when syndicating content, but apparently the signature of `bpi_validate_material` which is used for validation has been changed.

#### Checklist

The doc block for `bpi_validate_material` (https://github.com/ding2/ding2/blob/master/modules/bpi/bpi.module#L952-L962) states that `$ting_object_id` should be a faust number (“the number after the colon” as I understand it). Is this correct?

It seems that a full object id, e.g. `870970-basis:03258351` is required.


rimi-itk:
Thanks for clarifying and updating the DocBlock.

PR #1520, opened 2019-09-09
3800: Interlibrary loans (changes to core for the ding_ill module)

tripox:
#### Link to issue

https://platform.dandigbib.org/issues/3800

#### Description

* Hides holdings when not available
* Styling for new button for interlibrary loans
* Shows a message when the material is available for interlibrary loan

#### Screenshot of the result
Screenshot 2019-09-09 at 16 26 24
Screenshot 2019-09-09 at 16 23 50

#### Checklist

Buttons looks a bit odd on my local environment, however, this problem shows up without any changes from the current master branch:

Screenshot 2019-09-09 at 15 49 19

### DIFF
--- a/modules/ding_availability/ding_availability.module
+++ b/modules/ding_availability/ding_availability.module
@@ -251,10 +251,6 @@ function ding_availability_holdings($provider_ids) {
           'There are @count users in queue to loan the material.');
       }
 
-      if ($total_and_ordered_count == 0) {
-        ddbasic_body_class('no-holdings');
-      }
-
       // Theme the output.
       $item['html'] = theme('ding_holdings', $variables);
     }
--- a/modules/ding_availability/ding_availability.module
+++ b/modules/ding_availability/ding_availability.module
@@ -251,6 +251,10 @@ function ding_availability_holdings($provider_ids) {
           'There are @count users in queue to loan the material.');
       }
 
+      if ($total_and_ordered_count == 0) {
+        ddbasic_body_class('no-holdings');
+      }
+
       // Theme the output.
       $item['html'] = theme('ding_holdings', $variables);
     }
--- a/modules/ding_availability/js/ding_availability_labels.js
+++ b/modules/ding_availability/js/ding_availability_labels.js
@@ -139,7 +139,7 @@
 
           if (group.length === 0) {
             var msg = Drupal.t('Not available');
-            if (Drupal.settings.ding_ill) {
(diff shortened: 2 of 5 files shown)


modules/ding_availability/ding_availability.module
```diff
@@ -251,6 +251,10 @@ function ding_availability_holdings($provider_ids) {
           'There are @count users in queue to loan the material.');
       }
 
+      if ($total_and_ordered_count == 0) {
+        ddbasic_body_class('no-holdings');
```
cableman: Don't use DDBasic functions in backend modules as this will prevent libraries not using DDBasic to use DDB CMS and this module. E.g. ereolen, ereolenGo, Icelandic libraries etc.

themes/ddbasic/sass/components/ting-object/ting-object.scss
```diff
@@ -712,6 +714,9 @@
         clear: both;
         padding: 28px 0 26px;
         border-top: 1px solid $charcoal-opacity-light;
+        .no-holdings & {
```
cableman: You have removed the code that added the no-holdings class or is it added by ill module in som other way?

modules/ding_availability/js/ding_availability_labels.js
```diff
@@ -138,7 +138,11 @@
           group = $('.js-unavailable', groups_wrapper);
 
           if (group.length === 0) {
-            group = $('<p class="js-unavailable">' + Drupal.t('Not available') + ': </p>');
+            var msg = Drupal.t('Not available');
+            if (Drupal.settings.ding_ill) {
```
cableman: Nitpicking: if i understand it this "ding_ill" is a new module not part of core. So would the right check be:

`if (Drupal.settings.hasOwnProperty('ding_ill))`

themes/ddbasic/sass/components/ting-object/ting-object.scss
tripox: It's now being added by the `ding_ill` module. :)

modules/ding_availability/js/ding_availability_labels.js
tripox: Good point, I will update it.

themes/ddbasic/sass/components/ting-object/ting-object.scss
cableman: 👍

PR #1523, opened 2019-09-11
3199: Update antibot configuration

JrgenGNielsen:
#### Link to issue
https://platform.dandigbib.org/issues/3199

#### Description
Issue https://platform.dandigbib.org/issues/3696 adds antibot javascript to the fields user_login & user_pass in the login form. This trigger a "Synchronous XMLHttpRequest" error.

This PR removes the user_login & user_pass from antibot configuration

#### Screenshot of the result
N/A

#### Checklist

### DIFF
--- a/modules/ding_contact/ding_contact.install
+++ b/modules/ding_contact/ding_contact.install
@@ -225,3 +225,13 @@ function ding_contact_info() {
 function ding_contact_update_7000() {
   module_enable(array('antibot'));
 }
+
+/**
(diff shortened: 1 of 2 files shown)


holt83:
Do you have any details about why antibot JS would trigger the error on those forms? Antibot is still used on other forms, will we start seeing it there too?

holt83:
It should be sufficient to just update feature. Is the update function really needed?

PR #1525, opened 2019-09-13
4173 by Inlead: Make sure default panel is reverted upon change.

bozeak:
#### Link to issue

#### Description

Added hook_update() to restore Section "base" panel page settings to defaults from code.

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1526, opened 2019-09-13
4514: Ensure that admin users can not be logged out

cableman:
#### Link to issue

https://platform.dandigbib.org/issues/4514

#### Description

If you get `ding_adgangsplatformen` enabled but not disabled `ding_user_form` and the other way around you will end up with a site that non-provider users can't login.

#### Screenshot of the result

No screenshot

#### Checklist

No comments

### DIFF
--- a/ding2.install
+++ b/ding2.install
@@ -106,6 +106,13 @@ function ding2_install() {
     module_disable(array('overlay'));
   }
 
+  // Enable ding_user_form to allow provider login. It's enabled here as
+  //new/local sites should use "user form" and updated production should via
+  // update hooks turn on adgangsplatformen.
+  if (module_exists('ding_user_form')) {
+    module_enable(array('ding_user_form'));
+  }
+
   // Ignore any rebuild messages.
   node_access_needs_rebuild(FALSE);
 
--- a/ding2.install
+++ b/ding2.install
@@ -107,12 +107,10 @@ function ding2_install() {
   }
 
   // Enable ding_user_form to allow provider login. It's enabled here as
-  //new/local sites should use "user form" and updated production should via
+  // new/local sites should use "user form" and updated production should via
   // update hooks turn on adgangsplatformen.
-  if (module_exists('ding_user_form')) {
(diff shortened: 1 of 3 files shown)


ding2.install
```diff
@@ -106,6 +106,13 @@ function ding2_install() {
     module_disable(array('overlay'));
   }
 
+  // Enable ding_user_form to allow provider login. It's enabled here as
+  //new/local sites should use "user form" and updated production should via
+  // update hooks turn on adgangsplatformen.
+  if (module_exists('ding_user_form')) {
+    module_enable(array('ding_user_form'));
+  }
```
kasperg: Why add this when [we allow the user to select which modules to enable during install - including `ding_user_form`](https://github.com/ding2/ding2/blob/5a7a2f9de2a3feb816e8c453a0c0874ce1cccdb8/ding2.profile#L378-L388)?

ding2.install
```diff
@@ -106,6 +106,13 @@ function ding2_install() {
     module_disable(array('overlay'));
   }
 
+  // Enable ding_user_form to allow provider login. It's enabled here as
+  //new/local sites should use "user form" and updated production should via
```
kasperg:
```suggestion
  // new/local sites should use "user form" and updated production should via
```

Note the Scrutinizer error.

ding2.install
holt83: `module_exists` will only return TRUE if the module is already enabled.

cableman: Fixed

modules/ding_user_form/ding_user_form.install
```diff
@@ -0,0 +1,17 @@
+<?php
+
+/**
+ * @file
+ * Handle module installation.
+ */
+
+/**
+ * Implements hook_install().
```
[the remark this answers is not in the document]
cableman: Fixed

ding2.install
cableman: Fixed

kasperg:
Closed in favour of #1533.

PR #1527, opened 2019-09-16
4523 by Inlead: Fix nodelist carousel slick initialization.

bozeak:
#### Link to issue

https://platform.dandigbib.org/issues/4523

#### Description

Error thrown by ding_nodelist carousel on Slick functionality initialization was breaking Ding IPE js functionality execution.

#### Checklist

### DIFF
--- a/modules/ding_nodelist/js/carousel.js
+++ b/modules/ding_nodelist/js/carousel.js
@@ -13,23 +13,21 @@
           }
         });
 
-        if ($(this).find('.ding_nodelist-items').hasClass('slick-initialized') === false) {
-          $(this).find('.ding_nodelist-items').slick({
-            nextArrow: '<i class="icon-next"></i>',
-            prevArrow: '<i class="icon-prev"></i>',
-            autoplay: true,
-            speed: 500,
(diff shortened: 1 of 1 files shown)


modules/ding_nodelist/js/carousel.js
```diff
@@ -13,21 +13,23 @@
...
-          dots: true,
-          infinite: true,
-          slidesToScroll: 1,
-          slidesToShow: 1,
-          customPaging: function(slick, index) {
-            return '<a>' + (index + 1) + '</a>';
-          }
-        });
+        if ($(this).find('.ding_nodelist-items').hasClass('slick-initialized') === false) {
+          $(this).find('.ding_nodelist-items').slick({
```
kasperg: Is there a reason why you do not use [`:not()`](https://api.jquery.com/not-selector/) instead of running the query twice?

```js
$(this).find('.ding_nodelist-items:not(.slick-initialized)').slick({...});
```

bozeak: Fixed.

PR #1529, opened 2019-09-18
4161: Implement TOS active accept

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/4161

#### Description

Adds Terms of service active accept as required by EU-law. More information on redmine: https://platform.dandigbib.org/issues/4161#note-12 

There was no styling of disabled buttons in our main theme, so I had to add that also.

I added a translation of "I accept the terms of service". Should I trigger a translation import in an update hook or is that something we take care of once each release?

#### Screenshot of the result

4161-no-accept

4161-accept

#### Checklist

### DIFF
--- a/modules/ding_debt/plugins/content_types/debts/debts.inc
+++ b/modules/ding_debt/plugins/content_types/debts/debts.inc
@@ -79,6 +79,7 @@ function ding_debt_debts_form($form, &$form_state, $debts, $has_invoiced_fees, $
     $form['accept_terms'] = array(
       '#type' => 'checkbox',
       '#title' => t('I accept the terms of service'),
+      '#required' => TRUE,
     );
 
     $provider = ding_provider_get_provider_module_name('payment');
--- a/modules/ding_debt/plugins/content_types/debts/debts.inc
+++ b/modules/ding_debt/plugins/content_types/debts/debts.inc
@@ -234,4 +234,3 @@ function ding_debt_debts_perform_payment($debts, $amount, &$form_state) {
     }
   }
 }
-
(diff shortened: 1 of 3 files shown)


holt83:
I'd like to add that the books on screenshot are for test purposes and not my loans hehe :)

themes/ddbasic/sass/components/form/form-class.scss
```diff
@@ -12,6 +12,13 @@
   input[type=submit] {
     width: auto;
     margin-bottom: 20px;
+
+    &:disabled,
+    &:disabled:hover {
+      background-color: $grey;
+      color: $grey-dark;
+      cursor: not-allowed;
```
kasperg: :+1:

modules/ding_debt/plugins/content_types/debts/debts.inc
```diff
@@ -84,6 +89,11 @@ function ding_debt_debts_form($form, &$form_state, $debts, $has_invoiced_fees, $
         '#value' => t('Pay selected debts'),
         '#suffix' => '</div>',
         '#submit' => array('ding_debt_debts_form_submit_pay_selected'),
+        '#states' => array(
+          'disabled' => array(
+            ':input[name="accept_terms"]' => array('checked' => FALSE),
```
kasperg: :+1: for using `#states`.

If I recall correctly states is clientside only so in theory a user could enable the button clientside and not accept the terms of service. I do not think it is hugely important here but you could consider adding `#required`.

holt83: @kasperg good catch. In that sense the #states => disabled is only for visual purposes. I 
updated the PR with #required on checkbox.

It will show the red asteriks by the checkbox, but i don't think that matters. Also the form error message is very generic. I don't think it's worth providing a custom error message, since we don't expect users to do this.

4161-accept-required

holt83:
Rerunning behat tests..

PR #1531, opened 2019-09-25
4520: Re-adding popup javascript with drupal command

cableman:
#### Link to issue

https://platform.dandigbib.org/issues/4520

#### Description

When not added this way the Drupal main ajax framework is not loaded
and the JavaScript fails a stops execution. Using drupal_add_js
ensures that the core javascript framework is always loaded.

#### Screenshot of the result

The error when loaded in info file.

Skærmbillede 2019-09-25 kl  14 31 57

#### Checklist

No comments.


cableman:
Better solution in https://github.com/ding2/ding2/pull/1535

PR #1532, opened 2019-09-30
3826: Use pin-code during the SSO process

cableman:
#### Link to issue

https://platform.dandigbib.org/issues/3826

#### Description

This is some-what a temporally hack to get around issues with pre-authenticate in FBS and allow all types of users to use adgangsplatformen. 

#### Screenshot of the result

No screenshot.

#### Checklist

Please not that this hack requires that we disable the access token as soon as it has been used. The side effect is that the token can not be used later on to make request to the open-platform etc. This is due to security considerations and the information that the token now gives access to.

### DIFF
--- a/modules/ding_adgangsplatformen/ding_adgangsplatformen.module
+++ b/modules/ding_adgangsplatformen/ding_adgangsplatformen.module
@@ -334,6 +334,18 @@ function ding_adgangsplatformen_callback() {
       $user_info = $response->getBody()->getContents();
       $user_info = drupal_json_decode($user_info);
 
+      //
+      // HACK: this invalidates the access token and should be removed later on.
+      //       This only exists to get around pre-authentication issues
+      //       currently in FBS.
+      //
+      $request = $provider->getAuthenticatedRequest('DELETE', 'https://login.bib.dk/oauth/revoke', $access_token);
+      $response = $provider->getResponse($request);
+      if ($response->getStatusCode() != 200) {
+        watchdog('ding_adgangsplatformen', "https://login.bib.dk/oauth/revoke returned status code: %code", array('%code' => $response->getStatusCode()), WATCHDOG_CRITICAL);
+      }
+      unset($_SESSION['oauth2token']);
+
       if (module_exists('ding_registration') && ding_registration_is_registration_request()) {
         // This is an self registration request, so we store the information in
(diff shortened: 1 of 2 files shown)


cableman:
The scrutinizer errors is due to the LAGER code comments that are made that way to make us remember to remove the hacks later on.

modules/ding_adgangsplatformen/ding_adgangsplatformen.module
```diff
@@ -334,6 +334,18 @@ function ding_adgangsplatformen_callback() {
...
+      // HACK: this invalidates the access token and should be removed later on.
+      //       This only exists to get around pre-authentication issues
+      //       currently in FBS.
+      //
+      $request = $provider->getAuthenticatedRequest('DELETE', 'https://login.bib.dk/oauth/revoke', $access_token);
+      $response = $provider->getResponse($request);
+      if ($response->getStatusCode() != 200) {
+        watchdog('ding_adgangsplatformen', "https://login.bib.dk/oauth/revoke returned status code: %code", array('%code' => $response->getStatusCode()), WATCHDOG_CRITICAL);
+      }
+      unset($_SESSION['oauth2token']);
```
holt83: @cableman I was looking at this and wondering why you unset the oauth2token here?

cableman: Because this is a hack and the token is valid for a very long time. So if the token is not revoked and removed and it's leaked. The end users library information include username and password can be fetch with a simple curl call.

So so long that adgangsplatformen contains the user pincode in userinfo... this is a hack.

holt83: So this also involves a temporarily hack in adgangsplatformen that will return pincode?

cableman: Yes.

This breaks the whole security model behind oAuth2. So it should be removed as soon as possible. But as FBS changes don't happen over night this is the compromise decided by  DDB. The whole thing links back to the issue that FBS can't login all users types via pre-authenticate.

holt83: Ok I see :) thanks for the explanation @cableman 👍

PR #1535, opened 2019-10-01
4520: Prevent JavaScript errors in administration interface.

kasperg:
#### Link to issue

https://platform.dandigbib.org/issues/4520

#### Description

This is an alternate take on #1531 

AJAX commands in Ding Popup requires the existence of the Drupal AJAX
core library. This is not handled by the current definition in the
.info file.

Such an error can block the execution of other JavaScript code e.g.
the handling of the top menu.

To address this we declare the code in ding_popup a library and with the AJAX
library as a dependency.

We load the library is currently loaded for all HTML responses as we 
cannot establish exactly where popups are invoked.

By searching through the codebase it seems as if the AJAX command 
defined in ding_serendipity is not in use. Consequently we remove it.

#### Screenshot of the result

The error:

Moduler | Ding2 2019-10-01 16-30-30

#### Checklist

### DIFF
--- a/modules/ding_popup/ding_popup.info
+++ b/modules/ding_popup/ding_popup.info
@@ -5,5 +5,3 @@ version = "7.x-0.4"
 core = 7.x
 files[] = ding_popup.module
 files[] = ding_popup.test
-scripts[] = ding_popup-popupbar.js
-scripts[] = ding_popup.js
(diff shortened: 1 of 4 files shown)


cableman:
Code changes looks good, just minor code style detected by Scrutinizer

PR #1539, opened 2019-10-03
4568: Deep clone opensearch static cache

cableman:
#### Link to issue

https://platform.dandigbib.org/issues/4568

#### Description

When search the data-well that static cache in opensearch do not return the same result if you make the same request within the same execution.

The `collection` part of objects changes inside the static cache between calls.

#### Screenshot of the result

Cache content on first call:
Cache første kald

Cache content on second call:
Cache anden kald

#### Checklist

I'm not sure why we need to deep clone when setting the `$response` into the cache as all return statements are doing clone. But if you don't do this the object still changes between calls. I have test in both php 7.0 and 7.2 it's the same, so it don't seam to be an php error.

### DIFF
--- a/modules/opensearch/opensearch.client.inc
+++ b/modules/opensearch/opensearch.client.inc
@@ -660,17 +660,23 @@ function _opensearch_cache(TingClientRequest $request, $response = NULL) {
   // and we want to retrieve the cache value and where $value is set to NULL
   // explicitly.
   if (func_num_args() == 2) {
-    // Update the cache with the provided value.
-    $static_cache[$cid] = $response;
+    // Update the cache with the provided value. If you here just assign the
+    // response to the static variable here even though it's deep clone in the
+    // return statement changes to the collections inside the object from
+    // outside this function changes the cached object.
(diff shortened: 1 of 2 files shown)


modules/opensearch/opensearch.client.inc
```diff
@@ -660,17 +660,29 @@ function _opensearch_cache(TingClientRequest $request, $response = NULL) {
   // and we want to retrieve the cache value and where $value is set to NULL
   // explicitly.
   if (func_num_args() == 2) {
-    // Update the cache with the provided value.
-    $static_cache[$cid] = $response;
+    // Update the cache with the provided value. If you here just assign the
+    // response to the static variable here even though it's deep clone in the
+    // return statement changes to the collections inside the object from
+    // outside this function changes the cached object.
+    $static_cache[$cid] = unserialize(serialize($response));
```
kasperg: This is a repeating pattern.

I suggest implementing a `_opensearch_deep_clone($object)` function where the DocBlock can describe the reasoning behind this.

PR #1540, opened 2019-10-04
4567 by Inlead: Replace Paragraphs field with Body field in News RSS.

guddo:
#### Link to issue

https://platform.dandigbib.org/issues/4567

#### Description

The view of the Ding App Content module for the News list was re-exported after Paragraphs was added as a field and replaced with Body field in the RSS fee.

#### Checklist


guddo:
Behat is failing, but it's not related to this changeset.

PR #1541, opened 2019-10-04
4570: JS code style in webtrekk

cableman:
#### Link to issue

https://platform.dandigbib.org/issues/4570

#### Description

Minor code style changes made during JS debug.

#### Screenshot of the result

No screenshot.

#### Checklist

There is no changes in the code, only code style.

### DIFF
--- a/modules/ding_webtrekk/ding_webtrekk.js
+++ b/modules/ding_webtrekk/ding_webtrekk.js
@@ -159,11 +159,11 @@
         });
 
         $('.slick-arrow', this).once('js-ding-webtrekk').click(function() {
-          var linkId = Drupal.t('Carousel, clicked on previous link');
+          var linkId = 'Karousel, click på forrige knappen';
           var wtkId = DING_WEBTREKK_PARAMETER_CAROUSEL_PREV;
 
           if ($(this).hasClass('slick-next')) {
-            linkId =  Drupal.t('Carousel, clicked on next link');
+            linkId = 'Karousel, click på næste knappen';
             wtkId = DING_WEBTREKK_PARAMETER_CAROUSEL_NEXT;
           }
 
--- a/modules/ding_webtrekk/ding_webtrekk.js
+++ b/modules/ding_webtrekk/ding_webtrekk.js
@@ -10,7 +10,6 @@
  */
 
 (function($) {
-
   "use strict";
 
   // Webtrekk tracking parameters.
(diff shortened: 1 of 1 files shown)


modules/ding_webtrekk/ding_webtrekk.js
```diff
@@ -154,11 +159,11 @@
         });
 
         $('.slick-arrow', this).once('js-ding-webtrekk').click(function() {
-          var linkId = 'Karousel, click på forrige knappen';
+          var linkId = Drupal.t('Carousel, clicked on previous link');
           var wtkId = DING_WEBTREKK_PARAMETER_CAROUSEL_PREV;
 
           if ($(this).hasClass('slick-next')) {
-            linkId = 'Karousel, click på næste knappen';
```
kasperg: As I understand it these are strings used by administrators in Webtrekk - not end users on the site. Is there a reason why they should pass through the translation engine?

If someone changes the translation without knowing where it is sed I fear will cause unwanted changes in Webtrekk reporting.

modules/ding_webtrekk/ding_webtrekk.js
```diff
@@ -21,13 +20,13 @@
   var DING_WEBTREKK_PARAMETER_CAROUSEL_NEXT = 59;
   var DING_WEBTREKK_PARAMETER_CAROUSEL_PREV = 60;
 
-  var appendQueryParameter = function(url, key, value) {
```
kasperg: Why do we need to name our functions when we also assign them to variables?

modules/ding_webtrekk/ding_webtrekk.js
cableman: K, it's just wrong to use danish inside JS code.

modules/ding_webtrekk/ding_webtrekk.js
cableman: So when it don't work the console will write which function the error happened in a not just anonymous function.

It makes debugging easier when error message gets copied into a ticket.

modules/ding_webtrekk/ding_webtrekk.js
cableman: But, reverted back.

modules/ding_webtrekk/ding_webtrekk.js
kasperg: Thanks for elaborating. I think that is a totally valid argument. In that case: Have you considered whether you need to assign `function appendQueryParameter` to a variable at all?

I think it should be possible to declare the named functions without the assignment and then call them as you would normally.

cableman: The issue here is that they are "exported" into global "Drupal" space in line 43 (https://github.com/ding2/ding2/blob/74e40bb8b9c0d36c897d1359598dacb4d20e243d/modules/ding_webtrekk/ding_webtrekk.js#L43) if they where not I would have removed the assignment.

holt83: It's perfectly fine and normal to have a named function expression: https://kangax.github.io/nfe/#named-expr

kasperg: @holt83 I am not arguing against it.

As I see it @cableman is trying to [improve debugging for named functions](https://kangax.github.io/nfe/#names-in-debuggers) while my suggestion was then to simplify the existing code.

I see that the simplification is not possible so lets just roll with this.

holt83: @kasperg I think that is the way to go!

Maybe we could also prefix the function names with dingWebtrekk to make it more clear that this is ding_webtrekk functions when they appear in debugging. But it's a minor one.

PR #1544, opened 2019-10-08
4573: Fallback to CPR when it exists from adgangsplatformen

cableman:
#### Link to issue

https://platform.dandigbib.org/issues/4573

#### Description

Use CPR when exists as users may have both (CPR and card no.)  attached to their FBS user and both can then be used to login at  adgangsplatformen. But pre-authenticate can't use the card no.  so fallback to CPR if it's given.

#### Screenshot of the result

No screenshot.

#### Checklist

No comments

### DIFF
--- a/modules/ding_adgangsplatformen/ding_adgangsplatformen.module
+++ b/modules/ding_adgangsplatformen/ding_adgangsplatformen.module
@@ -450,7 +450,13 @@ function _ding_adgangsplatformen_provider_login(array $user_info) {
 
   try {
     $account = ding_user_authenticate(array(
-      'name' => $user_info['attributes']['userId'],
+      //
+      // HACK: Use CPR when exists as users may have both (CPR and card no.)
+      //       attached to their FBS user and both can then be used to login at
+      //       adgangsplatformen. But pre-authenticate can't use the card no.
+      //       so fallback to CPR if it's given.
+      //
+      'name' => is_null($user_info['attributes']['cpr']) ? $user_info['attributes']['userId'] : $user_info['attributes']['cpr'],
       //
       // HACK: The pincode used here should be removed later on.
       //       This only exists to get around pre-authentication issues


cableman:
The scrutinizer error is due to the "hack" comment to highlight that this should be removed later on.

modules/ding_adgangsplatformen/ding_adgangsplatformen.module
```diff
@@ -450,7 +450,13 @@ function _ding_adgangsplatformen_provider_login(array $user_info) {
...
   try {
     $account = ding_user_authenticate(array(
-      'name' => $user_info['attributes']['userId'],
+      //
+      // HACK: Use CPR when exists as users may have both (CPR and card no.)
+      //       attached to their FBS user and both can then be used to login at
+      //       adgangsplatformen. But pre-authenticate can't use the card no.
+      //       so fallback to CPR if it's given.
+      //
+      'name' => is_null($user_info['attributes']['cpr']) ? $user_info['attributes']['userId'] : $user_info['attributes']['cpr'],
```
kasperg: We now seem to have quite a few separate conditionals around single-signon or not but it seems as if we only have two cases. I think it would be preferable to reduce it to something like:

```php
      // HACK: Preauthenticate only works with CPR numbers. 
      //       Use CPR when exists as users may have both (CPR and card no.)
      //       attached to their FBS user and both can then be used to login at
      //       adgangsplatformen. But pre-authenticate can't use the card no.
      //       so fallback to CPR if it's given.
      //       This should be removed once pre-authentication issues in FBS are resolved.
if (!empty($user_info['attributes']['cpr'])) {
  $name = $user_info['attributes']['cpr'];
  $pass = '';
  $single_sign_on = TRUE;
} else {
  $name = $user_info['attributes']['userId'];
  $pass = $user_info['attributes']['pincode'];
  $single_sign_on = FALSE;
}
$account = ding_user_authenticate(array(
  'name' => $name,
  'pass' => $pass,
  'extra' => $user_info,
  'single_sign_on => $single_sign_on,
));
```

PR #1547, opened 2019-10-08
4522: Opensearch 5.2 boost configuration update

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/4522

#### Description

In opensearch 5.2 the boost weight value must now be a positive number and a decimal between 0 and 1 should now be used for down ranking. Hence we can not use element_validate_integer anymore.

Drupal's element_validate_number will validate float, but since we also
need to validate positive, this is not sufficient. Drupal does provide a
element_validate_integer_positive, but as the name suggest this only
works for integers. Hence we implement our own element validate callback
which does both.

In the current solution the decimal separator must be dot and not comma as requested in the case. The comma separator is a convention in Danish number formatting, but opensearch uses dot separator like all other systems. One solution would be to use `str_replace(',', '.', ..)`, but I do not think this is the correct approach. If we want to support this we could:

1. Add the [elements  module](https://www.drupal.org/project/elements) which provides a number HTML 5 element. In modern browsers, this element will follow the local number formatting in the display in frontend, but use the dot separator in memory and when form is submitted.
2.  Use the hack explained [here](https://drupal.stackexchange.com/a/228250), where intentionally add a space before the type parameter to avoid it being overriden in `theme_textfield()`. 

I think 2. is a hack and it produces invalid HTML (2 type attributes). So the question is if want to add a new module to support comma separator.

I will ask the business in the case if dot separator is acceptable.

#### Screenshot of the result

4522-boost-weight-element

#### Checklist

### DIFF
--- a/modules/opensearch/opensearch.module
+++ b/modules/opensearch/opensearch.module
@@ -172,6 +172,7 @@ function opensearch_boost_field_element_process($element, $form_state) {
 
   $element['weight'] = array(
     '#title' => t('Weight'),
+    '#description' => t('Must be a positive number with dot as decimal separator. Use values between 0 and 1 to downgrade.'),
     '#type' => 'textfield',
     '#default_value' => (isset($element['#value']['weight'])) ? $element['#value']['weight'] : NULL,
     '#element_validate' => array('opensearch_boost_validate_weight'),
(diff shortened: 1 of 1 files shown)


JrgenGNielsen:
reviewed and approved

PR #1549, opened 2019-10-10
4579: Added optional extra to provider login call

cableman:
#### Link to issue

https://platform.dandigbib.org/issues/4579

#### Description

When login in at eReolen we need more information from the login form and from adgangsplatformen to preform the login at the provider. This allow us to send this information via the `$user_info['extra']`in the call to `ding_user_authenticate` and forward that to the active provider.

#### Screenshot of the result

No screenshot.

#### Checklist

No comments.


JrgenGNielsen:
reviewed and approved

PR #1550, opened 2019-10-11
4580: Updated webtrekk to adgangsplatformen

cableman:
#### Link to issue

https://platform.dandigbib.org/issues/4580

#### Description

Webtrekk was using gatewayf code that have been removed from the code base.

#### Screenshot of the result

No screenshot.

#### Checklist

No comments


holt83:
Yes, but it did check that the module was present before calling the code :)

cableman:
yes, it did :-D

JrgenGNielsen:
reviewed and approved

PR #1552, opened 2019-10-13
4555: Fix reservation webtrekk event on teaser view

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/4555

#### Description

Ding Webtrekk was failing to attach tracking event in hook_entity_view() because reservation (and also see online) button was added manually in theme layer. This was probably done because only a some ding_entity buttons is wanted in teaser views and not all returned by ding_entity_buttons hook.

I think the correct approach is to add the buttons in field config as with other view modes, so that other module have a chance to know they are there. Then any unwanted stuff must then be hanlded in the theme layer after this.

#### Screenshot of the result

#### Checklist

### DIFF
--- a/modules/ding_ting_frontend/ding_ting_frontend.features.field_instance.inc
+++ b/modules/ding_ting_frontend/ding_ting_frontend.features.field_instance.inc
@@ -172,7 +172,7 @@ function ding_ting_frontend_field_default_field_instances() {
         'label' => 'above',
         'settings' => array(),
         'type' => 'hidden',
-        'weight' => 11,
+        'weight' => 12,
       ),
       'search_result' => array(
         'label' => 'above',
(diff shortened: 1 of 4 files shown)


modules/ding_ting_frontend/ding_ting_frontend.features.field_instance.inc
```diff
@@ -1175,7 +1179,7 @@ function ding_ting_frontend_field_default_field_instances() {
           'prefix_type' => 'no',
         ),
         'type' => 'ting_title_default',
-        'weight' => 4,
+        'weight' => 3,
       ),
       'search_result' => array(
         'label' => 'hidden',
```
holt83: There are a lot of unrelated weight chances, but this is how features exports the config after I changed, so I think it's best to keep it like this?

JrgenGNielsen:
reviewet og godkendt

holt83:
Rebased and fixed merge conflicts

PR #1553, opened 2019-10-16
4582: Ensure we use correct field to get phone number

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/4582

#### Description

Before every option not found on event node that was not 'name' was
incorrectly using the address field to look for phone number on library
node. Therefore the phonenumber was never actually send to p2b API, but
instead the translated value of 'None'.

This PR changes the logic so that we used the correct field_name. 

Also ensure that we send empty string if we can't find value as expected by the p2b API.

It looks like we have never send the phone number, regardless of whether it's set on library phone field, and p2b API have just started to react on this now. I have verified this by looking at what is send from our production site, which is still using last release:

```
[address1] => Willy Sørensens Plads 1
[address2] =>
[postal_code] => 7100
[city] => Vejle
[country] => DK
[phone] => Ingen
[name] => Vejle Bibliotek
```

And we also get the error now in production, when trying to create p2b-synced events. So something has changed in p2b API.

#### Screenshot of the result

#### Checklist

### DIFF
--- a/modules/ding_place2book/ding_place2book.module
+++ b/modules/ding_place2book/ding_place2book.module
@@ -310,7 +310,7 @@ function ding_place2book_node_presave($node) {
   try {
     $p2b = ding_place2book_instance();
     $event = $p2b->getEvent($event_maker_id, $event_id);
-    if (!$event->passive) {
+    if (empty($event->passive)) {
       $node_wrapper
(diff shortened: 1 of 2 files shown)


kasperg (approved):
:+1: LGTM

PR #1561, opened 2019-11-06
4599: Fix margin on right column

yepzdk:
#### Link to issue

https://platform.dandigbib.org/issues/4599

#### Description

Change margin on right column

#### Screenshot of the result

Screenshot 2019-11-06 15 01 08

#### Checklist

### DIFF
--- a/themes/ddbasic/sass/components/node/news.scss
+++ b/themes/ddbasic/sass/components/node/news.scss
@@ -492,7 +492,7 @@
   &.alternative-layout-full {
     .ding-news-right {
       float: right;
-      margin-right: 8%;
+      margin-right: getColumn(1);
 
       // Tablet
       @include media($tablet) {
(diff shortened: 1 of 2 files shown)


themes/ddbasic/sass/components/node/news.scss
```diff
@@ -492,7 +492,7 @@
   &.alternative-layout-full {
     .ding-news-right {
       float: right;
-      margin-right: getColumn(2) + getGutter(12);
+      margin-right: 8%;
```
kasperg: This change replaces a computed value related to the setup of the layout with a fixed - although percentage-based value.

Have you checked whether is possible to use the existing setup to generate the correct value?

yepzdk: I could not get the exact value for the width that the PM described. But I have now changed it to use the getColumn mixin since we prefer this approach, and I don't think the PM will notice the small width difference.

cableman:
Special circumstances have granted me permission to accept this PR.

PR #1566, opened 2019-11-08
4612: Improve redirection to /user/me from /user/uid in urls

kasperg:
#### Link to issue

https://platform.dandigbib.org/issues/4612

#### Description

This prevents uids from being picked up in trackers which can lead
to linking of user identity to specific user behavior.

Before this change users could access urls in the /user/uid format.
If not logged in then this url would display “page not found”. If the
user was logged in then this url would work just as well as the
preferred /user/me format.

To avoid these urls from being tracked we redirect immediately to
the frontpage for user pages without access and for users seeing
their own profile then redirect to /user/me.

Uids will still show up in urls where an administrative user accesses
another user but that is acceptable.

The change needed here can cause a lot of redirect trouble. I have 
tried to test the following combinations:

Active user / Viewed user  | Another user | Another editor | Another admin | The current user
-- | -- | -- | -- | --
**Anonymous** | Redirect to frontpage | Show profile with alias | Redirect to frontpage |  -
**User** | Redirect to frontpage | Show profile with alias | Redirect to frontpage | Redirect to /user/me
**Editor** | Redirect to frontpage | Show profile with alias | Redirect to frontpage | Show own profile with alias
**Admin** | Show profile with UID | Show profile with alias | Redirect to frontpage | Show own profile with alias

The important part here is the fact that the only user role able to see an uid is an administrator and tracking this url should not lead to any subsequent tracking of the user with the uid.

#### Checklist

Gitcop failure is caused by not being able to handle fancy quotes.

### DIFF
--- a/modules/ding_user_frontend/ding_user_frontend.module
+++ b/modules/ding_user_frontend/ding_user_frontend.module
@@ -274,18 +274,13 @@ function ding_user_frontend_build_menu($path = NULL) {
  */
 function ding_user_frontend_url_outbound_alter(&$path, &$options, $original_path) {
   global $user;
-  if (preg_match('/^user\/(\d+)\/(.*)$/', $path, $matches)) {
+  if (preg_match('/^user\/(\d+)\/?(.*)$/', $path, $matches)) {
     // To allow site administrators to edit other user they have to be filter
     // out of this alter. When accessing another drupal user account.
-    if ($user->uid == $matches[1] && !(in_array("administrators", $user->roles) || in_array("local administrator", $user->roles))) {
+    if ($user->uid == $matches[1]) {
       // This was not the above case, so we change the path to use "me".
-      $path = 'user/me/' . $matches[2];
+      $path = !empty($matches[2]) ? 'user/me/' . $matches[2] : 'user/me';
     }
-  }
-
-  // Fix user path (only for better paths to the user) for provider users.
-  if (preg_match('/^user\/(\d+)$/', $path, $matches) && $user->uid == $matches[1]) {
-    $path = 'user/me';
   }
 }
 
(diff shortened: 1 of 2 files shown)


GitCop:

 * Commit: bede5c0376e84817b9e95c6e5a310414d3eff65c

---

GitCop:

 * Commit: bede5c0376e84817b9e95c6e5a310414d3eff65c

---

GitCop:

 * Commit: c897a8656d932cecbb0f381bb1d7dff16cca5dbd

---

GitCop:

 * Commit: c897a8656d932cecbb0f381bb1d7dff16cca5dbd

---

GitCop:

 * Commit: 515bf909a5fbf523d749970391eb8b7c469cb8ef

---

GitCop:

 * Commit: 515bf909a5fbf523d749970391eb8b7c469cb8ef

---

modules/ding_user_frontend/ding_user_frontend.module
```diff
@@ -296,6 +291,9 @@ function ding_user_frontend_url_outbound_alter(&$path, &$options, $original_path
  * ding_user_frontend_build_uniform_menu_link().
  */
 function ding_user_frontend_url_inbound_alter(&$path, $original_path, $path_language) {
+  // Store the path. This allows us to check whether we modified it or not.
+  ding_user_frontend_original_path($original_path);
```
holt83: @kasperg why not use request_path() to get the original URL?

holt83: It looks like the `$path` variable passed to this function, might not even be the path the the user requested, since it will check to see if it's an aliased normal path before invoking hook: https://api.drupal.org/api/drupal/includes%21path.inc/function/drupal_get_normal_path/7.x

cableman: I hear you concerns, but it have been decided that this should be sent to test now. So please write this concerns into the issue tracker.

holt83: Nevermind :) This will work anyway. I think using `$request_path` would just have made it a bit cleaner, but if it needs to be included right now this will do.

And assuming the `$path` variable is original path in inbound_alter() will only be a "problem" if a provider user path is aliased and the only effect will be a redirect to a user/me path, which is unneeded, since the alias prob wont contain an ID.

PR #1569, opened 2019-11-15
4625 by Inlead: Adopt with recent service changes.

guddo:
#### Link to issue

https://platform.dandigbib.org/issues/4625

#### Description

This PR restores the original flow for pulling images during syndication of new content from BPI service. There is a long history for this logic to be broken and this is restored via this change-set. Images from title/list image fields are placed correspondingly during syndication along with inline images inserted into body via WYSIWYG.

#### Checklist

Relying on https://github.com/ding2/bpi-client/pull/16.

### DIFF
--- a/modules/bpi/bpi.images.inc
+++ b/modules/bpi/bpi.images.inc
@@ -44,7 +44,7 @@ function bpi_syndicate_images($type = 'ajax') {
  *
  * @ingroup forms
  */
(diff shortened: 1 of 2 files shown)


cableman:
Please look into the scrutinizer comment about watchdog: The second argument to watchdog() should not be enclosed with t()

PR #1572, opened 2019-11-20
4633: Make Adgangsplatformen handle non-privileged clients

kasperg:
#### Link to issue

https://platform.dandigbib.org/issues/4633

#### Description

Not all Adgangsplatformen clients are configured to return a CPR - 
not even as an attribute with a NULL value. The current state will 
cause such clients to fail because CPR will still be used anyway and 
generate notices because it is not defined.

This PR updates the code accordingly to make such clients work again.

#### Checklist

Behat tests fail for whatever reason. I will try rerunning them.


JrgenGNielsen:
reviewed and approved

PR #1574, opened 2019-11-22
4592 update globaleredirect

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/4592

#### Description

Global redirect uses the wrong variable when checking for maintenance mode, which has resulted in an error that was very hard to debug.

Also, if the module checks the wrong variable it will redirect in mainetenance mode which is shouldn't.

Additionally, by updating to latest version, we can remove a patch in make-file since these changes are included in the latest version.

The module contains a new configuration option to redirect comment URL's to their parent node. Since we're not using comments, we will just use the default value which is off.

#### Screenshot of the result

#### Checklist

### DIFF
--- a/project.make
+++ b/project.make
@@ -154,7 +154,7 @@ projects[geophp][subdir] = "contrib"
 projects[geophp][version] = "1.7"
 
 projects[globalredirect][subdir] = "contrib"
-projects[globalredirect][version] = "1.5"
+projects[globalredirect][version] = "1.6"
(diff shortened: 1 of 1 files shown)


JrgenGNielsen:
ding2.make has been renamed -> project.make

holt83:
> ding2.make has been renamed -> project.make

PR is rebased.

JrgenGNielsen:
reviewed and approved

PR #1582, opened 2020-01-10
4175: Don't pre-request covers

xendk:
#### Link to issue

https://platform.dandigbib.org/issues/4175

#### Description

As we're loading covers asynchronously anyway, there's no need to seed
the cache, it only slows down submission.

#### Checklist


xendk:
The test failing is on rating a material, unrelated to this change.

JrgenGNielsen:
reviewed and approved

PR #1583, opened 2020-01-10
4624: Upgrade admin_views

xendk:
#### Link to issue

https://platform.dandigbib.org/issues/4624

#### Description

Upgrades admin_views to newest version fixing security issue.

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1589, opened 2020-02-03
3290: add more specific empty messages on loan status

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/3290

#### Description

Some loaners are getting confused by the "You currently have no loans"-message that can be shown on both regular and overdue loans status. Even though one of them is not empty the message can still be show on the other status page, leading some loaners to wrongfully conclude, that they have no loans.

In this PR we provide separate and more specific translatable empty messages for regular and overdue loans.

#### Screenshot of the result

3290-overdue-loans-translation

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1591, opened 2020-02-13
4737: Added UTF8 safe functions to add_ellipsis

cableman:
#### Link to issue

https://platform.dandigbib.org/issues/4737

#### Description

When triming the title this is done without UTF8 safe functions, which results in json_encode returning FALSE and a empty result.

#### Screenshot of the result

See ticket,

#### Checklist

No comments

### DIFF
--- a/themes/ddbasic/utils.inc
+++ b/themes/ddbasic/utils.inc
@@ -131,9 +131,9 @@ function ddbasic_account_count_reservation_not_ready($account = NULL) {
  * Function to truncate strings
  */
 function add_ellipsis($string, $length = 200, $end = '...') {
-  if (strlen($string) > $length) {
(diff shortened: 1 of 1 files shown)


holt83:
I was certain that I had seen this before: https://github.com/ding2/ding2/pull/1581

Any reason why we're not using https://api.drupal.org/api/drupal/includes!unicode.inc/function/truncate_utf8/7.x?

holt83:
Another approach could be handling all this in CSS: https://css-tricks.com/almanac/properties/t/text-overflow/

JrgenGNielsen:
reviewed and approved

kasperg:
Merged with 6a57d37.

PR #1592, opened 2020-02-20
4720: Fix ting object view error when unpublishing/deleting related nodes

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/4720

#### Description

Fixes issues when nodes hosting material list paragraphs items are unpublished/deleted.

If they are deleted we will get a fatal error trying to use non-existing host entity. As described in the case, this could also happen when unpublishing nodes, but I've not been able to reproduce this. As far as I can see, the `ParagraphsItemEntity::hostEntity` uses `entity_load` without any `$conditions`, so they should be loaded regardless of `$node->status` value. Anyway if it's possible somehow, this fix protects against that too.

Also discovered that the entity was loaded twice, and we can just use the host entity loaded by `ParagraphsItemEntity::hostEntity`. The `$condition` array we passed to the manual `entity_load` call before, had no effect since [it was empty in case of entity of type `paragraphs_item`](https://github.com/ding2/ding2/blob/7.x-5.0.1/modules/ting_reference/ting_reference.module#L373)

Now we just check the status directly on the already loaded host node.

#### Error on deleted published node

4720-error-deleted-node

#### Unpulished node is shown as relation resulting in broken link

4720-no-error-node-unpublished

4720-error-when-clicking-node-unpublished

#### Checklist

### DIFF
--- a/modules/ting_reference/ting_reference.module
+++ b/modules/ting_reference/ting_reference.module
@@ -386,9 +386,15 @@ function ting_reference_field_formatter_view($entity_type, $entity, $field, $ins
           // loaded. Trying to get anything from the entity before this will
           // result in NULL.
           $host = $entity->hostEntity();
-          $entity_type = $entity->hostEntityType();
-          $host_wrapper = entity_metadata_wrapper($entity->hostEntityType(), $host);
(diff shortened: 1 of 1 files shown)


holt83:
Trying re-run on circle

holt83:
The behat test fails does not seem to be related to these changes.

PR #1593, opened 2020-02-24
Bypass Varnish cookie stripping for React user token javascript

kasperg:
#### Link to issue

#### Description

Our [current configuration in `ding2.vcl` will strip cookies from as
requests ending in `.js`](https://github.com/ding2/ding2/blob/master/ding2.vcl#L73-L76) assuming they are static files which should
always be cacheable. 

Our user token callback currently is `user.js` and is consequently also
affected by this. This results in the session cookie being stripped
and thus the callback will always reply as if the user is not logged
in.

Rename the file path of callback from `user.js` to `user-js` as that 
means the response will not be treated as a static file.

#### Checklist

Alternatively we could add `ding_react/user.js` to [our list of uncacheable paths](https://github.com/ding2/ding2/blob/master/ding2.vcl#L47-L58).

### DIFF
--- a/modules/ding_react/ding_react.module
+++ b/modules/ding_react/ding_react.module
@@ -47,7 +47,8 @@ function ding_react_menu() {
     'access arguments' => array('administer site configuration'),
     'type' => MENU_NORMAL_ITEM,
   );
-  $items['ding_react/user.js'] = [
+  // Use - instead of . to bypass ding2.vcl cookie stripping rules.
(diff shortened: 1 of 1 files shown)


GitCop:

 * Commit: a0ecc228891004d5a70fe1234a4c8e590f01c2c0

---

GitCop:

 * Commit: a0ecc228891004d5a70fe1234a4c8e590f01c2c0

---

PR #1594, opened 2020-02-26
4752: Ensure AJAX responder is loaded when using Adgangsplatformen

kasperg:
#### Link to issue

https://platform.dandigbib.org/issues/4752?next_issue_id=4748&prev_issue_id=4753#note-23

#### Description

We rely on redirect responses which will pass the user along to
Adgangsplatformen if she tries to perform an action while not logged
in.

If the CTools AJAX responder is not loaded then the responder will be
returned and loaded as a part of the response. 

Materiale | Harry Potter | - Mozilla Firefox (Private Browsing) 2020-02-26 15-58-03

This has two problems:

1. The user is not redirected to Adgangsplatformen immediately but
must try to perform the action again.
2. The user is not redirected to the correct path e.g. when trying to
make a reservation.

Being aware of this we originally intended to include the AJAX
responder with every anonymous response but a refactor broke the
hook name which should ensure this.

Rename the function `ding_page_build()` to 
`ding_adgangsplatformen_page_build()` to implement the hook properly.

Since the current code is already running in production I assume we 
have been saved by another part of the project including the file
which has subsequently been disabled/removed. Probably in the P2
modulespace.

#### Additional comments

Scrutinizer fails but because of an existing issue. I would prefer not to change the code at this point in time.

#### Checklist

### DIFF
--- a/modules/ding_adgangsplatformen/ding_adgangsplatformen.module
+++ b/modules/ding_adgangsplatformen/ding_adgangsplatformen.module
@@ -24,7 +24,7 @@ use League\OAuth2\Client\Provider\GenericProvider;
  * We don't known which pages this will happen on, so we add the script to every
  * page.
  */
-function ding_page_build() {
+function ding_adgangsplatformen_page_build() {
   if (user_is_anonymous()) {
     ctools_add_js('ajax-responder');
   }


holt83:
Looks like it was actually correctly implementing hook_init() before: https://github.com/ding2/ding2/commit/99badc99a80a9e789c737d03115e8ef960cc11cf

JrgenGNielsen:
reviewed and approved

PR #1596, opened 2020-02-28
4756: Use Adgangsplatformen authentication client when logging out

kasperg:
#### Link to issue

https://platform.dandigbib.org/issues/4756

#### Description

We currently use a separate secondary Adgangsplaformen client for
retrieving tokens without access to sensitive information. These 
tokens are suitable for use in client side applications.

However the secondary client is potentially shareable between 
installations and thus cannot be configured with a single valid 
logout url. This means that Adgangsplatform cannot redirect to the
correct site.

To address this we instead revoke the access token manually and then
generare a unique access token for an anonymous user using the
primary client.

We can then pass this token to Adgangsplatformen when logging out and
it will allow redirect to the corresponding site.

#### Checklist

### DIFF
--- a/modules/ding_adgangsplatformen/ding_adgangsplatformen.module
+++ b/modules/ding_adgangsplatformen/ding_adgangsplatformen.module
@@ -197,6 +197,7 @@ function ding_adgangsplatformen_get_configuration() {
     'singleLogout' => TRUE,
     'automaticallyAgency' => TRUE,
     'singleLogoutOrigin' => 'https://login.bib.dk/',
+    'revoke' => 'https://login.bib.dk/oauth/revoke',
   ];
 
   $config = variable_get('ding_adgangsplatformen_settings', $defaults);
@@ -559,8 +560,12 @@ function ding_adgangsplatformen_has_token_client() {
  *   Whether revoking the token succeeded or not.
  */
 function ding_adgangsplatformen_revoke_token(GenericProvider $provider, $access_token) {
-  $request = $provider->getAuthenticatedRequest('DELETE',
-    'https://login.bib.dk/oauth/revoke', $access_token);
+  $config = ding_adgangsplatformen_get_configuration();
+  $request = $provider->getAuthenticatedRequest(
+    'DELETE',
+    $config['revoke'],
+    $access_token
+  );
   $response = $provider->getResponse($request);
   $success = $response->getStatusCode() == 200;
   if (!$success) {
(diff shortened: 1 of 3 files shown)


modules/ding_adgangsplatformen/ding_adgangsplatformen.module
```diff
@@ -521,6 +547,34 @@ function ding_adgangsplatformen_has_token_client() {
...
+ *   The OAuth2 provider to use when revoking the token.
+ * @param string $access_token
+ *   The access token to revoke.
+ *
+ * @return bool
+ *   Whether revoking the token succeeded or not.
+ */
+function ding_adgangsplatformen_revoke_token(GenericProvider $provider, $access_token) {
+  $request = $provider->getAuthenticatedRequest('DELETE',
+    'https://login.bib.dk/oauth/revoke', $access_token);
```
cableman: Not super happy about the URL placement here. All the other URL's are kept together in the `ding_adgangsplatformen_get_configuration()`.

cableman: Maybe a minor rewrite of `ding_adgangsplatformen_get_configuration()` is requires as I remember that the return value is pasted directly to the oAuth library, which is not super happy about changes to the config array.

modules/ding_adgangsplatformen/ding_adgangsplatformen.module
```diff
@@ -275,20 +275,50 @@ function ding_adgangsplatformen_login() {
...
+    try {
+      // Retrieve a new anonymous token and use that for the logout procedure. By
+      // using the authentication provider we ensure that Adgangsplatformen will
+      // redirect to the right site.
+      $auth_provider = ding_adgangsplatformen_get_provider(TRUE);
+      $token = $auth_provider->getAccessToken('password', [
+        // Using @ as username and password works as an anonymous user.
+        'username' => '@',
+        'password' => '@',
+      ]);
```
cableman: I not sure here, but from usage of the open platform... should the `ting_agency` have been sent with the request here?

cableman: In the form `@775100` for both username and password? It's how we do it af the open platform at authentication to get an anonymous token for a given library?

kasperg: Not as I understand the situation. According to @vibjerg the important part here is the client id as that it is the client that knows which urls are allowed for logout redirects.

modules/ding_adgangsplatformen/ding_adgangsplatformen.module
kasperg: > All the other URL's are kept together in the ding_adgangsplatformen_get_configuration()

Apart from https://login.bib.dk/oauth/revoke in the current version, but OK - I will bite and see what can be done.

As I understand it [the OAuth library is looking for specific keys in the configuration](https://github.com/thephpleague/oauth2-client/blob/2.3.0/src/Provider/AbstractProvider.php#L95-L108). As long as we do not override these we should be fine. We already put in a lot of other options which are irrelevant to the provider.

PR #1598, opened 2020-02-28
4761: Never cache ding_react/user.js in Varnish

kasperg:
#### Link to issue

https://platform.dandigbib.org/issues/4761

#### Description

Never cache `ding_react/user.js` in Varnish.

This file contains a unique per user specific token and should not
be cached. However [our current Varnish configuration treats all files
ending in .js](http://github.com/ding2/ding2/blob/master/ding2.vcl#L74-L74) as a static file and strips cookies [no matter what
headers we send](http://github.com/ding2/ding2/blob/master/modules/ding_react/ding_react.module#L277-L277).
 
This results in the file always returning as if accessed by the
anonymous user.

To fix this we update the Varnish configuration to bypass caching of
this file.

This PR is alternative take on that is also proposed in #1593.

#### Checklist

### DIFF
--- a/ding2.vcl
+++ b/ding2.vcl
@@ -53,6 +53,7 @@ sub vcl_recv {
     req.url ~ "^.*/ajax/.*$" ||
     req.url ~ "^.*/ahah/.*$" ||
     req.url ~ "^.*/edit.*$" ||
+    req.url ~ "^.*/ding_react/user.js" ||
     req.url ~ "^.*/ding_availability.*$") {
     return (pass);
   }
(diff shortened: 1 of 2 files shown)


JrgenGNielsen:
reviewed and approved

holt83:
If using drupal_fast_404() in settings.php it's also necessary to add ding_react to exclude paths. For example:

`$conf['404_fast_paths_exclude'] = '/\/(?:styles)|(?:system\/files)|(?:ding_react)\//'; `

PR #1600, opened 2020-03-02
Core: Use menu item classes for debts, overdue loans and reservations

kasperg:
#### Description

The notification classes were a part of the P2 project and are no
longer available. Make the user menu title use existing classes which
are also used for menu items.

These do not have types per use case so we switch to the
success/warning variants that we already have.

#### Checklist


kasperg:
Closed in favour of #1601

PR #1602, opened 2020-03-05
4765: Use jQuery.attr() instead of data() in reservability checks

kasperg:
#### Link to issue

https://platform.dandigbib.org/issues/4765

#### Description

jQuery.data() will try to process the data value to an appropriate
type and this “00123” becomes 123.

That is not desirable for us and thus we fall back to using the
simpler jQuery.attr() function.

“Every attempt is made to convert the attribute's string value 
to a JavaScript value (this includes booleans, numbers, objects, 
arrays, and null) […] To retrieve a data-* attribute value as an unconverted string, use the attr() method.”
https://api.jquery.com/data/#data-html5

#### Checklist

### DIFF
--- a/modules/ding_reservation/js/ding_reservation_reservability.js
+++ b/modules/ding_reservation/js/ding_reservation_reservability.js
@@ -11,7 +11,7 @@
       var localIds = [];
       var selector = '.js-check-reservability';
       $(selector, context).once('js-check-reservability', function() {
-        localIds.push($(this).data("local-id"));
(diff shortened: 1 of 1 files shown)


GitCop:

 * Commit: e63729b54b65cf2a8b8c8d1802171e32810b9777

---

GitCop:

 * Commit: e63729b54b65cf2a8b8c8d1802171e32810b9777

---

PR #1603, opened 2020-03-05
4614: Periodical holdings based on data from library provider

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/4614

#### Description

Instead of using hook_entity_is periodical pseudo-class to decide
whether to show periodical holdings, and basing the decision on
bibliographic data, we now base the decision on the data returned
from the library provider. If there's periodical issues in the
holdings data, we show the periodical holdings.

We thereby give the library system full control over when to show
periodical holdings.

In the current approach [we only show the field if it's one of the three DKABM types 'tidsskrift', 'periodikum' or 'årbog'](https://github.com/ding2/ding2/blob/7.x-6.0.0-beta2/modules/opensearch/opensearch.module#L453)

Some issues with the current approach:
1. In FBS we can't assume that all periodicals is one of those types. This can easily be shown by searching `ma=pe not term.type any "tidsskrift periodikum årbog"`.
2. Some of the other DKABM types that have periodicals, does not only have periodicals. So we can't just add them to the list.
3. In opensearch some post have more than one type. This can give issues as show in: https://platform.dandigbib.org/issues/4716

A related issues is that we should not show normal and periodical holdings at the same time: https://platform.dandigbib.org/issues/4646. The display of normal holdings was controlled by the setting at /admin/config/ting/holdings, so before these changes you would have to ensure that all the right material types is selected except the three hardcoded periodical DKABM types. With FBS as library provider, materials is periodical if marc field 008t has value "p", which doesn't necessarily correlates with the DKABM types. So basing the decision on DKABM types doesn't really work and that goes for both the normal and periodical holdings.

Therefore this PR is also removing the holdings types configuration entirely and now the decision on whether to show normal holdings is also based on data from the library provider: if the material is considered to be part of the library catalog (library_material) we show the holdings if the provider returns any. (actually it has been like this since #1460, because of a [minor mistake](https://platform.dandigbib.org/issues/4646#note-5), but now we're going all in).

So now we have two fields basically doing the same thing and sorta fighting each other. The normal holdings will have to check holdings and ensure no periodical data is present. The periodical field will also have to fetch holdings, but have to ensure the opposite. And because of how we currently use the field groups in some cases both groups will be shown, but one of them will be empty.

To get to a less complex, better controllable situation, I have tried to merge the ding_periodical module into ding_availability, such that we have one holdings field, that handles both types of holdings display. In addition to this, I think this has the following advantages:

1. We can remove ding_periodical module and keep holdings related stuff to one module thereby reducing complexity.
2. By showing the periodical holdings in the same spot as the normal holdings, it makes them more visible to the users. Combined with the "Open"-setting for "Group holdings available" under /admin/config/ting/settings/local_config it makes the periodical holding even more accessible/visible. See screenshots below for the result.
3. Both types of holdings are handled asynchronously. Before the ding_periodical field was using a synchronous formatter (it does have an async formatter though).

We keep the ding_periodical module in the code base so that we can remove the field and disable/uninstall the module. The module can then be removed in the next release. I'm not sure if this is the best way to go about it?

I guess the business will have to decide if showing both type of holdings in the same field is a big enough issue to over weigh the advantages we get from merging them. If that is the case we might have to revert this, use ding_periodical async formatter and provide separate solution for: https://platform.dandigbib.org/issues/4646

#### Screenshot of the result

4614-holdings-local-settings-default

4614-holdings-local-settings-open

#### Checklist


holt83:
"Project admin_menu (7.x-3.0-rc5) could not be downloaded" Error on circle. Trying re-run.

JrgenGNielsen:
reviewed and approved

PR #1604, opened 2020-03-06
4772: Fix action links triggering login form for anonymous users

kasperg:
#### Link to issue

https://platform.dandigbib.org/issues/4772

#### Description

For some reason the current version uses nested behaviors: `open_login`
is defined within `ding_user_form()`. This means that `open_login` is
only attached after `open_login` has been attached.

This is unnecessary and we might just as well expand the selector 
from `open_login` to the current selector in `ding_user_form`.

#### Checklist

### DIFF
--- a/modules/ding_user_form/js/ding_user_form.js
+++ b/modules/ding_user_form/js/ding_user_form.js
@@ -24,19 +24,7 @@
         }
       }
 
-      /**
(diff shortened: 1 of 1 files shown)


JrgenGNielsen:
reviewed and approved

PR #1613, opened 2020-04-27
3659: GDPR contact form

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/3659

#### Description

1. Inserts a hardcoded privacy notice on contact form advicing users not the insert personal data.
2. Remove user cookies set by contact module (se commit message for details).

We can not prevent users from entering personal data, pin etc. and this getting send out unencrypted to various mail servers, but this should make them think twice before hitting that submit button :)

#### Screenshot of the result

3659-contact-form-privacy-notice

#### Checklist

### DIFF
--- a/modules/ding_contact/ding_contact.module
+++ b/modules/ding_contact/ding_contact.module
@@ -54,12 +54,14 @@ function ding_contact_form_contact_site_form_alter(&$form, &$form_state) {
   );
   $form['#submit'][] = 'ding_contact_contact_form_submit';
 
-  // Add a notice about not inserting personal data in the contact form. It's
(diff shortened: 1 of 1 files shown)


holt83:
PR updated to fix scrutinizer errors.

One issue remains which I think we can ignore:

`Only string literals should be passed to t() where possible`

We only pass it through `t()` to make it possible for administrators to change the link and maybe reword the text a bit, and have no need for it to be picked up by extractors.

holt83:
Updated the PR to use the same english source text as in #1626

JrgenGNielsen:
reviewed and approved

PR #1618, opened 2020-05-11
4814: Fixed loan list cache

cableman:
#### Link to issue

https://platform.dandigbib.org/issues/4814

#### Description

The loan list should be cached in session cache, but currently the cache key is stored under a not used key.

This actually stores the list in cache.

#### Screenshot of the result

No screenshot.

#### Checklist

No comments.


cableman:
The Scrutinizer errors are not related to this PR.

JrgenGNielsen:
reviewed and approved

PR #1620, opened 2020-05-12
4811: Re-add missing holdings suffix configuration

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/4811

#### Description

PR #1603 has rendered this missing, since it was added to the now
removed provider holdings form.

We re-add it to the FBS provider form, where it's perhaps better
placed, since the configuration is specific to this provider.

#### Screenshot of the result

4811-holdings-suffix-config

#### Checklist

### DIFF
--- a/modules/fbs/fbs.module
+++ b/modules/fbs/fbs.module
@@ -276,6 +276,31 @@ function fbs_settings_form() {
     '#default_value' => variable_get('fbs_holdings_branches_blacklist', array()),
   );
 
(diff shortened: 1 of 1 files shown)


holt83:
Behat erros is fixed in #1628.

Scrutinizer seems unrelated.

Setting this PR ready for review again.

JrgenGNielsen:
reviewed and approved

PR #1621, opened 2020-05-12
4614: Periodical holdings - Follow up

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/4614

#### Description

A follow up to #1603.

- Do not attempt to render holdings row or show any other availability information if provider didn't return any holdings. Instead we show an alternate empty text that can be translated. (see screenshot below).
- Only show reserved text if there was any reservable.
- Add translations and update.

#### Screenshot of the result

4614-no-holdings-library-material

Instead of:

image

#### Checklist

### DIFF
--- a/ding2.install
+++ b/ding2.install
@@ -944,3 +944,10 @@ function ding2_update_7080() {
 function ding2_update_7081() {
   ding2_translation_update();
 }
+
+/**
+ * Update translations.
+ */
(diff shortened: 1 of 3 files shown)


holt83:
Scrutinizer errors isn't releated to this PR.

JrgenGNielsen:
reviewed and approved

PR #1622, opened 2020-05-13
3583: Session expire follow up

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/3583

Follow up to: #1182

#### Description

Move session expire settings to ding_provider.

In ding2#1329 with the integration of adgangsplatformen, the Ding user
settings form was moved to ding_user_form.

This was overriding the expire settings added in this PR, which
was added to old the form.

Since ding_user module is no longer dealing with Drupal login,
which are now handled in ding_user_form and ding_adgangsplatformen
depending on login method, and the session expire functionality
should apply regardless of login method, we now move session
expire settings to ding_provider module on the general settings
form.

There's already a setting for deleting inactive provider users on
that form, so perpahps it makes sense to place it here.

#### Screenshot of the result

3583-expire-settings-ding-provider

#### Checklist


modules/ding_provider/ding_provider.module
```diff
@@ -693,4 +703,55 @@ function ding_provider_cron() {
...
+function ding_provider_expire_sessions() {
+  if (!$limit = variable_get('ding_provider_session_expire', 900)) {
+    // Return now if the session expire functionality is disabled.
+    return;
+  }
+
+  // Delete sessions with no activity since $limit seconds.
+  $timestamp = REQUEST_TIME - $limit;
+
+  // Construct subquere with uids for provider users.
```
JrgenGNielsen, line 738: typo

holt83:
Fixed typo!

JrgenGNielsen:
reviewed and approved

PR #1624, opened 2020-05-14
4640: Stop ting-object text overlay on buttons.

rasben:
#### Link to issue

https://platform.dandigbib.org/issues/4640

#### Description

See attached screenshots, they describe it best.

#### Screenshot of the result

**BEFORE**::

Screenshot 2020-05-14 at 15 03 50

Screenshot 2020-05-14 at 15 03 58

---------------------------

**AFTER**

Screenshot 2020-05-14 at 15 22 10

Screenshot 2020-05-14 at 15 22 21

#### Checklist

### DIFF
--- a/themes/ddbasic/sass/components/ting-object/ting-object.scss
+++ b/themes/ddbasic/sass/components/ting-object/ting-object.scss
@@ -59,7 +59,24 @@
         top: 0;
         left: 0;
         width: 100%;
-        height: 100%;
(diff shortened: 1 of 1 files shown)


rasben:
Scrutinizer is failing, but it's in completely unrelated files.
The only thing I've changed is a SCSS file - and scrutinizer is fine with that.

PR #1625, opened 2020-05-14
4783: update CKEditor to 4.14.0

JrgenGNielsen:
https://platform.dandigbib.org/issues/4783

Update CKEeditor JavaScript libraries to v4.14.0

The problem existed in CKEditor module for Drupal, not in JavaScript libraries with the same names, so it's not critical.

- [x ] My code passes our static analysis suite. If not then I have added a comment explaining why this change should be exempt from the code standards and process.


JrgenGNielsen:
Scrutinizer & behat errors don't seem to be related to this issue.

PR #1628, opened 2020-05-20
4826: Fix behat tests and some cleanup

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/4826

#### Description

The behat search tests was failing because it was still using 5.0 of the testsite, which is no longer available. I didn't see this at first because it was set again further down the pipeline, and not only during initial install.

This PR fixes the behat tests and does some additional cleanup:

- Use 5.2 of test site and only set opensearch variables once during initial install.
- Update autocomplete settings used in behat tests.
- Use newest version of infomedia in behat tests.
- Remove setting of p2 variables in behat tests.

#### Screenshot of the result

If your change affects the user interface you should include a screenshot of the result with the pull request.

#### Checklist

- [x] My complies with [our coding guidelines](../docs/code_guidelines.md).
- [x] My code passes our static analysis suite. If not then I have added a comment explaining why this change should be exempt from the code standards and process.
- [x] My code passes our continuous integration process. If not then I have added a comment explaining why this change should be exempt from the code standards and process.

#### Additional comments or questions

If you have any further comments or questions for the reviewer them please add them here.

PR #1629, opened 2020-05-22
4659: Do not publish news to app by default

kasperg:
#### Link to issue

https://platform.dandigbib.org/issues/4659

#### Description

We currently have a variable set which marks news for publication in
app by default.

The current default is that content is not published by default.
Unset this variable to make news follow this default. Also clean up
after Features which managed this variable previously.

#### Checklist

Behat tests involving searches currently fail. See #1628.


JrgenGNielsen:
reviewed and approved

PR #1631, opened 2020-05-26
bug4824: Update webforms to 4.22

JrgenGNielsen:
#### Link to issue

https://platform.dandigbib.org/issues/4824

#### Description

Updates Webform to 7.x-4.22

#### Checklist


JrgenGNielsen:
behat test errors, see: https://platform.dandigbib.org/issues/4826

PR #1632, opened 2020-05-26
bug4829: Update Drupal core to 7.70

JrgenGNielsen:
#### Link to issue

https://platform.dandigbib.org/issues/4829

#### Description

Drupal 7 has an Open Redirect vulnerability. For example, a user could be tricked into visiting a specially crafted link which would redirect them to an arbitrary external URL.

#### Checklist

### DIFF
--- a/project-core.make
+++ b/project-core.make
@@ -3,7 +3,7 @@ api = 2
 
 ; Core
 projects[drupal][type] = core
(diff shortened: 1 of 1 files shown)


JrgenGNielsen:
behat test errors, see: https://platform.dandigbib.org/issues/4826

PR #1633, opened 2020-05-28
4827: Align layout for action buttons

yepzdk:
#### Link to issue

https://platform.dandigbib.org/issues/4827#change-48140

#### Description

Align action buttons on material

#### Screenshot of the result

Screenshot 2020-05-25 11 57 58

#### Checklist

I see that the react component for the `Add to list` button/function does not use the class .action-button but tries to mimic the styles instead. This should be addressed in another issue since it does not follow the styleguide and it breaks in mobile view.


JrgenGNielsen:
reviewed and approved

PR #1634, opened 2020-06-01
4720: Fix error on deleted relation nodes

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/4720

#### Description

Fix error when viewing materials that has been attached to deleted nodes in paragraphs material lists.

It's only the node's field cache that is cleared when a ting_reference field is deleted. The ting_object's field_cache data will keep pointing back to paragraph/node until next cache flush, which will give error. 

It could take long before next cache flush, so in this PR we also attempt to clear the related ting_object's field cache, when ting_reference field is deleted.

To be safe we will still keep the empty-checks in the field formatter, to ensure we could load the referenced entity.

#### Screenshot of the result

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1636, opened 2020-06-03
4789: Nodelist promoted nodes link issue

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/4789

#### Description

The click handler was getting the data-href from it's outer scope, resulting in it always using the value from the first matches. In case of multiple nodelist promoted nodes on one page, this would make them all point to items in the first one.

To solve this we get the data-href directly from the clicked itemwrapper in the click handler.

#### Screenshot of the result

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1638, opened 2020-06-03
4033: Make comment module optional and disable

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/4033

#### Description

DDB CMS has very limited comment integration and the styling is currently all over the place (see screenshot below). To use it properly would require adjustments that webmasters are not able to make. It therefore makes sense to disable it as standard.

Another is issue that it's very easy to turn on comments by accident as described in the case.

The easiest approach is to make the comment module optional by removing dependencies and just disable it.

DDB CMS has some comment integration in ding_user and comment related strongarmed variables. We'll keep these, since they do not cause any issue with the comment module disabled, and if any library was actually using comments, they can just enable it again and everything will work as usual. 

#### Screenshot of the result

4033-ddbasic-comment-styling

#### Checklist

### DIFF
--- a/modules/ding_content/ding_content.info
+++ b/modules/ding_content/ding_content.info
@@ -4,7 +4,6 @@ core = 7.x
 package = Ding!
 project = ding_content
 dependencies[] = clone
-dependencies[] = comment
 dependencies[] = ctools
 dependencies[] = ctools_custom_content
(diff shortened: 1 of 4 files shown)


JrgenGNielsen:
reviewed and approved

PR #1639, opened 2020-06-04
4622: Fix carousel material display

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/4622

Also see: https://platform.dandigbib.org/issues/4562

#### Description

- Fix error in field group export.
- Fix missing reservation buttons in carousels.
- Only show "Other formats" button on entity path (explanation in commit message).
- Hide add to list button from in teaser views.

Since `hook_ding_entity_buttons()` shows all buttons, it's up to the implementer to ensure it's displayed correctly in the different view modes in frontend. Otherwise it will have to be removed in theming, like the approach in this PR.

#### Screenshot of the result

4622-result

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1648, opened 2020-06-18
4362: Fix order off manualy selected materials in Ting Smart Search

Arni:
#### Link to issue

https://platform.dandigbib.org/issues/4362

#### Description

When librarians manually select materials to boost the order wasn't always the intended one. This fixes that.


modules/ting_smart_search/ting_smart_search.module
```diff
@@ -250,28 +250,19 @@ function ting_smart_search_get_object($query) {
...
   $boosts = array();
-  $weight = 20000;
+  $weight = 100000;
 
   $boost_materials = $wrapper->field_sss_boost_materials->value();
 
-  // We insert material twice to rule out factors applied from the well.
-  // For some reason some materials get sticky in the top off the search
-  // results. In order to ensure our ranking we boost twice. A bit crasy
-  // but works.
```
kasperg, line 260: :+1: Thank goodness we can get rid of this.

PR #1650, opened 2020-06-19
4862: Added cover upload usage texts

cableman:
#### Link to issue

https://platform.dandigbib.org/issues/4862

#### Description

Adds new usage texts to cover upload that may not be changeable by the individual libraries, henc they are hardcoded by order of FDDF.

#### Screenshot of the result

No screenshot.

#### Checklist

No comments


cableman:
The Scrutinizer is hard to solve due to the wishes from FDDF about the tekst not being changeable and not translateable.

modules/ddb_cover_upload/templates/ddb-cover-upload-meta-copyright.tpl.php
```diff
@@ -4,22 +4,16 @@
  * @file
  * Template for showing meta copyright information.
  *
+ * Note: The danish text here have been decided by the steering committee to be
```
kasperg, line 7: :+1:

PR #1665, opened 2020-07-10
4762 mobilesearch null tags

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/4762

#### Description

Fix tags with NULL value in mobilesearch export.

When using `field_get_items()` it will only return the term ID. To get the name of the term we need to load the full term object with `taxonomy_term_load()`. See the changes here: https://github.com/ding2/ding2/pull/1665/commits/952f82cf45cd66ba279dd00efe58a48026f7a28b

Also fixes some potential `Invalid argument supplied for foreach` Warnings in the export code.

#### Screenshot of the result

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1667, opened 2020-08-19
1519: Fix ding_campagin_plus_search's query munging

xendk:
#### Link to issue

https://platform.dandigbib.org/issues/1519

#### Description

Spurious query errors show up on sites. Turns out that ding_campagin_plus_search does a search where it takes the existing query and wraps it in double quotes before adding it's own extesnions. But this fails horribly when the query already
contains double quotes. Change to parenthesises.

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1668, opened 2020-08-20
1519 translatable errors

xendk:
#### Link to issue

https://platform.dandigbib.org/issues/1519

#### Description

Makes user visible errors from opensearch translatable. Goes together with https://github.com/ding2/ting-client/pull/33

Different errors are categorized by ting-client into syntax or internal errors depending on error message. This PR adds generic translatable error messages for these types. Also, it only shows each error message once.

And a bit of code style cleanup.

#### Screenshot of the result

image

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1669, opened 2020-08-20
4800: Fix eresource tags styling

xendk:
#### Link to issue

https://platform.dandigbib.org/issues/4800

#### Description

Styles the tags on eresources.

#### Screenshot of the result

Before: 
image

After:
image

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1672, opened 2020-08-26
4116: Remove padding from "on the site"

xendk:
#### Link to issue

https://platform.dandigbib.org/issues/4116#change-49550

#### Description

Add `padding: 0` to the `ul` element. The default styling adds some padding for the bullets we don't use.

#### Screenshot of the result

image

#### Checklist


xendk:
Don't know why the Behat test can't find a cover, the screenshot shows a `ting_covers_placeholder` cover.

JrgenGNielsen:
reviewed and approved

PR #1673, opened 2020-08-26
4896: Remove JS hover class

xendk:
#### Link to issue

https://platform.dandigbib.org/issues/4896

#### Description

The combination of CSS hover effects and JS driven hover effects
causes odd behaviour.

Remove the JS hover effect, the CSS one is just fine.

#### Screenshot of the result

new-event-overlay-2020-08-26

#### Checklist

### DIFF
--- a/themes/ddbasic/sass/components/node/event.scss
+++ b/themes/ddbasic/sass/components/node/event.scss
@@ -16,9 +16,6 @@
     position: relative;
     float: left;
     width: 100%;
-    &.is-hovered {
-      z-index: 2;
-    }
     a {
       position: absolute;
       width: 100%;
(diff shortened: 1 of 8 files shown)


themes/ddbasic/sass/components/node/event.scss
```diff
@@ -16,9 +16,9 @@
     position: relative;
     float: left;
     width: 100%;
-    &.is-hovered {
-      z-index: 2;
-    }
+    // &:hover {
```
JrgenGNielsen, line 19: just delete, rather than outcomment

JrgenGNielsen (changes requested):
bit of nitpicking...

themes/ddbasic/sass/components/node/event.scss
xendk: Doh. Fixed.

xendk:
Test failure unrelated.

JrgenGNielsen:
reviewed and approved

PR #1674, opened 2020-08-26
4866: Add css to hide coverupload button on carousel

yepzdk:
#### Link to issue

https://platform.dandigbib.org/issues/4866

#### Description

Remove button on frontpage carousel

#### Screenshot of the result

Screenshot 2020-08-26 at 14 53 22

#### Checklist


JrgenGNielsen:
reviewed and approved

holt83:
@yepzdk Nitpicking: I think this would have been better placed in the theme layer CSS, where the other buttons is hidden.

PR #1676, opened 2020-08-27
4638: Limit width of single material

xendk:
#### Link to issue

https://platform.dandigbib.org/issues/4638

#### Description

To ensure that the overlay stays in viewport.

#### Screenshot of the result

single-overlay2

#### Checklist


xendk:
Other solution coming up.

PR #1677, opened 2020-08-27
4913 fix notices

xendk:
#### Link to issue

https://platform.dandigbib.org/issues/4913

#### Description

Fix sloppy code, comparing DateObject to timestap, and missing parameters to `ding_user_is_provider_user()`, which all causes notices.
#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1678, opened 2020-08-28
4638: Fix single material overlay

xendk:
#### Link to issue

https://platform.dandigbib.org/issues/4638

#### Description

Make the material overlay for full width materials move to the left, and not quite so much, so it should stay in
the viewport.

#### Screenshot of the result

image

#### Checklist


xendk:
Behat failure er opensearch der ikke lige gad returnere noget resultat.

JrgenGNielsen:
reviewed and approved

PR #1679, opened 2020-08-28
4493: Make search help translatable

xendk:
#### Link to issue

https://platform.dandigbib.org/issues/4493

#### Description

Replaces custom content with a content type that translates strings.

#### Screenshot of the result

image

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1680, opened 2020-08-31
648: Ensure no usernames in title on pages under /user

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/648

#### Description

Ensure that a title without username is always set on pages under /user.

If no title is set the system will end up calling [menu_get_active_title()](https://api.drupal.org/api/drupal/includes%21menu.inc/function/menu_get_active_title/7.x) exposing the username in the <title>-tag on the page.

There was a previous attempt at solving this in `ding_user_preprocess_html()`, but since the title variable is handled in `template_preprocess_html()` this has no effect. Consequently we move the logic to `hook_build_page()` instead as it runs earlier. Note that I have moved all our preprocess logic to page_build, since it gives the same result in this situation and we can avoid an extra hook implementation.

Also adds title to known pages vi metatag_panels page manager integration.

#### Screenshot of the result

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1682, opened 2020-09-03
4886: Limit related materials to those available at the local library

xendk:
#### Link to issue

https://platform.dandigbib.org/issues/4886

#### Description

Limit related materials to ones available at the local library.

Depends on https://github.com/danskernesdigitalebibliotek/ddb-react/pull/109 being merged.

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1683, opened 2020-09-03
4854: Clean up opening hours for deleted nodes

xendk:
#### Link to issue

https://platform.dandigbib.org/issues/4854

#### Description

Add patch to make opening_hours clean up opening hours when nodes are deleted. Has an update hook that cleans out orphaned opening hours.

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1685, opened 2020-09-15
4803 rolltab a11y

xendk:
#### Link to issue

https://platform.dandigbib.org/issues/4803

#### Description

Add a play/pause button to rolltabs. And lot of cleanups. 
Makes ding_nodelist use the JS from ding_tabroll, rather than an old copy.
Removes unused ids that would cause invalid HTML with multiple tabrolls.
Removes jquery-ui-tabs-rotate, and implements rotation using an interval. Pausing when user clicks on a tab was virtually impossible with rotate.
Simplifies focus handling to only use one set of handlers.
Fixes nodelist rolltab tablet tab display.
Fixes nodelist rolltab mobile tab display.
Removed the select element meant for mobile, it was completely hidden by https://github.com/ding2/ding2/commit/4caca3c84360b3a68ff3dbc0867773e27ad20872 ( https://platform.dandigbib.org/issues/1685 ), thus unnessecary.

#### Screenshot of the result

Rolltab with button:
image

Buggy nodelist tablet tabs:
node-list-tabroll-tablet
Fixed:
node-list-tabroll-tablet-fixed

Buggy nodelist mobile tabs:
node-list-tabroll-mobile
Fixed:
node-list-tabroll-mobile-fixed

#### Checklist

Ding nodelists CSS should be fixed/moved to ddbasic. It's not automatically generated in a docker setup (and I wasn't willing to take the pain of trying to get a compass container running), and css-generate only takes ddbasic CSS into account (so I had to save the theme settings page after each change for it to be visible).


xendk:
Test failure is unrelated.

JrgenGNielsen:
reviewed and approved

PR #1686, opened 2020-09-18
4810: Add underline to body links

yepzdk:
#### Link to issue

https://platform.dandigbib.org/issues/4810

#### Description

Add underline to body links

#### Screenshot of the result

Screenshot 2020-09-18 at 10 26 18

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1687, opened 2020-09-18
4397: Change styling for checkboxes and radiobuttons

yepzdk:
#### Link to issue

https://platform.dandigbib.org/issues/4397

#### Description

Change the way we style checkbox and radiobuttons

#### Screenshot of the result

Screenshot 2020-09-18 at 14 41 52

#### Checklist

This is a generel solution to issues with non accessible checkbox and radio form elements.

### DIFF
--- a/themes/ddbasic/sass/base/form.scss
+++ b/themes/ddbasic/sass/base/form.scss
@@ -153,103 +153,143 @@ textarea {
 }
 
 // ==========================================================================
-// Checkbox and radiobutton
+// Checkbox, radiobutton and switch
 // ==========================================================================
 
-input[type="checkbox"],
-input[type="radio"] {
-  display: none;
-  & + label {
-    display: block;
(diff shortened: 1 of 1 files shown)


holt83:
Looks good :) 

How does it help screen readers select reservations?

yepzdk:
Thanks @holt83 
The issue with Reservations is that you could not tab to focus the checkbox needed to mark an item for deletion.
With this change we now have the default behaviour of checkbox and radio buttons, so it is possible to tab focus and toggle selection with keyboard.
The old method used display: none; witch completely hides the element from SR.

holt83:
@yepzdk Sounds good. Thanks for the explanation 👍

JrgenGNielsen:
reviewed and approved

guddo:
@yepzdk Is this on purpose? 
image

PR #1693, opened 2020-10-16
4891: Reintroduce the possibility of using submissions values token in webforms

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/4891

#### Description

Instead of always removing the token in submissions mails send by webform, we override the default template and remove the token there.

This allows to use submissions values token when using "Custom template", but if using default template submissions values will not be send.

So if the administrator choose not to use the default template, we assume that they know what they are doing and are aware of the security risks associated with sending user submitted data unencrypted in emails. 

#### Screenshot of the result

4891-webform-template-options

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1704, opened 2020-11-02
4803: Don't pause tabroll on pause button focus

xendk:
#### Link to issue

https://platform.dandigbib.org/issues/4803

#### Description

Mobile browsers leaves the focus hanging on the button after
clicking (really, desktop browsers ought to too?), which means that
the rolltab will stay paused after pressing play. Fix it but ensuring the 
handlers aren't added to the play/pause button.

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1707, opened 2020-11-04
4908: Giving permission to local administrators and local editors to use smart search form

arniloftsson:
#### Link to issue

https://platform.dandigbib.org/issues/4908

#### Description

Adding permission to to local administrators and local editors to use smart search form. Fixed bug where roles where hard coded into the ting smart search module.


kasperg (approved):
:+1: LGTM

PR #1709, opened 2020-11-04
4976: Fix related materials carousel error

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/4976

#### Description

The getAudience() method can return FALSE if there's no data, giving a fatal error on the example presented in the case, because of the type hint.

Seems instance of OpenSearchTingObject doesn't follow the doc of TingObjectInterface and can both return array or false where
as the doc for the interface says string is returned. We therefore cannot count on data being an array.

I suggest to change to formatting function to handle all cases and remove type hint.

#### Screenshot of the result

#### Checklist

### DIFF
--- a/modules/ding_react/plugins/content_types/related_materials.inc
+++ b/modules/ding_react/plugins/content_types/related_materials.inc
@@ -83,7 +83,7 @@ function ding_react_related_materials_content_type_render($subtype, $conf, $pane
 /**
  * Format a set of string values to a format suitable for React app data.
  *
- * @param mixed $data
+ * @param string|string[]|FALSE $data
  *   The data to format. Either string or array.
  *
  * @return string
--- a/modules/ding_react/plugins/content_types/related_materials.inc
+++ b/modules/ding_react/plugins/content_types/related_materials.inc
@@ -62,10 +62,10 @@ function ding_react_related_materials_content_type_render($subtype, $conf, $pane
     $sources = explode(',', $sources);
 
     $data = [
-      'subjects' => ding_react_related_materials_format_strings($object->getSubjects()),
-      'categories' => ding_react_related_materials_format_strings($object->getAudience()),
-      'sources' => ding_react_related_materials_format_strings($sources),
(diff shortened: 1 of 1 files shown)


holt83:
@kasperg I noticed that circle test are not started lately when I make a PR. Any suggestions?

holt83:
@kasperg seems like I got it working. Nevermind :)

modules/ding_react/plugins/content_types/related_materials.inc
```diff
@@ -83,15 +83,21 @@ function ding_react_related_materials_content_type_render($subtype, $conf, $pane
 /**
  * Format a set of string values to a format suitable for React app data.
  *
- * @param string[] $strings
- *   The string values.
+ * @param mixed $data
+ *   The data to format. Either string or array.
```
kasperg, line 87: If we are reacting to the signature of [`getAudience()`](https://github.com/ding2/ding2/blob/df300c08f29d4d05b0207c3304959ecc51cd4943/modules/ting/src/TingObjectInterface.php#L101-L109) and other functions I think it would be reasonable to document the types accepted to be `string|string[]|FALSE` instead of `mixed`.

modules/ding_react/plugins/content_types/related_materials.inc
```diff
@@ -62,10 +62,10 @@ function ding_react_related_materials_content_type_render($subtype, $conf, $pane
     $sources = explode(',', $sources);
 
     $data = [
-      'subjects' => ding_react_related_materials_format_strings($object->getSubjects()),
-      'categories' => ding_react_related_materials_format_strings($object->getAudience()),
-      'sources' => ding_react_related_materials_format_strings($sources),
-      'exclude-title' => ding_react_related_materials_format_string($object->getTitle()),
```
kasperg, line 68: :+1: you are right in changing this as well as `getTitle()` may return `FALSE`.

kasperg (changes requested):
:+1: I have one minor comment that I would like see addressed.

kasperg (approved):
:+1: LGTM

PR #1710, opened 2020-11-05
4985: Provide better default values for subsearch triggers

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/4985

#### Description

Provide better default values for subsearch triggers.

#### Screenshot of the result

#### Checklist


kasperg (commented):
The ticket mentions that the position for bibliotek.dk suggestions should be after - not before. That does not seem to be reflected in the changes?

holt83:
Good catch :) The PR is updated to change default value to after

kasperg (approved):
:+1: LGTM

PR #1711, opened 2020-11-05
4969: Fix periodical reservation issue

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/4969

#### Description

The changes from #1603 has resulted in some potential errors when making periodical reservations. Since periodical holdings is now pulled in async via AJAX, we now have to take extra care that everything is done right, since these holdings can now contain 'use-ajax' - reservation links:

1. Ensure Drupal AJAX script is loaded when holdings field is rendered on initial page request. We cannot do this in the AJAX-callback.
2. Call `drupal.attachBehaviors()` when recieving holdings HTML. This was not done before, but since holdings can now contain 'use-ajax' links it has consequences.

Additional I've removed all unneeded HTML IDs in periodical holdings HTML. For some periodicals with many issues it could result in a lot HTML IDs being posted back in reservation callback. In rare cases this could result in the error described here:

https://www.drupal.org/project/drupal/issues/2541722

#### Screenshot of the result

#### Checklist

### DIFF
--- a/modules/ding_availability/ding_availability.field.inc
+++ b/modules/ding_availability/ding_availability.field.inc
@@ -115,6 +115,11 @@ function ding_availability_field_formatter_view($entity_type, $entity, $field, $
         // If item is of type 'library materials' (where we can lookup holding
         // data), prepare to check holding on the item.
         if ($entity->is('library_material')) {
+          // Ensure that Drupal ajax is loaded. If the holdings is periodical
+          // there might be reservation buttons for individual issues, which
+          // uses 'use-ajax' links.
+          drupal_add_library('system', 'drupal.ajax');
+
           $attached['js'][] = array(
             'data' => array(
               'ding_availability_mode' => 'holdings',
--- a/modules/ding_availability/ding_availability.field.inc
+++ b/modules/ding_availability/ding_availability.field.inc
@@ -118,7 +118,7 @@ function ding_availability_field_formatter_view($entity_type, $entity, $field, $
           // Ensure that Drupal ajax is loaded. If the holdings is periodical
           // there might be reservation buttons for individual issues, which
           // uses 'use-ajax' links.
-          drupal_add_library('system', 'drupal.ajax');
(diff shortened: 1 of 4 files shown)


modules/ding_availability/ding_availability.field.inc
```diff
@@ -115,6 +115,11 @@ function ding_availability_field_formatter_view($entity_type, $entity, $field, $
         // If item is of type 'library materials' (where we can lookup holding
         // data), prepare to check holding on the item.
         if ($entity->is('library_material')) {
+          // Ensure that Drupal ajax is loaded. If the holdings is periodical
+          // there might be reservation buttons for individual issues, which
+          // uses 'use-ajax' links.
+          drupal_add_library('system', 'drupal.ajax');
```
kasperg, line 121: This should be added as an attached library for the render array instead of calling `drupal_add_library()`.

modules/ding_availability/js/ding_availability.js
```diff
@@ -114,6 +114,7 @@
         // Insert/update holding information for material.
         var holdings = Drupal.DADB[entity_id];
         $('#' + id).html(holdings.html);
+        Drupal.attachBehaviors(holdings.html);
```
kasperg, line 117: Is it correct to call `Drupal.attachBehaviors()` on the returned HTML and not on the `$('#' + id)` HTML element to update?

kasperg (changes requested):
:+1: Well done.

My main gripe is regarding the handling of the library.

modules/ding_availability/ding_availability.field.inc
holt83: You're right. Fixed!

modules/ding_availability/js/ding_availability.js
holt83: Yes, passing `$('#' + id)` would be the correct here. Updated PR!

kasperg (approved):
:+1: LGTM

PR #1718, opened 2020-11-16
3916: Add event list filter not null check to views filter

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/3916

#### Description

For old events the event list filter might not be set at all. For some reason this filters them out of event feed even when mobile flag is set.

Turns out we can fix this by adding a not null check filter to second views filter group, before the filter that actually checks the value. Otherwise it must give some query error resulting in the whole row being discarded. See screenshot below.

Additionally the feature export also includes some cleanup:
1. Remove body field from news rss feed views export. This was not
done in #1540.
2. Some rules config and other export was refactored. Better keep it as
features exported it to avoid getting overriden status.

#### Screenshot of the result

3916-event-filter-not-null-check

#### Checklist

### DIFF
--- a/modules/ding_app_content_rss/ding_app_content_rss.rules_defaults.inc
+++ b/modules/ding_app_content_rss/ding_app_content_rss.rules_defaults.inc
@@ -1,7 +1,7 @@
 <?php
 /**
  * @file
- * Rules configuration.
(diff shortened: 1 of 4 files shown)


kasperg (approved):
:+1: LGTM

PR #1724, opened 2020-11-25
4983 opening hours

xendk:
#### Link to issue

https://platform.dandigbib.org/issues/4983#change-51283

#### Description

Always render the opening hours table in the ajax callback, regardless of whether there's any opening hours on the node at all. 

#### Checklist


xendk:
Behat test failures are unrelated.

kasperg:
We have had a problem with our PR environments on Platform.sh. I'll close/reopen the PR to rebuild.

PR #1725, opened 2020-11-25
5005: Clone page handler properly

xendk:
#### Link to issue

https://platform.dandigbib.org/issues/5005

#### Description

When adding terms, the new variants wasn't properly cloned, resulting in sections sharing content.

Try cloning like `page_manager_handler_clone_submit()` does it instead.

#### Checklist

### DIFF
--- a/modules/ding_sections/modules/ding_sections_term_panel/ding_sections_term_panel.module
+++ b/modules/ding_sections/modules/ding_sections_term_panel/ding_sections_term_panel.module
@@ -262,92 +262,25 @@ function ding_sections_term_panel_create_term_variant($term) {
   $variant_title = t('Sections term (@term)', array('@term' => $term->name));
   $pipeline = module_exists('panels_ipe') ? 'ipe' : 'standard';
 
-  $base_handler = page_manager_load_task_handler('term_view', '', 'term_section_panel_context');
+  // Clone handler by exporting it and evaluating the export. Seems hacky, but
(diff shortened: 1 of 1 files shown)


xendk:
Behat failure is unrelated, and the scrutinizer error is difficult to fix. Page manager itself uses eval when cloning, and avoiding it would require writing a parser that parses enough of the PHP language to turn an export string back into an object.

modules/ding_sections/modules/ding_sections_term_panel/ding_sections_term_panel.module
```diff
@@ -262,93 +262,24 @@ function ding_sections_term_panel_create_term_variant($term) {
   $variant_title = t('Sections term (@term)', array('@term' => $term->name));
   $pipeline = module_exists('panels_ipe') ? 'ipe' : 'standard';
 
-  $base_handler = page_manager_load_task_handler('term_view', '', 'term_section_panel_context');
+  $export = page_manager_export_task_handler(page_manager_load_task_handler('term_view', '', 'term_section_panel_context'));
```
cableman, line 265: This "hack" with eval seams like something that needs an inline comment! 

So this loads an pane/panels as an string encode config (like seen in features exports) an creates an new variant of that pane/panel?

cableman (changes requested):

cableman (approved):

PR #1728, opened 2020-12-06
4436 by Inlead: Top ranking UI changes.

OIOIOOOI:
#### Link to issue

https://platform.dandigbib.org/issues/4436

#### Description

1. Rename the _Amount_ header in all tables to _Amount of articles_;
2. For tables, related to library data, _Title_ heading is renamed to _Library_;
3. Provide an edit link for each article in the table which lists articles pushed by current library (rightmost);
4. Change the title of all tables to be as in the original requirement;
5. Fixed a bug where agency id 999999 was hardcoded.

#### Screenshot of the result

image

#### Checklist

- [ ] My complies with [coding guidelines](../docs/code_guidelines.md).
- [ ] My code passes static analysis suite. If not then I have added a comment explaining why this change should be exempt from the code standards and process.
- [ ] My code passes continuous integration process. If not then I have added a comment explaining why this change should be exempt from the code standards and process.

None.


kasperg:
I have closed/reopened the PR to trigger a testing environment.

modules/bpi/bpi.module
```diff
@@ -148,46 +149,90 @@ function bpi_menu() {
...
+
+/**
+ * Custom access checker for BPI actions.
+ *
+ * BPI overview:
+ * The logged in user must both have the
+ * "bpi syndicate content" permission and access to create content from BPI.
+ * BPI edit link:
+ * The logged in user must have access to the target node. This is the place
+ * where BPI id parameter is used.
```
kasperg, line 187: This function `bpi_access()` is basicly one big switch statement with no shared logic and no shared arguments. If new permissions are added the function will only grow larger.

Please split it up into multiple smaller function e.g. `bpi_access_content_syndicate()` and `bpi_access_content_edit()`.

modules/bpi/bpi.module
```diff
@@ -148,46 +149,90 @@ function bpi_menu() {
...
+    case 'bpi syndicate content':
+      // Get roles that are allowed to create content from BPI according to
+      // Workflow settings.
+      $record = NULL;
+      try {
+        $query = db_select('workflow_transitions', 'w');
+        $query->fields('w');
+        $query->condition('w.name', '_creation_to_created_bpi');
+        $record = $query->execute()->fetchAssoc();
+      }
```
kasperg, line 212: Have you considered using the API provided by the Workflow module to fetch this data? [`Workflow::getTransitions()`](https://git.drupalcode.org/project/workflow/-/blob/7.x-2.x/includes/Entity/WorkflowInterface.php#L168) seems relevant.

modules/bpi/bpi.syndicate.inc
```diff
@@ -923,18 +927,25 @@ function bpi_build_top_rank(&$form) {
...
+    ),
     '#rows' => $rows,
     '#empty' => t('No data so far.'),
   );
 
-  $topNodesSyndicatedCreatedByMe = bpi_get_top_ranked('-3 months', BPI_TOP_RANK_AMOUNT, 'syndicate', 'node', ['999999']);
-  $rows = bpi_build_top_rank_rows($topNodesSyndicatedCreatedByMe);
+  $self_agency = variable_get('ting_agency', '');
+  $topNodesSyndicatedCreatedByMyLibrary = bpi_get_top_ranked('-3 months', BPI_TOP_RANK_AMOUNT, 'syndicate', 'node', [$self_agency]);
+  $rows = bpi_build_top_rank_rows($topNodesSyndicatedCreatedByMyLibrary, '/bpi/edit');
```
kasperg, line 941: In general we use snake_case for variable names - not camelCase.

kasperg (changes requested):
Reviewed. I have a few questions - primarily regarding the changes to access controls.

modules/bpi/bpi.module
OIOIOOOI: Well, this was not added in this PR.
But, answering this, I found Workflow methods unrealiable, as it would turn out into such construct:
```
/** @var \Workflow $workflow */
  $workflow = entity_load_single('Workflow', 'bpi');
  /** @var \WorkflowTransition[] $transitions */
  $transitions = $workflow->getTransitions();
  foreach ($transitions as $transition) {
    if ('_creation_to_created_bpi' === $transition->name) {
      $record = $transition;
      break;
    }
  }
```

modules/bpi/bpi.syndicate.inc
OIOIOOOI: My bad.

modules/bpi/bpi.module
kasperg: Thanks for explaining.

kasperg (approved):
:+1: LGTM

PR #1729, opened 2020-12-07
Revert "5010: Patch ECK to avoid PDOException"

kasperg:
Behat tests are broken at the moment. This is the last PR we merged. Might this be the problem?

Reverts ding2/ding2#1726


kasperg:
That did not fix the problem.

PR #1739, opened 2021-01-13
3199: Update antibot configuration

kasperg:
#### Link to issue

https://platform.dandigbib.org/issues/3199

#### Description

This is a rebased version of the work in #1715. Tests failed in the old PR because of obsolete versions of OpenSearch.

#### Checklist


modules/ding_contact/ding_contact.strongarm.inc
```diff
@@ -15,8 +15,6 @@ function ding_contact_strongarm() {
   $strongarm->api_version = 1;
   $strongarm->name = 'antibot_form_ids';
   $strongarm->value = 'comment_node_*
-user_login
-user_pass
```
kasperg, line 19: As far as I can tell updating these values will not take effect without also reverting the feature. Please add that to an update hook.

PR #1743, opened 2021-01-28
5036: Update Media Youtube

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/5036

#### Description

Update Media Youtube to latest release to get the fix for "The YouTube video ID is invalid or the video was deleted.", when trying to embed videos from the wysiwig editor:

See: https://www.drupal.org/project/media_youtube/issues/1572550

The problem is it's not using https in the link and using HEAD mehtod, which is no longer supported.

Rather than applying a patch I think we should just update the module from 3.0 (2015) to 3.10 (2020). I've tested and it seems to work.

#### Screenshot of the result

#### Checklist

### DIFF
--- a/project.make
+++ b/project.make
@@ -244,7 +244,7 @@ projects[media_vimeo][subdir] = "contrib"
 projects[media_vimeo][version] = "2.1"
 
 projects[media_youtube][subdir] = "contrib"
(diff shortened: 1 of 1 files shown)


JrgenGNielsen:
reviewed and approved

PR #1744, opened 2021-02-01
5013: Fix overriding title

xendk:
#### Link to issue

https://platform.dandigbib.org/issues/5013

#### Description

Shouldn't be necessary, ctools should handle the override_title
elements, but apparently there's so much mucking about in the nodelist
form that it get's confused.

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1745, opened 2021-02-01
4733: Sort events by event date on tags pages

xendk:
#### Link to issue

https://platform.dandigbib.org/issues/4733

#### Description

Add event date field, an sort by it before node submitted date. As other node types doesn't have an event date, they'll get null values in that column, which makes it fall back on submitted date.

#### Screenshot of the result

image

#### Checklist


JrgenGNielsen:
reviewed and approved

JrgenGNielsen:
reviewed and approved again

PR #1747, opened 2021-02-02
5049: Update Drupal Core to 7.78

kasperg:
#### Link to issue

https://platform.dandigbib.org/issues/5049

#### Description

This addresses [SA-CORE-2021-001](https://www.drupal.org/sa-core-2021-001).

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1748, opened 2021-02-02
1370: Improve periodical sorting

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/1370

#### Description

We are currently sorting each periodical "volume", but not the issues in each volume. This PR fixes that.

Additionally we swicth to using PHP's natural sort in both cases, since we can't know how the provider keys the periodical arrays. For example FBS uses both the volume and volume-number returned by the API to form a string as key for each issue. Ding availability also makes no assumptions about what the provider uses as keys for each volume. FBS uses years but it could be anything. Since we don't know if we can sort numerically, natural sort is a good solution since it will order the numbers correct in the string: http://sourcefrog.net/projects/natsort/.

If we determine that every issue (or volume of issue) is keyed by month name, we instead sort by month name chronologically.

Also removes the unneeded hyphen prefix in the periodical holdings display when only using volume number in FBS. See screenshots below.

NOTICE: There's some mismatch between the terminology in Ding availability and FBS. Ding availability expects the provider to return an array of volumes each with a nested array of issues. But FBS module gets year, volume and volumeNumber from the API and uses year for key in the returned volumes array and volume+volumeNumber for keys in the issues array. I think it best just to leave it at that and not change anything in Ding availability API at this point.

#### Screenshot of the result

1370-fixed

1370-volume-number-prefix-fixed

**Sort by month name**:

1370-periodical-sort-by-month-name

**BEFORE THE FIX ABOVE (Notice the prefix hyphens):**

1370-unwanted-volume-number-prefix

#### Checklist

### DIFF
--- a/modules/fbs/includes/fbs.availability.inc
+++ b/modules/fbs/includes/fbs.availability.inc
@@ -253,8 +253,18 @@ function fbs_availability_holdings($provider_ids) {
             $tmp_issues[$key]['total'] += 1;
           }
           else {
+            // Generate "vol key". Both volume and volumeNumber can be empty so
+            // if one is empty just use the other or if both are present we use
+            // "volume-volumeNumber".
+            $separator = '';
+            $volume = $material->periodical->volume;
+            $volume_number = $material->periodical->volumeNumber;
+            if (!empty($volume) && !empty($volume_number)) {
+              $separator = '-';
+            }
+
             $tmp_issues[$key] = array(
-              'vol' => $material->periodical->volume . (!empty($material->periodical->volumeNumber) ? '-' . $material->periodical->volumeNumber : ''),
+              'vol' => $volume . $separator . $volume_number,
               'year' => $material->periodical->volumeYear,
               'available' => $material->available ? 1 : 0,
               'total' => 1,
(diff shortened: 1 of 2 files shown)


modules/fbs/includes/fbs.availability.inc
```diff
@@ -253,8 +253,18 @@ function fbs_availability_holdings($provider_ids) {
...
           else {
+            // Generate "vol key". Both volume and volumeNumber can be empty so
+            // if one is empty just use the other or if both are present we use
+            // "volume-volumeNumber".
+            $separator = '';
+            $volume = $material->periodical->volume;
+            $volume_number = $material->periodical->volumeNumber;
+            if (!empty($volume) && !empty($volume_number)) {
+              $separator = '-';
+            }
```
kasperg, line 264: This could be reduced to something like:

```php
$volume_data = [$material->periodical->volume, $material->periodical->volumeNumber];
$volume_string = implode('-', array_filter($volume_data);
```

kasperg (approved):
:+1: LGTM

modules/fbs/includes/fbs.availability.inc
holt83: This is much more clever, so I think we should use that ;) I'll update PR.

holt83: Updated, thanks for input 👍

JrgenGNielsen:
reviewed and approved

PR #1749, opened 2021-02-03
5061: Added support for wildcard session cache clear

cableman:
#### Link to issue

https://platform.dandigbib.org/issues/5061

#### Description

After changes to the session cache set in loans list (https://github.com/ding2/ding2/pull/1618) so loans are cached. A cache key is used with set that is not available when clearing cache. So wildcard clear is added in this PR.  

#### Screenshot of the result

No screenshot.

#### Checklist

### DIFF
--- a/modules/ding_session_cache/ding_session_cache.module
+++ b/modules/ding_session_cache/ding_session_cache.module
@@ -112,8 +112,7 @@ function ding_session_cache_get($module, $cid, $default = FALSE) {
  * @param string $cid
  *   Index data is stored under.
  * @param bool $wildcard
- *   If TRUE, the $cid argument must contain a string value and cache IDs
- *   starting with $cid are deleted.
+ *   If TRUE cache IDs starting with $cid are deleted.
  */
 function ding_session_cache_clear($module, $cid, $wildcard = FALSE) {
   cache_clear_all(_ding_session_cache_get_cid($module, $cid), 'cache_ding_session_cache', $wildcard);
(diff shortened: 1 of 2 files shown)


modules/ding_session_cache/ding_session_cache.module
```diff
@@ -111,9 +111,12 @@ function ding_session_cache_get($module, $cid, $default = FALSE) {
  *   Name of the module that the date is stored under (prefix).
  * @param string $cid
  *   Index data is stored under.
+ * @param bool $wildcard
+ *   If TRUE, the $cid argument must contain a string value and cache IDs
+ *   starting with $cid are deleted.
```
kasperg, line 116: This comment is somewhat misleading. According to the definition of `$cid` it must be a string. `NULL` is not allowed although it is supported by `cache_clear_all()`.

kasperg (changes requested):
:+1: but please fix the comment. A change should also rerun the build and hopefully address the test failure.

Scrutinizer error is not related to the change at hand.

modules/ding_session_cache/ding_session_cache.module
cableman: The comment is copied from Drupal core!

cableman: Fixed

cableman:
I have now tried to re-run the test x time and it dies with `ReferenceError: jQuery is not defined` with is not related to loans cache.

modules/ding_session_cache/ding_session_cache.module
kasperg: > The comment is copied from Drupal core!

Copy/pasting from Drupal Core does not make it right. Drupal Core does not specify that `$cid` is a string. We do. It is OK for us to set stricter requirements. Then we have to follow up on them ourselves.

kasperg (approved):

kasperg (approved):
:+1: LGTM

PR #1750, opened 2021-02-07
4877: Fix online materials [WIP]

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/4877

#### Description

Currently we have three ways of checking whether a ting object is online:

1. `$ting_object->isOnline()`: checks for the presence of `dcterms:URI`.
2. `$ting_object->is('online')`: invokes `hook_ding_entity_is('online')` and lets module decide whether the object should be consideres online. Ting module currently implements this and returns `TRUE` if `$ting_object->getOnlineUrl()` is non-empty.  The `getOnlineUrl()` method first checks if the object has an `hasOnlineAccess` relation and if it's there uses the URI from there. Otherwise it will look for `dcterms:URI` and use that if available.
3. `in_array(drupal_strtolower($object->getType()), $online_types)`: Check if the ting object's material type is one of the online types configured at /admin/config/ting/online_types.

The problem is that a different mix of the above is methods used when the system needs to check online availability giving wrong and inconsistent results:

- When search results are checked for online avail on the search pages 1. and 3. is used.
- In ting module's implementation of `hook_ding_entity_buttons()` it needs to check online avail when determining whether to show a "See online"-button. Here it uses 2. `$ting_object->is('online')`.

**Proposed solution**

The `$ting_object->isOnline()` method should be the main way to decide

#### Screenshot of the result

#### Checklist


holt83:
The error on CI behat tests does not seem related to the changes in this PR.

modules/opensearch/src/OpenSearchTingObject.php
```diff
@@ -209,38 +209,42 @@ public function getLanguage() {
    * {@inheritdoc}
    */
   public function isOnline() {
-    // Check if the material has its own URI - if so we assume it is online.
-    return !empty($this->getRecordEntry('dc:identifier', 'dcterms:URI'));
+    // Check if the material has an online URL - if so we assume it is online.
+    return !empty($this->getOnlineUrl());
```
kasperg, line 213: This may cause problems with all the work we have done with caching.

By calling `getOnlineUrl()` you may also trigger a call to `getRelations()` which may again trigger a new request to OpenSearch(). On other words we may end up in a n+1 situation where displaying a search result with 10 items causes a total of 11 requests.

It may be that this has been addressed elsewhere but I need some kind of explicit assurance that this is the case. Otherwise I fear performance regressions.

Also see cb58235.

kasperg (changes requested):
I understand the reasoning behind the approach in this PR but I need more assurance regarding the consequences before I can OK this.

modules/opensearch/src/OpenSearchTingObject.php
holt83: Hmm.. I did actually consider this (sorry for not including in description) and the reason I went ahead with it, was that I noticed that all objects was always fetched with full relation data.

See: https://github.com/ding2/ding2/blob/7.x-6.2.1/modules/opensearch/opensearch.client.inc#L380

Hence I thought that there would be no need for the extra request, since the relation data will always be there. 

I must admit I did not test this full and would like to investigate further.

I also see your work in #1269 and that your tests finds otherwise.

holt83: @kasperg 

I have investigated and believe I've found the reason for the extra requests to the well and a solution.

I was right that all search and object request are fetched will full relation data and that there are no need for the extra requests. I have verified that this will prevent the following request from being performed:

https://github.com/ding2/ding2/blob/7.x-6.2.1/modules/opensearch/src/OpenSearchTingObject.php#L76 

Since we always have relationData there's simply no need for that request.

BUT.. since we don't take care to cache object's relations in predict cache, the following code where it tries to load the relation as entity WILL result in an extra request:

https://github.com/ding2/ding2/blob/7.x-6.2.1/modules/opensearch/src/OpenSearchTingObject.php#L94 

At this point the relations data is only present on an object in the cache - not as a single relation object.

The solution is to also construct cached responses for the object's relation objects in predict cache.

holt83: Also noticed that `opensearch_get_object()` has a `$with_relations` which basicly makes no sense, since they will always be fetched with relations.

I think we should just leave it be, in case anyone as actually calling the function with that parameter?

kasperg: I have also tested this and with the latest changes then I can run the test case from #1269 without additional requests.

I agree that we should keep `opensearch_get_object()` as is for now for compatibility reasons.

kasperg (approved):
:+1: LGTM

PR #1754, opened 2021-02-11
5068: Fix bibdk subsearch

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/5068

#### Description

This PR fixes two issues with the bibdk subsearch mechanism:

1. Since opensearch tries to predict cache by putting collections
from the search result in cache, the bibdk subsearch would result
in full bibdk collections later being shown to the user. As a
solution we implement hook_opensearch_cache_key_alter() and alter
the cache key if bibdk subsearch is currently being performed.

2. We currently turn off bibdk subsearch immediatly after the
subsearch is performed. This means that collections will be
fetched without bibdk agency and profile. This actually relied
on bug 1. to work, but when it's fixed the bug manifests by
displaying empty collections. The solution is turn off bibdk
subsearch after we have fetched collections.

Also refector the code that controls the static variables into
functions to make it cleaner.

#### Screenshot of the result

#### Checklist

### DIFF
--- a/modules/ting_subsearch/ting_subsearch_bibdk/ting_subsearch_bibdk.module
+++ b/modules/ting_subsearch/ting_subsearch_bibdk/ting_subsearch_bibdk.module
@@ -132,10 +132,10 @@ function ting_subsearch_bibdk_is_bibdk_subsearch($new_value = NULL) {
  * Get the query of the last bibdk subsearch performed in the request.
  *
  * @param string $new_value
- *  Used internally to control the static variable.
+ *   Used internally to control the static variable.
  *
  * @return mixed
- *  Returns NULL if no bibdk subearch has been performed in the request.
+ *   Returns NULL if no bibdk subearch has been performed in the request.
  */
 function ting_subsearch_bibdk_get_bibdk_subsearch_query($new_value = NULL) {
   $bibdk_subsearch_query = &drupal_static(__FUNCTION__);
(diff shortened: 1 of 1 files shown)


modules/ting_subsearch/ting_subsearch_bibdk/ting_subsearch_bibdk.module
```diff
@@ -104,6 +102,49 @@ function ting_subsearch_bibdk_opensearch_pre_execute($request) {
...
+/**
+ * Get the query of the last bibdk subsearch performed in the request.
+ *
+ * @param string $new_value
+ *   Used internally to control the static variable.
+ *
+ * @return mixed
+ *   Returns NULL if no bibdk subearch has been performed in the request.
+ */
+function ting_subsearch_bibdk_get_bibdk_subsearch_query($new_value = NULL) {
```
kasperg, line 140: I do not think it is a good idea to have a function with the verb `get` in the name which accepts a value that changes the internal state.

I suggest `ting_subsearch_bibdk_subsearch_query`.

kasperg (approved):
:+1: I have a comment regarding naming that you may want to look into - but not necessarily blocking.

modules/ting_subsearch/ting_subsearch_bibdk/ting_subsearch_bibdk.module
holt83: I agree. I wasn't really satisfied with the name either. I'll update the function name to your suggestion.

PR #1755, opened 2021-02-15
4857: Purge webform submissions older than 6 months

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/4857

#### Description

Introduce Webform purge module and configure it to automatically purge webform submissions older than 180 days (~6 months).

Unfortunately a functionality was introduces that makes this problematic, so I've had to create a patch for us to be able to purge ALL submissions automatically going forward. See issue below for more info:

https://www.drupal.org/project/webform_purge/issues/3198574 

#### Screenshot of the result

webform-purge-all

#### Checklist


project.make
```diff
@@ -477,6 +477,15 @@ projects[watchdog_filtering][version] = "1.3"
...
 projects[webform][version] = "4.10"
 
+projects[webform_purge][type] = "module"
+projects[webform_purge][subdir] = "contrib"
+projects[webform_purge][download][type] = "git"
+projects[webform_purge][download][url] = "http://git.drupal.org/project/webform_purge.git"
+; Need to get this dev version since our patch is based on work not released yet.
+projects[webform_purge][download][revision] = "1ec24ff"
+; Re-add the posibility to purge all submissions.
+projects[webform_purge][patch][] = "https://www.drupal.org/files/issues/2021-02-15/webform-purge-all-submissions-3198574-1.patch"
```
kasperg, line 487: I agree with the approach suggested in the patch. So did the module maintainer and [it is now included in release 7.x-1.3 of the module](https://www.drupal.org/project/webform_purge/releases/7.x-1.3).

I would much prefer that we include this instead of keeping a patched git checkout.

kasperg (changes requested):
:+1: and great job with the patch. Please switch to the newly released version of the module.

holt83:
> 👍 and great job with the patch. Please switch to the newly released version of the module.

Thanks :)

Now using latest version of module!

kasperg (approved):
:+1: LGTM

PR #1756, opened 2021-02-18
5037: Introduce section teaser

xendk:
#### Link to issue

https://platform.dandigbib.org/issues/5037

#### Description

Add a teaser field to sections for rendering in section groups.

#### Screenshot of the result

image

#### Checklist


holt83:
Nice looks useful :)

Was wondering: any reason the section (and fields) is not exported with features?

xendk:
Ask the original developer. :shrug: 

I've mentioned that it would be smart to get that fixed on https://platform.dandigbib.org/issues/4276

cableman (changes requested):
It seams that the new field only is installed in an update hook and not on new sites?

Should the field also have been added to `_ding_sections_create_fields`?

cableman (approved):

PR #1759, opened 2021-02-21
4098: Append "Logged-in" to head title

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/4098

#### Description

Append "Logged-in" to head title if `user_is_logged_in()`.

Since we use metatag module to manage head title tag, it makes sense to use its API to do this: `hook_metatag_metatags_view_alter()`

Also enable metatags for `ding_frontpage` page manager to ensure correct ouput.

#### Screenshot of the result

4098-logged-status-in-title

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1760, opened 2021-02-22
4869: Expose smart search records

arniloftsson:
#### Link to issue

https://platform.dandigbib.org/issues/4869

#### Description

Exposes smart search records as a json array.

#### Screenshot of the result

#### Checklist

### DIFF
--- a/modules/ting_smart_search/ting_smart_search.module
+++ b/modules/ting_smart_search/ting_smart_search.module
@@ -590,7 +590,7 @@ function ting_smart_search_get_records() {
     foreach ($string_items_nids as $string_item_nid) {
       $item_boostvalues = $data[$string_item_nid]->field_custom_boost_values_p;
       if (!empty($item_boostvalues)) {
-        foreach ($item_boostvalues['und'] as $key => $field_collection_id) {
+        foreach ($item_boostvalues[LANGUAGE_NONE] as $key => $field_collection_id) {
           $fce_fields = array();
           $field_collection_entity = entity_load('paragraphs_item', $field_collection_id);
           $fce_fields = array(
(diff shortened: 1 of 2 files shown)


modules/ting_smart_search/ting_smart_search.module
```diff
@@ -550,3 +557,52 @@ function ting_smart_search_search_stats_importer($search_data) {
...
 }
+
+/**
+ * Exposes smart search records as a json array.
+ */
+function ting_smart_search_records() {
+  $data = ting_smart_search_get_records();
+  if (isset($data)) {
+    $json = json_encode($data, JSON_PRETTY_PRINT);
+    printf("<pre>%s</pre>", $json);
```
kasperg, line 568: Wrapping JSON in HTML tags should not work. Please use [`drupal_json_output()`](https://api.drupal.org/api/drupal/includes%21common.inc/function/drupal_json_output/7.x).

modules/ting_smart_search/ting_smart_search.module
```diff
@@ -19,6 +19,13 @@ function ting_smart_search_menu() {
     'access arguments' => array('configure smart search'),
     'file' => 'includes/ting_smart_search.admin.inc',
   );
+
+  $items['ting/smartsearch/smartsearchrecords'] = array(
+    'title' => 'Ting Smart Search Recors',
```
kasperg, line 24:
```suggestion
  $items['ting/smart_search/records'] = array(
    'title' => 'Ting smart search records',
```

There is no need to duplicate smart_search in multiple parts of the path. This is one way it could be simplified.

I also suggest making the title casing consistent with the other menu entry.

kasperg (changes requested):
I have a couple of comments. At least look into the JSON response.

arniloftsson:
There are a couple off scrutinizer errors. It doesn't like the use off 'und' in getting values from fields. This wasn't a part off the code I really touched. I just refactored exisiting code into a method. So the same code that passed last year now has errors.

arniloftsson:
Changed menu_path to ding/smartsearch because it was specified in the issue.

kasperg (approved):
:+1: LGTM

PR #1761, opened 2021-03-01
3437: Fix availability types for collection with online entities

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/3437

#### Description

Fixes ding availabilty types display. 

When there are online types in the collection all types will currently be displayed as online.

The main problem is in frontend where it will always use the `js-online` container if it's present.

But there was also an incorrect assumption made in backend, where all entities of a type in a collection would be considered online if the first one was.

This PR fixes both issues and also some cleanup.

Check commit and code comments for details.

This PR is closely related to #1750, so we might have som conflicts.

#### Screenshot of the result

**Before fix**

3437-availability-types-before-fix

**After fix**

3437-availability-types-after-fix

#### Checklist

### DIFF
--- a/modules/ding_availability/ding_availability.field.inc
+++ b/modules/ding_availability/ding_availability.field.inc
@@ -96,12 +96,6 @@ function ding_availability_field_formatter_view($entity_type, $entity, $field, $
       ),
     );
 
-    // Extract records' source.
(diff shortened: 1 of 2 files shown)


kasperg (approved):
:+1: LGTM. Excellent commenting in the commits and nice fixups along the way.

PR #1763, opened 2021-03-05
5078: Persist link underline on hover

yepzdk:
#### Link to issue

https://platform.dandigbib.org/issues/5078

#### Description

Persist link underline on hover

#### Screenshot of the result

Screenshot 2021-03-05 at 11 09 02

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1764, opened 2021-03-09
5073: Enforce h2 and panel titles when using IPE

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/5073

#### Description

Add some modification to panels configure forms, when they are added directly in frontend with Panels IPE.

The goal is to promote better accessibility by requiring title and nudging administrators to provide a better more specific one. Also enforce h2 tag for title.

Administrators with access to page manager backend will still have all options to omit title and use different title tag.

There is a minor issue with using required on the title field: if title field is empty the ajax button to add a carousel tab will fail silently. We therefore insert the panel's content type's default title (with ctools %title token) as default value to minimize to chance of this happening. Using the default title when title field is empty is also the current behavior.

Additionally:

- Provide some generic default titles for ting_search_carousel, nodelist and ding_section.
- Cleanup: remove old default selected panel content types in Ding IPE filter.

#### Screenshot of the result

5073-solution-screenshot

#### Checklist

### DIFF
--- a/modules/ding_content/ding_content.info
+++ b/modules/ding_content/ding_content.info
@@ -6,6 +6,7 @@ project = ding_content
 dependencies[] = clone
 dependencies[] = ctools
 dependencies[] = ctools_custom_content
+dependencies[] = ding_ipe_filter
 dependencies[] = ding_page
 dependencies[] = entity
 dependencies[] = features
--- a/modules/ding_nodelist/plugins/content_types/ding_nodelist.inc
+++ b/modules/ding_nodelist/plugins/content_types/ding_nodelist.inc
@@ -33,7 +33,7 @@ $plugin = array(
 function ding_nodelist_content_type_render($subtype, $conf, $args, $context) {
   $block = new stdClass();
 
-  $block->title = t('Nodelist');
+  $block->title = t('Latest');
 
   // Prepare classes for widget container.
   $classes = array();
--- a/modules/ding_nodelist/plugins/content_types/ding_nodelist.inc
+++ b/modules/ding_nodelist/plugins/content_types/ding_nodelist.inc
@@ -33,6 +33,8 @@ $plugin = array(
 function ding_nodelist_content_type_render($subtype, $conf, $args, $context) {
   $block = new stdClass();
 
+  $block->title = t('Nodelist');
+
   // Prepare classes for widget container.
   $classes = array();
   $classes[] = 'ding_nodelist';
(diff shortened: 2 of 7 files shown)


holt83:
I discovered another error while working on this:

For the ting_search_carousel panel pane, if the "Add one more" AJAX-button is used once when filling the form, the value entered in the title tag will be ignored.

The only way to update the title tag for a ting search carousel is to change the title and submit without adding any carousel tabs.

After investigating it appears to be beacuse the ctools wizard form is not rebuild correctly in the AJAX callback, since ctools fiddles with the form id, hence Drupal is not calling the correct form builder. 

This results in the `ctools_content_configure_form_defaults_submit()`, which takes care of updating the title, not being added as submit handler to the form.

It could be fixed by adding it manually, but I choose not to address it in this PR, since I think it will be better to investigate this in a separate case.

modules/ding_content/ding_content.info
```diff
@@ -6,6 +6,7 @@ project = ding_content
 dependencies[] = clone
 dependencies[] = ctools
 dependencies[] = ctools_custom_content
+dependencies[] = ding_ipe_filter
```
kasperg, line 9: :+1: For moving this dependency.

modules/ding_nodelist/plugins/content_types/ding_nodelist.inc
```diff
@@ -33,6 +33,8 @@ $plugin = array(
 function ding_nodelist_content_type_render($subtype, $conf, $args, $context) {
   $block = new stdClass();
 
+  $block->title = t('Nodelist');
```
kasperg, line 36: I recognize what you are doing here using default pane titles as the default values to avoid validation errors. That being said I am not sure that these somewhat technical titles (patrons should not know what a nodelist is) that relate directly to the corresponding module is helpful.

If editors place a pane without changing the title the titles will leak - although only for a short while, 

If we translate the concepts in the module then the translations will carry over to the default node titles and the other way around. I do not think that is helpful.

I wonder whether it would be possible to provide something which is just a bit more friendly to the end-user?

kasperg (approved):
:+1: LGTM

modules/ding_nodelist/plugins/content_types/ding_nodelist.inc
holt83: I agree with your concern and that we should come up with something more user-friendly. I've posted a comment about it in the issue on redmine.

holt83: I updated default titles to the suggestions from the issue, which I think are fine seeing as they can always be translated another title, if the library wants to do so.

PR #1769, opened 2021-03-26
4276: Prevent overlap between teaser and button in nodelist + blocks

kasperg:
#### Link to issue

https://platform.dandigbib.org/issues/4276

#### Description

You can use nodelist to display nodes in blocks. Each node has a
teaser text and read more link in the form of a button. This is shown
when the user hovers node. 

However when the truncated teaser spans the entire last line then the
read more button will overlap with the teaser text.

To prevent this we increase the size of the teaser when hovered. This
provides more space for the button which will no longer overlap with
the text.

#### Screenshot of the result

Before as seen on https://upgrade-fbs.ddbcms.dk/section/sektion:

https://user-images.githubusercontent.com/73966/112604923-04272b80-8e17-11eb-9b41-869fc1416445.mp4

After:

https://user-images.githubusercontent.com/73966/112604948-0c7f6680-8e17-11eb-8a16-65a76f4d6dd1.mp4

#### Checklist

### DIFF
--- a/modules/ding_nodelist/js/node_blocks.js
+++ b/modules/ding_nodelist/js/node_blocks.js
@@ -53,7 +53,7 @@
       $('.node-ding-news.nb-item', context).mouseenter(function() {
         var title_and_lead_height;
         // Set height for title and lead text.
-        title_and_lead_height = $(this).find('.title').outerHeight(true) + $(this).find('.field-name-field-ding-news-lead').outerHeight(true) + 50;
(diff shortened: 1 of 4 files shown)


guddo:
Please note that this only solves the problem for News CTs.

mihaelad:
For news yes but for page and eresource CTs no .

cableman (approved):

PR #1770, opened 2021-04-07
4927: Set top-level heading on section titles

rimi-itk:
#### Link to issue

https://platform.dandigbib.org/issues/4927

#### Description

Changes section titles to always be displayed as `h1` elements.

#### Checklist

This may be considered a hack, but doing it in the template is even more hackish …


JrgenGNielsen:
reviewed and approved

PR #1772, opened 2021-04-16
5080: Fix bibdk button background color

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/5080

#### Description

Since we use "text on secondary" for bibdk external search button color, we also have to use secondary color as background.

Also add better hover effect.

#### Screenshot of the result
New background color (notice default visual is changed):
5080-secondary-background-on-button

Hover effect:
5080-added-hover-effect

#### Checklist


holt83:
Behat test error does not seem related to changes at hand.

kasperg (approved):
:+1: LGTM

PR #1773, opened 2021-04-16
4101: Add translatable FBS on hold field description

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/4101

#### Description

Add english source for on hold description

This will make it translatable via the field translate module.

Note that we also need to update Token module for this to work.
Version 1.7 contains a fix to the following issue:

https://www.drupal.org/node/2474403

Since 1.8 contains additional bug fixes we might as well upgrade
to the newest version.

#### Screenshot of the result

4101-on-hold-description

4101-on-hold-description-translation-with-link

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1775, opened 2021-04-27
5129: Use h1 for debt pane headline to improve accessibility

kasperg:
#### Link to issue

https://platform.dandigbib.org/issues/5129

#### Description

Update the panel pane configuration to use H1 for the title.

For this to work we also have to update the template to respect the 
selected title tag.

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1776, opened 2021-04-27
5151: Ensure optOut cookies are removed from users browsers

holt83:
#### Link to issue

https://platform.dandigbib.org/issues/5151

#### Description

Removing the cookie from ECC whitelist is the most convenient of
achieving this without writing any frontend code to handle this
ourselves. 

It was left there to keep support of opt-in/out with cookie-based
tracking, but in current situation it's preffered to just remove
it.

#### Screenshot of the result

image

#### Checklist

### DIFF
--- a/modules/ding_webtrekk/ding_webtrekk.install
+++ b/modules/ding_webtrekk/ding_webtrekk.install
@@ -6,17 +6,26 @@
  */
 
 /**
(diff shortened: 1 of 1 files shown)


kasperg:
Work continues in #1782.

PR #1777, opened 2021-04-27
5137: Add underline to links in WYSIWYG text on debts page

kasperg:
#### Link to issue

https://platform.dandigbib.org/issues/5137

#### Description

Add underline to links in WYSIWYG text on debts page to improve accessibility.

For ancient reasons the default display of links is without
decoration. Consequently we have to add it on a per-case basis.

By adding a class ending with -body we get the decoration we want.
It is not pretty but it works.

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1778, opened 2021-04-27
5133: Map readable FBS dept type to Ding dept type

kasperg:
#### Link to issue

https://platform.dandigbib.org/issues/5133

#### Description

Reason message is a free text field where library employees can enter
whatever. This does not go well with the design. Instead use the type
field.

We expect the type field to be restricted regarding allowed values
but official documentation of actual values is nowhere to be found.
Consequently we reference types which can be found in examples like
https://platform.dandigbib.org/issues/5059 and fall back to a generic
“other” for the rest.

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1779, opened 2021-04-27
5139: Ensure debt form always has a button

kasperg:
#### Link to issue

https://platform.dandigbib.org/issues/5139

#### Description

Ensure debt form always has a button. This is required to ensure accessibility.

Adding the button role to the external payment url should be
sufficient for assistive technologies to determine its relevance.

We also have to ensure that either form of payment button is enabled.
Otherwise there will actually be no buttons. Restructuring the form
to handle such a setup will be quite tedious and probably not a
relevant usage scenario. Consequently we do not accept such a
configuration.

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1780, opened 2021-04-27
5134: Fix material item information floating on mobile

kasperg:
#### Link to issue

https://platform.dandigbib.org/issues/5134

#### Description

Fix material item information floating on mobile.

Without clearing and with sufficiently long dates as the first entry
then the third entry may creep up to the right underneath the second
one on mobile.

With the clearing odd numbered items should stay to the left.

#### Checklist


JrgenGNielsen:
reviewed and approved

kasperg:
Closed. The fixed is already included as a part of #1781 .

PR #1781, opened 2021-04-27
5136: Debt layout

kasperg:
#### Link to issue

https://platform.dandigbib.org/issues/5136

#### Description

Improve layout of material item information elements

Currently we rely on dynamic resizing based on content. This results
in an uneven rhythm where you are not guaranteed that the same info
aligns vertically across items in a list.

Introduce a shared grid where each info has the same width - 33% on
desktop and 50% on mobile. Clear accordingly to avoid overlapping
infos.

Note that the original issue deals with debts but this actually
targets all material item lists.

Also make debt totals align with other infos in material list. This goes for both spacing (margin + padding) and fonts.

#### Checklist

### DIFF
--- a/themes/ddbasic/sass/components/panel/class-panel.scss
+++ b/themes/ddbasic/sass/components/panel/class-panel.scss
@@ -282,9 +282,26 @@
   margin-bottom: 25px;
 
   @include clearfix();
+
   .total-amount {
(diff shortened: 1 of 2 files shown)


JrgenGNielsen:
reviewed and approved

PR #1787, opened 2021-05-24
4497: More changes to Redux handling

kasperg:
**Do not merge while temporary builds of React components are included!**

#### Link to issue

https://platform.dandigbib.org/issues/4497

#### Description

This PR contains a few more fixes in relation to Redux handling.

Please check the commit comments for additional detail.

This PR depends on danskernesdigitalebibliotek/ddb-react#138.

#### Checklist

### DIFF
--- a/modules/ding_react/js/ding-react.reset.js
+++ b/modules/ding_react/js/ding-react.reset.js
@@ -1,0 +1,3 @@
+// For some reason resetting will not work if run to early. Perhaps the
+// persisted store is not quite ready. Consequently we have to wait a bit.
+setTimeout(() => window.ddbReact.reset(), 500);
--- a/project.make
+++ b/project.make
@@ -639,7 +639,7 @@ libraries[smart-app-banner][destination] = "libraries"
 
 ; For ding_react.
 libraries[ddb-react][download][type] = "get"
-libraries[ddb-react][download][url] = https://github.com/danskernesdigitalebibliotek/ddb-react/releases/download/latest/dist.zip
+libraries[ddb-react][download][url] = https://github.com/danskernesdigitalebibliotek/ddb-react/releases/download/2.3.0/dist.zip
 libraries[ddb-react][directory_name] = "ddb-react"
 libraries[ddb-react][destination] = "libraries"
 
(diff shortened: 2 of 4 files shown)


project.make
```diff
@@ -639,7 +639,7 @@ libraries[smart-app-banner][destination] = "libraries"
 
 ; For ding_react.
 libraries[ddb-react][download][type] = "get"
-libraries[ddb-react][download][url] = https://github.com/danskernesdigitalebibliotek/ddb-react/releases/download/latest/dist.zip
+libraries[ddb-react][download][url] = https://github.com/reload/ddb-react/releases/download/2.3.1-beta1/dist.zip
```
cableman, line 642: Should use the code base at https://github.com/danskernesdigitalebibliotek/ddb-react

kasperg: Agreed but I need this change for the PR environment to reflect the latest changes. I have updated the text accordingly.

modules/ding_react/js/ding-react.reset.js
```diff
@@ -0,0 +1,3 @@
+// For some reason resetting will not work if run to early. Perhaps the
+// persisted store is not quite ready. Consequently we have to wait a bit.
+setTimeout(() => window.ddbReact.reset(), 500);
```
cableman, line 3: Will this not assume at all uses have "fast" browsers? What on a slow/older mobil? Could this be solved better in react in ddbReact. So it makes the user logged out checks and resets it self. I don't know the react code... so just making suggestions.

kasperg: I am not sure this is a problem. 

The comments on rt2zz/redux-persist#1015 do not mention anything about device performance and generally refer to 200 ms timeout. This has 500 ms - to be sure I guess.

In general we need some kind of signal to know when to run our code. This has to be in relation to Drupal. The React components by design do not know anything about the underlying application layer anyway.

As I see it we have two options:

1. Hijack the logout button and run `reset()` when the user clicks the button but before the serverside logout logic is called. This means that the React code will have to be added to every page compared to the current situation where it is only loaded on pages with actual components.
2. Run our code on the logout page as soon as it loads.

This PR chooses option #2 in the current form.

cableman (approved):

PR #1788, opened 2021-05-26
2175: Update example url in ting_das module

kasperg:
#### Link to issue

https://platform.dandigbib.org/issues/2175

#### Description

There is no reason to reference the developer version instead of the
production version in the text.

#### Checklist


JrgenGNielsen:
reviewed and approved (I can't imagine failing behat test got anything to do with this)

PR #1791, opened 2021-06-08
5178: JavaScript fixes

cableman:
#### Link to issue

https://platform.dandigbib.org/issues/5178

#### Description

1. Every time an ajax request fires a new click event handler is attached to links/buttons resulting in the samme handler being fire on click x times and if the handler `toggles` some thing and the number of ajax request are odd... the effect will be that click handler do not do anything.
2. The user information (sub-menu) header sometimes just show a spinner and newer gets loaded. Changing lazy pane to load on page load and not on page scroll for the header element fixes this.
3. Change reuse of $item variabel inside loops in user menu tree build

#### Screenshot of the result

No screenshot

#### Checklist

No comments

### DIFF
--- a/modules/ding_frontend/ding_frontend.pages_default.inc
+++ b/modules/ding_frontend/ding_frontend.pages_default.inc
@@ -884,6 +884,7 @@ search/node',
     $pane->uuid = '07dec5bc-64c2-4632-ac67-72598e98d159';
     $display->content['new-07dec5bc-64c2-4632-ac67-72598e98d159'] = $pane;
     $display->panels['footer'][5] = 'new-07dec5bc-64c2-4632-ac67-72598e98d159';
+
     $pane = new stdClass();
     $pane->pid = 'new-d9c1432b-77cc-432c-987a-eb64569a0b88';
     $pane->panel = 'header';
@@ -900,7 +901,7 @@ search/node',
     $pane->cache = array(
       'method' => 'lazy',
       'settings' => array(
-        'load_strategy' => 'pane-visible',
+        'load_strategy' => 'page-loaded',
         'show_spinner' => 1,
         'load_text' => '',
       ),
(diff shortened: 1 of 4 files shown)


modules/ding_user_form/js/ding_user_form.js
```diff
@@ -24,18 +24,24 @@
...
 
-        evt.preventDefault();
+        event.preventDefault();
         body.removeClass('pane-login-is-open');
         body.removeClass('overlay-is-active');
-      });
+      }
+
+      $('a.js-topbar-link-user,a.open-login', context).unbind('click', openHandler);
+      $('a.js-topbar-link-user,a.open-login', context).on('click', openHandler);
```
kasperg, line 41: I understand the reasoning behind this change but I am not that keen on the implementation.

1. [`unbind()` is deprecated in jQuery 3 and has been replaced by `off()`](https://api.jquery.com/unbind/#entry-longdesc)
2. Could we use [`one()`](https://api.jquery.com/one/) instead of rebinding to ensure that handlers are only bound once?

kasperg (approved):
:+1: I have some questions regarding the implementation.

modules/ding_frontend/ding_frontend.pages_default.inc
```diff
@@ -900,7 +901,7 @@ search/node',
     $pane->cache = array(
       'method' => 'lazy',
       'settings' => array(
-        'load_strategy' => 'pane-visible',
+        'load_strategy' => 'page-loaded',
```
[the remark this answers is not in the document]
cableman: It's simple don't trigger the visible event i lazy-panel (every time, it's on of these reload 5 times and 1 out of the five it did not fire). But debugging and ending up changing it to loaded event and it works 5 out 5 times... so

modules/ding_user_form/js/ding_user_form.js
cableman: Changed to one() :-)

PR #1793, opened 2021-06-08
5148 webform no saving ip

arniloftsson:
Link to issue
https://platform.dandigbib.org/issues/5148#change-53896

Description
Webform collects ipaddresses per default which is not in accordance with GPDR. This PR is based on this patch: https://www.drupal.org/files/issues/2019-05-07/2972651-webform-no-ip-addresses-28.patch from this issue: https://www.drupal.org/project/webform/issues/2972651 .

The patch puts a new option in admin settings off only saving ip-addresses as neede for flood protection. It also uses the core flood protection mechanism in Drupal.

The Patch has been reworked mostly because naming off test files has been changed. It also now defaults to not saving a ipaddress

#### Screenshot of the result

image

#### Checklist


kasperg (changes requested):
:+1: for using an existing implementation to solve this issue. However I do not think it is a good idea to have local patches in our repository if it can be avoided in a sensible way. Local patches makes it harder to track the reasoning behind a change in a third party module and they do not provide any value for the Drupal community at large.

As I understand it the patch contains two things:

**A reroll of the patch from https://www.drupal.org/project/webform/issues/2972651 to match newer versions of the Webform module**

I suggest uploading the rerolled patch to the issue and referencing it from our make file. This is similar to what we do for most other patches.

**A change in the settings provided by the patch to not save ip addresses by default**

This can be controlled by setting the the value of the `webform_store_ip_address` variable to `as-needed` in a feature or in an update hook. This should probably be placed in the `ding_base` module which is also responsible for other related webform configuration.

arniloftsson:
This PR is now updated. The patch is now uploaded to the issue. It failes to aply on the issue because it is made for version 7.24 and it is applied to the dev version in the test. The difference is minimal .One line in the webform.info file.

kasperg (approved):
:+1: Thanks for the update @Arni.

PR #1798, opened 2021-06-17
5109: Styling popup find-more-button

yepzdk:
#### Link to issue

https://platform.dandigbib.org/issues/5109

#### Description

In some installations the button is not visible due to a fallback to a black text color.
This pull request addresses that by adding styling to the `.find-more-button`. Also that element is not a child of `#popup-buttons` hence we have moved the styling of it.

#### Screenshot of the result

Default:
Screenshot 2021-06-15 at 12 34 20

With `:hover`:
Screenshot 2021-06-15 at 12 34 26

#### Checklist

### DIFF
--- a/themes/ddbasic/sass/components/cookie.scss
+++ b/themes/ddbasic/sass/components/cookie.scss
@@ -25,9 +25,25 @@
       position: relative;
     }
 
+    .find-more-button {
+      color: $white;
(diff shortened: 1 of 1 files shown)


JrgenGNielsen (approved):
reviewed and approved

PR #1800, opened 2021-06-19
5193: Fix section panels template being overriden

kasperg:
#### Link to issue

https://platform.dandigbib.org/issues/5193

#### Description

Ding Sections are supports to create a new panels variant for each
section page where editors can decide what panes to use for the
specific term/page. This is controlled by the term page field which
is set by default.
 
In #1756 the weight for the section tab which places
the field in the vertical tabs section was set in order to move the
section to the bottom of the section form.

Unfortunately for some reason this did not result in moving the
section down. Instead the term page field is removed from the form
entirely. This means that is does not have a value. Due to the way
the module is constructed this will instead result in editing the
general panels variant for sections.

Sections | Københavns Biblioteker | Logget ind 2021-06-19 22-05-06

This makes it seem as if new sections copy the old ones instead of
starting from scratch. Working with new sections will also affect
other sections which also accidentally use the general variant.

Skabelon for taksonomitermer | Københavns Biblioteker | Logget ind 2021-06-19 22-04-04

To address this this commits introduces the following changes:

1. Remove the #weight setting causing problems,
2. Set field order for current sites. Since the ding_sections do not
use features we have to do this manually using an update hook
combining field instances and bundle settings,
3. Set field order for new sites mirroring the changes in 2
4. Revert the ding sections general panels variant to the default
setup.

#### Screenshot of the result

Sections | Ding2 | Logget ind 2021-06-19 22-06-12

#### Checklist

### DIFF
--- a/modules/ding_sections/ding_sections.install
+++ b/modules/ding_sections/ding_sections.install
@@ -326,7 +326,7 @@ function _ding_sections_create_teaser_field() {
     'description' => 'Tekst shown in teasers.',
     'display' => [
       'default' => [
-        'label' => 'above',
+        'label' => 'hidden',
         'type' => 'text_default',
         'weight' => -5,
         'settings' => [],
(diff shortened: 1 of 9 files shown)


JrgenGNielsen:
reviewed and approved

JrgenGNielsen:
reviewed and approved

PR #1803, opened 2021-06-22
3429: Addded views ajax history module and required library

martinyde:
#### Link to issue

#### Description

Please include a short description of the suggested change and the reasoning behind the approach you have chosen.

#### Screenshot of the result

#### Checklist

- [ x ] My complies with [our coding guidelines](../docs/code_guidelines.md).
- [ x ] My code passes our static analysis suite. If not then I have added a comment explaining why this change should be exempt from the code standards and process.
- [ x ] My code passes our continuous integration process. If not then I have added a comment explaining why this change should be exempt from the code standards and process.

### DIFF
--- a/modules/ding_event/ding_event.info
+++ b/modules/ding_event/ding_event.info
@@ -40,6 +40,7 @@ dependencies[] = text
 dependencies[] = ting_reference
 dependencies[] = uuid
 dependencies[] = views
+dependencies[] = views_ajax_history
 dependencies[] = views_content
 dependencies[] = virtual_field
 dependencies[] = webform
--- a/project.make
+++ b/project.make
@@ -445,6 +445,9 @@ projects[virtual_field][version] = "1.2"
 projects[views][subdir] = "contrib"
 projects[views][version] = "3.23"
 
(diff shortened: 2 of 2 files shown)


JrgenGNielsen (changes requested):
you also need an update_hook to enable views_ajax_history

modules/ding_event/ding_event.install
```diff
@@ -259,3 +259,10 @@ Tip: Indicate different ticket prices for more than one audience group by enteri
   db_add_field('field_data_field_ding_event_price', 'field_ding_event_price_format', $field);
   db_add_field('field_revision_field_ding_event_price', 'field_ding_event_price_format', $field);
 }
+
+/**
+ *  Enable views_ajax_history module.
```
JrgenGNielsen, line 264: Please fix the Scrutinizer error

JrgenGNielsen (changes requested):

modules/ding_event/ding_event.install
martinyde: Fixed

PR #1806, opened 2021-06-28
5127+5145: Update Drupal Core to 7.81 with SameSite cookie fallback

kasperg:
#### Link to issue

https://platform.dandigbib.org/issues/5127
https://platform.dandigbib.org/issues/5145

#### Description

Update Drupal Core to 7.81 to address latest security issues
including SA-CORE-2021-002.

Drupal 7.80 added core support for SameSite cookie handling. This
means that our current core patch no longer apply.

However to keep support for legacy PHP versions and legacy browsers
we also introduce the SameSite Cookie module and configure it
to replicate the current situation.

The module uses another variable name for overriding the SameSite
cookie value. Consequently we have to update handling it tests and
documentation.

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1811, opened 2021-08-02
5211: Fix material-item styling

xendk:
#### Link to issue

https://platform.dandigbib.org/issues/5211

#### Description

Remove ul width, as it seems to be the problem.

#### Screenshot of the result

image

#### Checklist


xendk:
Behat failure seems unrelated.

JrgenGNielsen (approved):
reviewed and approved

PR #1814, opened 2021-08-17
5046: Use recommendation service for related materials

xendk:
#### Link to issue

https://platform.dandigbib.org/issues/5046

#### Description

Use the PAPPS service instead of crafted search.

#### Screenshot of the result

ddb-5046

#### Checklist


modules/ding_react/ding_react.module
```diff
@@ -419,6 +420,16 @@ function ding_react_cover_service_url() {
   return variable_get('ding_react_cover_service_url', DING_REACT_COVER_SERVICE_URL);
 }
 
+/**
+ * Returns the url to the instance of the related materials service to use.
+ *
+ * @return string
+ *   Url to related materials service instance.
+ */
+function ding_react_related_materiols_service_url() {
```
JrgenGNielsen, line 429: is the "materiols" spelling error on purpose?

JrgenGNielsen (changes requested):

modules/ding_react/ding_react.module
xendk: Sadly no, fixed.

JrgenGNielsen (approved):

PR #1817, opened 2021-08-27
5228 search button mobile

arniloftsson:
#### Link to issue

https://platform.dandigbib.org/issues/5228?next_issue_id=5226

#### Description

Search was not shoving when using extended search profile on mobile on pages other than frontpage. The reason was that the search button has different classes on mobile vs. desktop so the code only handled the desktop version. This adds handling to the mobile version.

#### Screenshot of the result

image

#### Checklist


themes/ddbasic/scripts/menu.js
```diff
@@ -97,6 +97,9 @@
         } else {
           body.removeClass('overlay-is-active');
         }
+        if(body.hasClass('extended-search-is-not-open')) {
```
JrgenGNielsen, line 100: it's a bit of nitpicking, but the code standard require a space: if (...) {

JrgenGNielsen (changes requested):

JrgenGNielsen:
reviewed and approved

PR #1820, opened 2021-09-01
5698: Moved external debts into own theme function

cableman:
#### Link to issue

https://platform.dandigbib.org/issues/5139

#### Description

The issues exists when you only have external debts payments and the first part of the form is empty. Then the form do not have any submit buttons.

Fixed by not rendering external inside the form, but in it's own theme function.

#### Screenshot of the result

No screenshot

#### Checklist

No comments


JrgenGNielsen:
reviewed and approved

PR #1821, opened 2021-09-03
5237 debt count

xendk:
#### Link to issue

https://platform.dandigbib.org/issues/5237

#### Description

Fix debt count on `user/me/view` by using `ding_debt_count()` which accounts for hidden debts, rather than just counting the list.

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1823, opened 2021-09-15
4923 open graph

arniloftsson:
#### Link to issue

https://platform.dandigbib.org/issues/4923

#### Description

Enables Metatag:OpenGraph module. Defines Open Graph tags for news, events and pages

#### Screenshot of the result

#### Checklist


kasperg:
There is considerable overlap between this PR and #1587. The other PR introduces schema.org metadata for content using the `metagtag` module. OpenGraph and Schema.org are different, but the data we want to use is the same.

I suggest that we take the intent of adding OpenGraph data and apply it to #1587 instead of continuing the work here.

PR #1804, opened 2021-06-22
4578: Changed order of paragraphs

martinyde:
#### Link to issue

https://platform.dandigbib.org/issues/4578

#### Description

Changes order of paragraphs in dropdown.

#### Screenshot of the result
screenshot-ddbcms local itkdev dk-2021 06 22-13_35_10

#### Checklist

- [ x ] My complies with [our coding guidelines](../docs/code_guidelines.md).
- [ x ] My code passes our static analysis suite. If not then I have added a comment explaining why this change should be exempt from the code standards and process.
- [ x ] My code passes our continuous integration process. If not then I have added a comment explaining why this change should be exempt from the code standards and process.

#### Additional comments or questions

If you have any further comments or questions for the reviewer them please add them here.

PR #1825, opened 2021-09-30
Core: Revert features when deploying on Platform.sh

kasperg:
#### Description

Without this it can be difficult to know whether a configuration
change has been implemented properly as it is up to the tester to
revert features using admin access.

This way features are reverted automatically which also matches the
deployment procedure on staging and production environments.

This problem has been identified during the test of #1804.

#### Checklist


JrgenGNielsen (approved):
fino!

PR #1831, opened 2021-10-26
5107 remove migration 2

xendk:
#### Link to issue

https://platform.dandigbib.org/issues/5107

#### Description

Removes the missed migration variable and admin. Includes some codestyle fixes for good meassure.

#### Checklist


JrgenGNielsen (approved):
reviewed and approved

PR #1833, opened 2021-11-01
5092: Added user exists exception when status 409 is returned

cableman:
#### Link to issue

https://platform.dandigbib.org/issues/5092

#### Description

Throw `DingProviderUserExistsError` exception when FBS returns statusCode 409 at user create request.

#### Screenshot of the result

No screenshot.

#### Checklist

Scrutinizer error exists because I followed the used convention in the file.


JrgenGNielsen (changes requested):
Do you mean?: "When trying to register a user that exists, FBS returns status 409."

JrgenGNielsen (approved):

JrgenGNielsen:
reviewed and approved

PR #1834, opened 2021-11-03
5260 interest month

arniloftsson:
#### Link to issue

https://platform.dandigbib.org/issues/5260

#### Description

The text about interest period was not getting translated. Updated the field definition. The interest period did not update when updated in the dropdown menu. That was because the page did not reload on submit. That is now changed so the page reloads on submit.

#### Screenshot of the result

image

#### Checklist


JrgenGNielsen (approved):
reviewed and approved

PR #1844, opened 2021-11-25
5290: Change link hover color to inherit

yepzdk:
#### Link to issue

https://platform.dandigbib.org/issues/5290#change-55324

#### Description

Change link hover color to inherit the text color, and use opacity to distinguish the state.
Based on case where secondary color was dark, and therefore unreadable on a also dark background.

#### Screenshot of the result

Resting:
Screenshot 2021-11-25 at 09 49 00

Hover:
Screenshot 2021-11-25 at 09 55 31

#### Checklist


JrgenGNielsen (approved):
reviewed and approved

PR #1845, opened 2021-11-29
5292: Removed unnecessary fields in das request

arniloftsson:
#### Link to issue

https://platform.dandigbib.org/issues/5292

#### Description

Removed unnecessary fields from the request to das service

#### Screenshot of the result

image

#### Checklist

### DIFF
--- a/modules/ting_das/includes/ting_das.client.inc
+++ b/modules/ting_das/includes/ting_das.client.inc
@@ -18,11 +18,7 @@ class TingDasCopyRequest {
   private $wsPassword;
 
   private $pid = NULL;
-  private $userLoanerId = NULL;
-  private $userName = NULL;
   private $mail = NULL;
-  private $userInterestDays = NULL;
-  private $pickupAgencyId = NULL;
   private $agencyId = NULL;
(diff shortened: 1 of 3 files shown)


modules/ting_das/includes/ting_das.client.inc
```diff
@@ -274,11 +212,7 @@ class TingDasCopyRequest {
     $xml->addChild('ws_user', $this->wsUsername);
     $xml->addChild('ws_password', $this->wsPassword);
     $xml->addChild('pid', $this->pid);
-    $xml->addChild('user_loaner_id', $this->userLoanerId);
-    $xml->addChild('userName', $this->userName);
     $xml->addChild('userMail', $this->mail);
-    $xml->addChild('user_interest_date', $user_interest_date);
```
kasperg, line 280: Please remove the `$user_interest_date` variable above. It is no longer used.

kasperg (changes requested):
:+1: I spotted a few lines that also needs to be cleaned up. Please do so. Otherwise it looks good.

PR #1846, opened 2021-12-02
5295: Merge IMS placements into holdings

agnetemoos:
#### Link to issue

https://platform.dandigbib.org/issues/5295

#### Description

Modify holdings to provide users with more detailed placement information. This information is acquired from [IMS](https://lyngsoesystems.com/intelligent-material-management-system/). They have recently released a [SOAP Webservice endpoint](https://lyngsoe-imms.com/docs/webservices/ME-Ims4Ils-GetItemsOfBibliographicRecord.html) that exposes placement-information.

#### Screenshot of the result

ims placement

This code is at a very early stage.
The integration to the ims webservice is not completed. Mainly because it has only very recently been released. I have added stub code for now.
The code in fbs.availability.inc is in a functional state

Questions
- Does the ims client look ok? I have used the old ting_covers_addi module as boilerplate code. Is this an acceptable approach?
- The code in fbs_availability_holdings is cluttered and difficult to read. This pull request makes the mess even worse. Should code be cleaned up? If so I need help on how. 

#### Checklist

### DIFF
--- a/ding2.info
+++ b/ding2.info
@@ -4,6 +4,7 @@ description = Install a Ding based library site.
 core = 7.x
 distribution_name = "Ding 2"
 exclusive = TRUE
+version = 7.x-6.3.2
 
 ; Core modules.
 dependencies[] = block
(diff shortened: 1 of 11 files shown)


agnetemoos:
The IMS-module will be developed as a stand alone module outside DDB CMS

PR #1847, opened 2021-12-06
5292 das text

arniloftsson:
#### Link to issue

https://platform.dandigbib.org/issues/5292?next_issue_id=5291&prev_issue_id=5293

#### Description

The texts in the form where not showing if the admin page had not been saved. That was because the default settings were in the admin page. The default setting is now moved to a function so it works out off the box. Also minor css styling fixes.

#### Screenshot of the result

image

#### Checklist

### DIFF
--- a/modules/ting_das/css/ting_das.css
+++ b/modules/ting_das/css/ting_das.css
@@ -22,7 +22,7 @@
   }
 
   .popupbar .digital-order-form .form-item.form-item-mail {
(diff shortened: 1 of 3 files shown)


kasperg:
I'll close and reopen this PR to rebuild the Platform.sh site. It seems to be in a weird state.

kasperg (approved):

PR #1849, opened 2021-12-13
Redirect existing users in registration

xendk:
#### Link to issue

https://platform.dandigbib.org/issues/5092

#### Description

When trying to register, redirect existing users to their profile edit
page.

Includes #1833.

#### Checklist


JrgenGNielsen (approved):

JrgenGNielsen:
reviewed and approved

PR #1850, opened 2021-12-13
5284: Attach JS needed for Infomedia button

xendk:
-------

#### Link to issue

https://platform.dandigbib.org/issues/5284

#### Description

Infomedia button only worked on second click. Apparently the JS needed
for the ajax response to work only got loaded on the first click,
requiring another to trigger the redirect. So explicitly add the
ctools `ajax-responder` library.

#### Checklist


JrgenGNielsen (approved):
reviewed and approved

PR #1855, opened 2022-01-26
3750: Send user to login page when not logged in

arniloftsson:
#### Link to issue

https://platform.dandigbib.org/issues/3750

#### Description

Uses the same alter_inbound_url that already handles the user/me urls. There is a question about Varnish here. In order to ensuere there are no caching problems caching is disabled for these pages. May not be needed but could cause weird problems.

#### Screenshot of the result

#### Checklist

### DIFF
--- a/modules/ding_user_frontend/ding_user_frontend.module
+++ b/modules/ding_user_frontend/ding_user_frontend.module
@@ -308,6 +308,9 @@ function ding_user_frontend_url_inbound_alter(&$path, $original_path, $path_lang
         $path .= '/' . $matches[1];
       }
     } else {
+      // Don't cache this page so we don't get any weird caching problems.
+      drupal_page_is_cacheable(FALSE);
+
       // If the user is not logged in we send them to login page.
       $options = ['query' => ['destination' => $path]];
       $login_url = ding_provider_invoke( 'openplatform_token', 'login_url');
--- a/modules/ding_user_frontend/ding_user_frontend.module
+++ b/modules/ding_user_frontend/ding_user_frontend.module
@@ -307,6 +307,11 @@ function ding_user_frontend_url_inbound_alter(&$path, $original_path, $path_lang
       if (isset($matches[1])) {
         $path .= '/' . $matches[1];
       }
+    } else {
+      // If the user is not logged in we send them to login page.
+      $options = ['query' => ['destination' => $path]];
+      $login_url = ding_provider_invoke( 'openplatform_token', 'login_url');
+      drupal_goto($login_url, $options);
     }
   }
 }


modules/ding_user_frontend/ding_user_frontend.module
```diff
@@ -307,6 +307,14 @@ function ding_user_frontend_url_inbound_alter(&$path, $original_path, $path_lang
...
         $path .= '/' . $matches[1];
       }
+    } else {
+      // Don't cache this page so we don't get any weird caching problems.
+      drupal_page_is_cacheable(FALSE);
+
+      // If the user is not logged in we send them to login page.
+      $options = ['query' => ['destination' => $path]];
+      $login_url = ding_provider_invoke( 'openplatform_token', 'login_url');
+      drupal_goto($login_url, $options);
```
kasperg, line 317: I do not think this is the right way to address this:

1. It creates a dependency between the `ding_user_frontend` module and the `openplatform_token` provider.
2. The `openplatform_token` provider is used to determine the login url in a context where there is no clear line between login and Open Platform.

Drupal does not provide a hook to determine login urls but it does have a variable `site_403` which contains the path users will be shown in this context.

As an alternate solution to this problem I suggest adding an update hook to the `ding_adgangsplatformen` module which sets `site_403` to `DING_ADGANGSPLATFORMEN_LOGIN_URL`.

A similar approach is used in other Drupal modules such as [`r4032login`](https://git.drupalcode.org/project/r4032login/-/blob/7.x-1.x/r4032login.install) and [`logintoboggan`](https://git.drupalcode.org/project/logintoboggan/-/blob/7.x-1.x/logintoboggan.admin.inc#L203).

This will effectively redirect all 403 responses to Adgangsplatformen login page but I do not think that should be a problem.

modules/ding_user_frontend/ding_user_frontend.module
```diff
@@ -307,6 +307,14 @@ function ding_user_frontend_url_inbound_alter(&$path, $original_path, $path_lang
       if (isset($matches[1])) {
         $path .= '/' . $matches[1];
       }
+    } else {
+      // Don't cache this page so we don't get any weird caching problems.
+      drupal_page_is_cacheable(FALSE);
```
kasperg, line 312: :+1: for remembering this.

kasperg (changes requested):
Based on the current information I do not agree with this solution.

Perhaps it is the right way to go but I need more information about which alternatives and thoughts you might have considered.

modules/ding_user_frontend/ding_user_frontend.module
arniloftsson: I dont think the proposed solution works. The most common user case is the user accessing there status pages. E.g /user/me/status-reservations . But that returns a 404 Not Found for users who are not logged in. So the code which uses the site_403 variable is never runned.

arniloftsson: On possible solution would be to create a hook for this situation and then implement it in ding_adgangsplatformen. You get the same result as this code but without  creating a depency between ding_user_frontend and the ding_adgangsplatformen.

kasperg: :+1: I see your point @arniloftsson and I think your alternate solution sounds like a good idea. If you also still think it is a reasonable solution then let's go with that.

kasperg:
Closed in favour of #1906

PR #1859, opened 2022-02-07
5325: update Drupal to v7.86

JrgenGNielsen:
#### Link to issue

https://platform.dandigbib.org/issues/5325

#### Description

[Drupal core - Moderately critical - Cross Site Scripting - SA-CORE-2022-001](https://www.drupal.org/sa-core-2022-001)

#### Checklist

### DIFF
--- a/project-core.make
+++ b/project-core.make
@@ -3,7 +3,7 @@ api = 2
 
 ; Core
 projects[drupal][type] = core
-projects[drupal][version] = 7.82
+projects[drupal][version] = 7.86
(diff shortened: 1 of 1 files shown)


JrgenGNielsen:
Scrutinizer : not compatible with newer PHP CodeSniffer versions.  
Behat installation fails:  
cweagans/composer-patches contains a Composer plugin which is currently not in your allow-plugins config.

kasperg (approved):
:+1: LGTM

PR #1863, opened 2022-02-16
5332: Ensure payment intro text is always shown

cableman:
#### Link to issue

https://platform.dandigbib.org/issues/5332

#### Description

Always display intro text on debts page.

#### Screenshot of the result

Screenshot 2022-02-16 at 10 50 25

#### Checklist

No comments


cableman:
The Scrutinizer is not related to this change

PR #1866, opened 2022-02-25
Fixed switching between test/live in nets easy integration

cableman:
#### Link to issue

https://platform.dandigbib.org/issues/5117#change-55920

#### Description

Fixed logic statement in switching between testing and live nets easy end points.

#### Screenshot of the result

No screenhost

#### Checklist

No comments

### DIFF
--- a/modules/ding_debt_easy/ding_debt_easy.module
+++ b/modules/ding_debt_easy/ding_debt_easy.module
@@ -460,7 +460,7 @@ function ding_debt_easy_get_api_config() {
   ]);
 
   return [
-    'endpoint' => empty($endpoints['testing']) || $endpoints['testing'] ? $endpoints['test'] : $endpoints['prod'],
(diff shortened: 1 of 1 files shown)


modules/ding_debt_easy/ding_debt_easy.module
```diff
@@ -459,8 +459,9 @@ function ding_debt_easy_get_api_config() {
     'test' => 'https://test.api.dibspayment.eu/',
   ]);
 
+  $t=1;
```
JrgenGNielsen, line 462: unused variable

JrgenGNielsen (changes requested):

modules/ding_debt_easy/ding_debt_easy.module
cableman: :-) fixed

JrgenGNielsen (approved):

PR #1867, opened 2022-02-25
5117: Fixed retries type in easy payment retries

cableman:
#### Link to issue

https://platform.dandigbib.org/issues/5117

#### Description

When loading the retries counter from the database it is a string and during strict comparison with a max retries count that is an int fails.

#### Screenshot of the result

Screenshot 2022-02-25 at 11 29 24

Screenshot 2022-02-25 at 11 42 32

#### Checklist

No comments.


JrgenGNielsen (approved):
👍

PR #1874, opened 2022-03-17
5288: New placement for material-item messages

yepzdk:
#### Link to issue

[Please add a link to the issue being addressed by this change.](https://platform.dandigbib.org/issues/5288)

#### Description

Position messages in flow with content instead of absolute, wich resulted in content being hidden behind.

#### Screenshot of the result

Screenshot 2022-03-17 at 13 11 45

#### Checklist


JrgenGNielsen (approved):
reviewed and approved

PR #1875, opened 2022-03-18
4935 bpi material list

arniloftsson:
#### Link to issue

https://platform.dandigbib.org/issues/4935

#### Description

The getobject method in opensearch was used to validate the existence of materials locally on BPI imports. That method doesn't not check whether the library has holdings locally or not. Only search does. So the validation method was changed to use a search instead. The paragraphs module expects the bpi_validate_material to exist so that method was retained. 

#### Screenshot of the result

#### Checklist

### DIFF
--- a/modules/bpi/bpi.module
+++ b/modules/bpi/bpi.module
@@ -1049,7 +1049,7 @@ function bpi_validate_materials(array $ting_object_ids) {
  * @return array
  *   Ids of found objects.
  */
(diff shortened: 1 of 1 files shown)


kasperg (changes requested):
I have a few comments but in general things look good.

I appreciate description of description of the problem and solution in the pull request :+1:.

kasperg (approved):
:+1: LGTM

PR #1876, opened 2022-03-21
5307: Fixed confirmation message styling

arniloftsson:
#### Link to issue

https://platform.dandigbib.org/issues/5307

#### Description

The problem was that the form was still there which made the confirmation page extra long and scrolled down to the bottom. Fixed by not displaying the form on confirmation page.

#### Screenshot of the result

image

#### Checklist


JrgenGNielsen:
reviewed and approved

PR #1881, opened 2022-04-01
#5355: Update Role Delegation to 1.3

kasperg:
#### Link to issue

https://platform.dandigbib.org/issues/5355

#### Description

This addresses https://www.drupal.org/sa-contrib-2022-031.

#### Checklist


JrgenGNielsen (approved):
reviewed and approved

PR #1891, opened 2022-05-24
5150: Shows fulltext for lektørudtalelser on object pages

arniloftsson:
#### Link to issue

https://platform.dandigbib.org/issues/5445

#### Description

This is a part off the Seo project. It moves the text for lektørudtalelser into the object page for Seo purposes. It is done via ajax for performance reasons.

#### Screenshot of the result

image

#### Checklist

### DIFF
--- a/modules/ting_fulltext/js/ting_fulltext.js
+++ b/modules/ting_fulltext/js/ting_fulltext.js
@@ -1,0 +1,14 @@
+(function ($) {
+    "use strict";
+
+    var SearchExpansionResults = function (response) {
+        $("#ting-lektor-fulltext").html(response);
+    }
+
+    Drupal.behaviors.fullTextReview = {
+        attach: function (context) {
+            $.get(Drupal.settings.tingFulltext.url, null, SearchExpansionResults);
+        }
+    };
+
+})(jQuery);
--- a/modules/ting_fulltext/js/ting_fulltext.js
+++ b/modules/ting_fulltext/js/ting_fulltext.js
@@ -3,14 +3,14 @@
 
   Drupal.behaviors.fullTextReview = {
     attach: function (context) {
-      var selector = '#js-ting-lektor-fulltext';
+      var selector = '.js-ting-lektor-fulltext';
       $(selector, context).once('js-ting-lektor-fulltext', function () {
          var url = $(this).attr("data-fulltext-url");
          $.ajax({
            type: 'POST',
            url: url,
            success: function (result) {
-             $("#js-ting-lektor-fulltext").html(result);
+             $(".js-ting-lektor-fulltext").html(result);
            }
          });
       });
--- a/modules/ting_fulltext/js/ting_fulltext.js
+++ b/modules/ting_fulltext/js/ting_fulltext.js
@@ -1,14 +1,20 @@
 (function ($) {
-    "use strict";
+  "use strict";
 
-    var SearchExpansionResults = function (response) {
-        $("#ting-lektor-fulltext").html(response);
+  Drupal.behaviors.fullTextReview = {
+    attach: function (context) {
+      var selector = '#js-ting-lektor-fulltext';
+      $(selector, context).once('js-ting-lektor-fulltext', function () {
+         var url = $(this).attr("data-fulltext-url");
+         $.ajax({
+           type: 'POST',
+           url: url,
+           success: function (result) {
+             $("#js-ting-lektor-fulltext").html(result);
+           }
+         });
+      });
     }
-
-    Drupal.behaviors.fullTextReview = {
-        attach: function (context) {
-            $.get(Drupal.settings.tingFulltext.url, null, SearchExpansionResults);
-        }
-    };
+  };
 
 })(jQuery);
--- a/modules/ting_fulltext/js/ting_fulltext.js
+++ b/modules/ting_fulltext/js/ting_fulltext.js
@@ -6,11 +6,12 @@
       var selector = '.js-ting-lektor-fulltext';
       $(selector, context).once('js-ting-lektor-fulltext', function () {
          var url = $(this).attr("data-fulltext-url");
+         var selectedElement = $(this);
          $.ajax({
            type: 'POST',
            url: url,
            success: function (result) {
-             $(".js-ting-lektor-fulltext").html(result);
+            selectedElement.html(result);
            }
          });
       });
--- a/modules/ting_fulltext/ting_fulltext.module
+++ b/modules/ting_fulltext/ting_fulltext.module
@@ -127,6 +127,7 @@ function ting_fulltext_get_review($object_id) {
   }
   catch (Exception $e) {
     watchdog_exception('ting_fulltext', $e);
-    return drupal_json_output('');
+    drupal_add_http_header('Status', '500 Internal Server Error');
+    drupal_exit();
   }
 }
--- a/modules/ting_fulltext/ting_fulltext.module
+++ b/modules/ting_fulltext/ting_fulltext.module
@@ -84,7 +84,7 @@ function ting_fulltext_preprocess_ting_relation(&$variables) {
     if ($ting_object instanceof OpenSearchTingObject && in_array('docbook', $relation->getObject()->getTingObject()->getFormatsAvailable())) {
       // For performance reasons we load the review in full text as ajax.
       drupal_add_js(drupal_get_path('module', 'ting_fulltext') . '/js/ting_fulltext.js');
-      $url = url('ting/object/' . $relation->uri . '/fulltext/ajax') ;
+      $url = url('ting/object/' . $relation->uri . '/fulltext/ajax');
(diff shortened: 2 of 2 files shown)


modules/ting_fulltext/js/ting_fulltext.js
```diff
@@ -0,0 +1,14 @@
...
+
+    var SearchExpansionResults = function (response) {
+        $("#ting-lektor-fulltext").html(response);
+    }
+
+    Drupal.behaviors.fullTextReview = {
+        attach: function (context) {
+            $.get(Drupal.settings.tingFulltext.url, null, SearchExpansionResults);
+        }
+    };
```
kasperg, line 12: Drupal behaviors can trigger multiple times on a page e.g. for the initial pageload and if new content is loaded via AJAX. The current implementation will cause the fulltext content to load multiple times. Also the current implementation requires only a single instance of the fulltext element. This maybe correct at the moment but it is not something that we know to hold.

I suggest that the current approach is updated to look for fulltext containers (probably identified with a `js-ting-fulltext-lektor` class or the like) within `context` and that the data needed is retrieved from a url set as a data-attribute within that element instead of a global JavaScript setting.

modules/ting_fulltext/ting_fulltext.module
```diff
@@ -23,6 +23,14 @@ function ting_fulltext_menu() {
     'type' => MENU_CALLBACK | MENU_VISIBLE_IN_BREADCRUMB,
   );
 
+  $items['ting/fulltext/ajax/%'] = array(
```
kasperg, line 26: Is there a reason the path for this does not follow the other path declared by the module e.g. 'ting/object/%ting_fulltext_object/review`?

Also: If you use `%ting_fulltext_object` as the parameter name like we do with the path above `ting/object/%ting_fulltext_object/fulltext` then our loader `ting_fulltext_object_load()` will be called automatically and the result passed as the argument to `ting_fulltext_get_review()` and you do not have to call it yourself.

modules/ting_fulltext/ting_fulltext.module
```diff
@@ -92,3 +102,32 @@ function ting_fulltext_ctools_plugin_directory($module, $plugin) {
...
 }
+
+/**
+ * Ajax call function wich returns the review in full text.
+ */
+function ting_fulltext_get_review($object_id) {
+  try {
+    $path = drupal_get_path('module', 'ting_fulltext');
+    // Include functions to parse the xml.
+    include_once $path . '/includes/ting_fulltext.pages.inc';
```
kasperg, line 113: Please use [`module_load_include()`](https://api.drupal.org/api/drupal/includes%21module.inc/function/module_load_include/7.x) for this.

modules/ting_fulltext/ting_fulltext.module
```diff
@@ -92,3 +102,32 @@ function ting_fulltext_ctools_plugin_directory($module, $plugin) {
...
+
+    $xml = ting_fulltext_object_load($object_id);
+    $full_text = ting_fulltext_parse($xml);
+    // We dont need to display the title again.
+    unset($full_text['title']);
```
kasperg, line 118: :+1: for explaining.

modules/ting_fulltext/ting_fulltext.module
```diff
@@ -92,3 +102,32 @@ function ting_fulltext_ctools_plugin_directory($module, $plugin) {
...
+    );
+    $output = drupal_render($results);
+
+    return drupal_json_output($output);
+  }
+  catch (Exception $e) {
+    watchdog_exception('ting_fulltext', $e);
+    drupal_add_http_header('Status', '500 Internal Server Error');
+    drupal_exit();
+  }
```
kasperg, line 132: I am not sure you need to do this error handling yourself. If an exceptions bubbles up from a page callback Drupal should log the error and return a error code 500 response anyway.

modules/ting_fulltext/ting_fulltext.module
```diff
@@ -67,15 +75,17 @@ function ting_fulltext_object_load($object_id) {
...
     $ting_object = $relation->getObject()->getTingObject();
 
     // This will only work for the opensearch search provider, so check what
     // type of object we have.
     if ($ting_object instanceof OpenSearchTingObject && in_array('docbook', $relation->getObject()->getTingObject()->getFormatsAvailable())) {
-      $variables['fulltext_link'] = l(t('See the full-text version'), 'ting/object/' . $relation->uri . '/fulltext');
+      // For performance reasons we load the review in full text as ajax.
+      drupal_add_js(drupal_get_path('module', 'ting_fulltext') . '/js/ting_fulltext.js');
+      drupal_add_js(array('tingFulltext' => array('url' => '/ting/fulltext/ajax/' . $relation->uri)), 'setting');
+      $variables['abstract'] = '<div id="ting-lektor-fulltext">' . $variables['abstract'] . '</div>';
```
kasperg, line 88: If the original path `'ting/object/' . $relation->uri . '/fulltext` is obsolete then consider removing it from the codebase altogether.

modules/ting_fulltext/ting_fulltext.module
```diff
@@ -92,3 +102,32 @@ function ting_fulltext_ctools_plugin_directory($module, $plugin) {
...
+    // We dont need to display the title again.
+    unset($full_text['title']);
+
+    $results = array(
+      '#theme' => 'ting_fulltext',
+      '#fields' => $full_text,
+    );
+    $output = drupal_render($results);
+
+    return drupal_json_output($output);
```
kasperg, line 126: Do we need to encode the render as JSON when what we are really doing is returning the HTML from our theme function which is to be inserted as is in a DOM node?

kasperg (changes requested):
:+1: I have some comments regarding how this has been solved.

modules/ting_fulltext/ting_fulltext.module
arniloftsson: Whether to remove the original path is a design decision. I dont know if there are use cases for it.

modules/ting_fulltext/ting_fulltext.module
```diff
@@ -67,15 +75,17 @@ function ting_fulltext_object_load($object_id) {
...
     $ting_object = $relation->getObject()->getTingObject();
 
     // This will only work for the opensearch search provider, so check what
     // type of object we have.
     if ($ting_object instanceof OpenSearchTingObject && in_array('docbook', $relation->getObject()->getTingObject()->getFormatsAvailable())) {
-      $variables['fulltext_link'] = l(t('See the full-text version'), 'ting/object/' . $relation->uri . '/fulltext');
+      // For performance reasons we load the review in full text as ajax.
+      drupal_add_js(drupal_get_path('module', 'ting_fulltext') . '/js/ting_fulltext.js');
+      $url = url('ting/object/' . $relation->uri . '/fulltext/ajax');
+      $variables['abstract'] = '<div id="js-ting-lektor-fulltext" data-fulltext-url="' . $url . '">' . $variables['abstract'] . '</div>';
```
kasperg, line 88: Please do not use HTML ids for elements that you cannot guarantee to be unique.

As far as I can tell it should be fine to use `class="js-ting-lektor-fulltext"` instead of `id="js-ting-lektor-fulltext"` and update the behavior accordingly.

kasperg (changes requested):
Please use a class to target elements instead of an id. HTML ids must be unique. We cannot (and need not) guarantee this to be the case.

modules/ting_fulltext/js/ting_fulltext.js
```diff
@@ -0,0 +1,20 @@
...
+    attach: function (context) {
+      var selector = '.js-ting-lektor-fulltext';
+      $(selector, context).once('js-ting-lektor-fulltext', function () {
+         var url = $(this).attr("data-fulltext-url");
+         $.ajax({
+           type: 'POST',
+           url: url,
+           success: function (result) {
+             $(".js-ting-lektor-fulltext").html(result);
+           }
```
kasperg, line 14: This approach to using behaviors does not get the concepts of ids, context and selectors quite right.

1. Because we cannot know if there are multiple fulltext lektørudtalelser on a page we cannot use an HTML id to identify or select them. HTML IDs must be unique. Instead this PR currently uses a `js-` prefixed classname. This is good.
2. Drupal behaviors work within a context. This may be the entire page on the initial pageload or elements that are added at a later point in time through AJAX.  This PR currently takes the context into account when looking for new  elements to process. This is good.
3. A behavior may operate multiple times on the same element. This may not be a good thing(tm) and can be avoided using `once()` as is the case in this PR. This is good.
4. After completing the AJAX request to retrieve a lektørudtalelse the content should be inserted into the element for which it has been retrieved. This PR inserts the content into any HTML element matching the selector. This is a problem if there are multiple elements on the same page. The content should only be inserted into the element of which the content was retrieved. This PR inserts the content into all elements matching the selector. This is bad.

kasperg (changes requested):

kasperg (approved):
:+1: LGTM

PR #1903, opened 2022-07-27
5282: Changed height to work on horisontal phones

arniloftsson:
#### Link to issue

https://platform.dandigbib.org/issues/5282

#### Description

Scrolling of a big menu was not working on mobilephone when they where horisontal. Needed a little tweak of the height. There is a bigger problem with the combination of using fixed position and high zindex in the menu which causes problems with scrolling. But thats outside the scope of this issue.

#### Screenshot of the result

#### Checklist


JrgenGNielsen (approved):
reviewed and approved

PR #1904, opened 2022-08-01
4935: Dont validate materials which the library does not have

arniloftsson:
#### Link to issue

https://platform.dandigbib.org/issues/4935#note-46

#### Description

Handles the issue that nodes with materials imported from BPI which the library does not have cant be saved. That happens because the validation function does not allow it. The solution is another check so the function does not validate materials which are not in the libraries holdings.

#### Screenshot of the result

#### Checklist


JrgenGNielsen (approved):
reviewed and approved

PR #1906, opened 2022-08-15
3750 log in page hook

arniloftsson:
#### Link to issue

https://platform.dandigbib.org/issues/3750

#### Description

When user arent logged in on user status pages. Eg. /user/me/loans they are sent automatically to a login page. In order not to create dependency this is done with a hook. There is a reguirement that page parameters are preserved in the process. See PR: https://github.com/ding2/ding2/pull/1855 for discussion.

#### Screenshot of the result

#### Checklist

### DIFF
--- a/modules/ding_adgangsplatformen/ding_adgangsplatformen.module
+++ b/modules/ding_adgangsplatformen/ding_adgangsplatformen.module
@@ -333,7 +333,7 @@ function ding_adgangsplatformen_login() {
  */
 function ding_adgangsplatformen_login_provider_user() {
   // We save the current page as destination.
-  $options = ['query' => ['destination' => request_uri()]];
+  $options = ['query' => drupal_get_destination()];
   // The classes in the oauth2-client don't get loaded unless we go to the login page.
   drupal_goto(DING_ADGANGSPLATFORMEN_LOGIN_URL, $options);
 }
--- a/modules/ding_adgangsplatformen/ding_adgangsplatformen.module
+++ b/modules/ding_adgangsplatformen/ding_adgangsplatformen.module
@@ -335,8 +335,9 @@ function ding_adgangsplatformen_login() {
  *   If required libraries are not loaded.
  */
 function ding_adgangsplatformen_login_provider_user() {
-  // We save the current page including query parameters before sending the user to the login page.
-  ding_user_set_destination(request_uri());
+  // We save the current page including query parameters as destination parameters. That gets saved in the
+  // ding_adgangsplatformen_generate_login_url method.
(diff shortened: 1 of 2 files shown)


modules/ding_adgangsplatformen/ding_adgangsplatformen.module
```diff
@@ -326,6 +326,19 @@ function ding_adgangsplatformen_login() {
   drupal_exit($authorization_url);
 }
 
+/**
+ * Implements hooK_login_provider_user.
+ *
+ * Saves the current path and sends the user to the login page.
+ *
+ */
```
kasperg, line 334:
```suggestion
/**
 * Implements hook_login_provider_user().
 *
 * Saves the current path and sends the user to the login page.
 */
```

Please ensure that the code follows our coding standards.

modules/ding_user_frontend/ding_user_frontend.module
```diff
@@ -307,6 +307,9 @@ function ding_user_frontend_url_inbound_alter(&$path, $original_path, $path_lang
       if (isset($matches[1])) {
         $path .= '/' . $matches[1];
       }
+    } else {
+      // Send the user to a login page if there is a module that implements the login_provider_user hook.
+      module_invoke_all('login_provider_user');
```
kasperg, line 312:
```suggestion
    }
    else {
      // Send the user to a login page if there is a module that implements the login_provider_user hook.
      module_invoke_all('login_provider_user');
```

Please ensure that the code follows our coding standards.

kasperg (changes requested):
Please ensure that the code follows our coding standards.

modules/ding_adgangsplatformen/ding_adgangsplatformen.module
```diff
@@ -326,6 +326,19 @@ function ding_adgangsplatformen_login() {
...
+function ding_adgangsplatformen_login_provider_user() {
+  // We save the current page as destination.
+  $options = ['query' => ['destination' => request_uri()]];
```
kasperg, line 337:
```suggestion
  $options = ['query' => drupal_get_destination()];
```

We should be able to use [`drupal_get_destination()`](https://api.drupal.org/api/drupal/includes!common.inc/function/drupal_get_destination/7.x) here to use the framework and avoid edge cases.

kasperg (changes requested):
:+1: I like the approach here.

Please ensure that Drupals coding standards are followed. I also have a suggestion for how we can use the framework provided by D7 better.

kasperg (approved):
:+1: LGTM

PR #1909, opened 2022-08-24
5448: Patch cookie informations module to downgrade PHP requirements

cableman:
#### Link to issue

https://platform.dandigbib.org/issues/5448

#### Description

Downgrade the requirement for PHP 7.1

#### Screenshot of the result

No screenshot

#### Checklist

No comment


kasperg (approved):
:+1: LGTM

PR #1910, opened 2022-08-25
5490 admin tables styling

arniloftsson:
#### Link to issue

https://platform.dandigbib.org/issues/5490

#### Description

The cookieinformation service adds a css definition of table which overrides the admin theme definition. So we need a more specific css definition. This happens in the Seven  admin theme so changing the DDBasic theme has no effect. Therefore we need to add a css file somewhere. Here we have choosen the ding_content module which serves as a generel module. Other placement could also be used.

#### Screenshot of the result

image

#### Checklist


kasperg:
This is now the second time that we have to change our own CSS because of Cookieinformation. That does not seem right.

I would expect the Cookieinformation CSS to be as self contained as possible e.g. scoped to only target their own elements. Is that not the case?

arniloftsson:
It almost is selfcontained. Most of the css are definition like this:

.coi-consent-banner__found-cookies {
  display: flex;
  flex-direction: column;
}

But it has one problematic definition:
table td,
table th{
  padding: 0;
  border: none;
}

Which is the one causing al the problems.

kasperg:
That is a bad approach for something you would expect to work across many different sites. [Based on the documentation](https://support.cookieinformation.com/en/articles/5462550-styling-the-consent-pop-up) it seems as if you can control (some of) the CSS from their solution. Are we sure that the problematic CSS is not inserted here?

arniloftsson:
The css seems to be standard. I found it in 3 different consent solutions. But we can change it and remove the problematic line from the cookieinformation css. But then we have to maintain that css. I don't know if it gets updated when we change it.

As I see it there are 3 solutions:

1. Fix it in our end.
2. Fix it on the cookieinformation admin page and change their css.
3. Contact them at get them to fix it.

3 is the best solution but it takes time which is a problem right now.

kasperg:
> The css seems to be standard. I found it in 3 different consent solutions.

Do you have any reference for this? I checked a couple of sites that use Cookieinformation including https://cookieinformation.com/ and https://www.telia.dk/. I could not see it here. None of these pages include any tables either.

I get your point that we do not want to maintain that CSS but in the current state it forces us to force all sorts of changes to this project anyway which we then own. In that regard I would prefer to fix it at the root instead of finding more places which is affected, doing more releases etc.

PR #1911, opened 2022-09-04
5448: Handles removing cookies from sub domains

arniloftsson:
#### Link to issue

https://platform.dandigbib.org/issues/5448#note-30

#### Description

There is problem that cookies don't get removed on pages with a subdomain. The cookie is registered to the root domain but windows.location.hostname returns the full hostname.  This solution is based on code from the eu cookie complience modules and iteratets over all the domains and attempts to remove the cookies.

Note this can't be properly tested. In order to test this we need a domain which includes a subdomain which is registred with KPI index as a domain to test Mapp on. My local dev domain is bur contains only a root domain. Upgrade-fbs can be used..

#### Screenshot of the result

#### Checklist


kasperg (approved):
:+1: LGTM

PR #1917, opened 2022-09-23
5448: added background and styling to placeholder

arniloftsson:
#### Link to issue

https://platform.dandigbib.org/issues/5448#note-45 issue b.

#### Description

Styling of the video placeholder which is sat when the user hasn't given consent to cookies.

#### Screenshot of the result

image

#### Checklist

### DIFF
--- a/themes/ddbasic/sass/components/node/node.scss
+++ b/themes/ddbasic/sass/components/node/node.scss
@@ -76,15 +76,10 @@
 
 // Styling of cookie consent placeholders. They can be on any node type
 // and on ekursus objects.
-// The cookieinformation script automatically sets style="display: block;" on the element.
-// Therefore the need for !important override.
 .consent-placeholder {
-  display: flex !important;
-  flex-direction: column;
-  justify-content: center;
-  align-items: center;
-  background-color: #e5e5e5;
-  padding: 0 20px;
+  background-color: $grey;
+  padding: 100px 20px;
+	text-align: center;
 }
 
 .ekursus .consent-placeholder {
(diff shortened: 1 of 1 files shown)


themes/ddbasic/sass/components/node/node.scss
```diff
@@ -73,3 +73,20 @@
     }
   }
 }
+
+// Styling of cookie consent placeholders. They can be on any node type
+// and on ekursus objects.
+// The cookieinformation script automatically sets style="display: block;" on the element.
+// Therefore the need for !important override.
+.consent-placeholder {
+  display: flex !important;
```
kasperg, line 82: :+1: Excellent commenting.

themes/ddbasic/sass/components/node/node.scss
```diff
@@ -73,3 +73,20 @@
...
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+  background-color: #e5e5e5;
```
kasperg, line 86: As far as I can tell this is the value for [the SCSS variable `$grey`](https://github.com/ding2/ding2/blob/5c0c287e75b6d353e737b0cf48d2441d91086121/themes/ddbasic/sass/configuration/_variables.scss#L22). Is there a reason not to use it?

kasperg (changes requested):
:+1: I have a single thing I would like you to look into.

themes/ddbasic/sass/components/node/node.scss
arniloftsson: Changed the css. Display block was being used to toogle the element on or off so we cant use display:flex here. Styled it in a different way. Updated background color to use variable.

kasperg (approved):

kasperg (approved):
:+1: LGTM

